PR #1, opened 2018-05-11
Use spring-boot-template with small customisations

KarolJastrzebski:
- Renamed package: demo -> bulkscanning
- Removed docker-related files as the architecture is not known yet.
- Wrote minimalistic README.md


src/main/java/uk/gov/hmcts/reform/bulkscanning/controllers/RootController.java
```diff
@@ -0,0 +1,28 @@
...
+     *
+     * <p>Azure application service has a hidden feature of making requests to root endpoint when
+     * "Always On" is turned on.
+     * This is the endpoint to deal with that and therefore silence the unnecessary 404s as a response code.
+     *
+     * @return Welcome message from the service.
+     */
+    @GetMapping
+    public ResponseEntity<String> welcome() {
+        return ok("Welcome to spring-boot-template");
```
doncem: oopsie

src/main/resources/application.yaml
```diff
@@ -0,0 +1,11 @@
+server:
+  port: 4550
```
doncem: here too

.env
```diff
@@ -0,0 +1 @@
+SERVER_PORT=4550
```
doncem: what will the port number be?

src/test/java/uk/gov/hmcts/reform/bulkscanning/controllers/GetWelcomeTest.java
```diff
@@ -0,0 +1,28 @@
...
+public class GetWelcomeTest {
+
+    @Autowired
+    private transient MockMvc mockMvc;
+
+    @Test
+    public void should_welcome_upon_root_request_with_200_response_code() throws Exception {
+        MvcResult response = mockMvc.perform(get("/")).andExpect(status().isOk()).andReturn();
+
+        assertThat(response.getResponse().getContentAsString()).startsWith("Welcome");
```
nitinprabhu: You could Assert with complete String :)

web.config
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nitinprabhu: Platform config can be done as a separate PR.This way it would be very clear to understand.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanning/ApplicationTests.java
```diff
@@ -0,0 +1,19 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class ApplicationTests {
+
+    @Test
+    public void sample_test() {
+        assertThat(2).isGreaterThan(1);
```
nitinprabhu: Not sure why we want this ?

.travis.yml
```diff
@@ -0,0 +1,29 @@
+language: java
```
nitinprabhu: Enabling Travis could be a separate PR

.env
nitinprabhu: This file could be added as part of Docker PR

.travis.yml
KarolJastrzebski: True, but it's simple enough to have it as a base, even if it's only for running tests and jacoco.

PR #4, opened 2018-06-13
Let services use separate containers

KarolJastrzebski:
https://tools.hmcts.net/jira/browse/RPE-511

Each service will have its own container so that documents will not be
mixed up. A container is named after the service it was provided for.

`STORAGE_CONTAINER_NAME` variable is removed as names are read from
storage account.

```
[x] Yes
[ ] No
```


rkondratowicz:
could you remove stuff that's not related to the Jira story / branch description from this PR?

KarolJastrzebski:
@rkondratowicz is that about sonarqube? If so, then please note that pipeline failed trying to run the task, hence it was added.

PR #6, opened 2018-06-18
Create SAS endpoint to generate SAS token

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-493

### Change description ###

- Added functionality to generate SAS token as per service configuration.

- Added unit tests.Could not add negative test cases as Azure Storage classes are final and hence could not be mocked.

- Functional tests will be added as a separate PR.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+    public String generateSasToken(String serviceName) {
+        //Based on the service name container reference would be returned for specific service
+        try {
+            log.info("SAS Token request received for service {} ", serviceName);
+
+            CloudBlobContainer cloudBlobContainer = cloudBlobClient.getContainerReference(serviceName);
+
+            return cloudBlobContainer.generateSharedAccessSignature(createSharedAccessPolicy(serviceName), null);
+        } catch (URISyntaxException | StorageException | InvalidKeyException e) {
+            throw new UnableToGenerateSasTokenException(e);
```
doncem: `URISyntaxException` doesn't sound to me as `Internal Server Error`. Or is it something like "badly configured server with uri to container"? Then it is `Internal Server Error`

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+
+        return policy;
+    }
+
+    private AccessTokenConfiguration.TokenConfig getTokenConfigForService(String serviceName) {
+        return accessTokenConfiguration.getServiceConfig().stream()
+            .filter(tokenConfig -> tokenConfig.getServiceName().equalsIgnoreCase(serviceName))
+            .findFirst()
+            .orElseThrow(
+                () -> new ServiceConfigNotFoundException("No service configuration found for service " + serviceName)
```
doncem: Does the exception reach spring advice? Is it handled somewhere?

src/main/java/uk/gov/hmcts/reform/bulkscanning/controllers/SasTokenController.java
```diff
@@ -0,0 +1,30 @@
...
+@RestController
+public class SasTokenController {
+
+    private final SasTokenGeneratorService tokenGeneratorService;
+
+    public SasTokenController(SasTokenGeneratorService tokenGeneratorService) {
+        this.tokenGeneratorService = tokenGeneratorService;
+    }
+
+    @GetMapping(path = "/{serviceName}", produces = MediaType.APPLICATION_JSON_VALUE)
```
KarolJastrzebski: I would add a prefix to avoid even accidental name clashes. Anything would do, `sas`, `token`, or `services`, or `keys`, or... you get the point :grin:

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+    private EnumSet<SharedAccessBlobPermissions> getBlobPermissions(AccessTokenConfiguration.TokenConfig config) {
+        EnumSet<SharedAccessBlobPermissions> blobPermissions = EnumSet.noneOf(SharedAccessBlobPermissions.class);
+
+        List<String> configuredPermissions = Arrays.asList(config.getPermissions().split(","));
+
+        configuredPermissions.forEach(perm -> {
+            if (EnumUtils.isValidEnum(SharedAccessBlobPermissions.class, perm)) {
+                blobPermissions.add(SharedAccessBlobPermissions.valueOf(perm));
+            }
+        });
```
doncem: Looks like this block of code can be simplified into classic for loop. Correct me if I'm wrong, but rather then assigning variable and looping over, we could just

```java
for (String permission : config.getPermissions().split(",")) {
    if (...) {
        blobPermissions.add(...);
    }
}
```

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+        return blobPermissions;
+    }
+}
```
doncem: General note on the class: there are bunch of `AccessTokenConfiguration.TokenConfig` use cases. Can we just import the `TokenConfig` and use that instead?

src/main/resources/application.yaml
```diff
@@ -16,5 +16,11 @@ info:
     description: Ingest scanned documents.
 
 storage:
-  account_name: ${STORAGE_ACCOUNT_NAME}
-  key: ${STORAGE_KEY}
+  account_name: ${STORAGE_ACCOUNT_NAME:bulkscanning}
+  key: ${STORAGE_KEY:dGVzdA==} #Base 64 encoded
+
+accesstoken:
```
doncem: Just a suggestion: let's keep camel case everywhere for readability. Or underscores. Or whichever format is OK with us

src/test/java/uk/gov/hmcts/reform/bulkscanning/controllers/GetWelcomeTest.java
```diff
@@ -14,7 +15,8 @@
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 @RunWith(SpringRunner.class)
-@WebMvcTest
+@SpringBootTest
```
doncem: Becomes an integration test

src/test/java/uk/gov/hmcts/reform/bulkscanning/controllers/SasTokenControllerTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Date;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
```
doncem: Integration test here too

src/main/java/uk/gov/hmcts/reform/bulkscanning/exceptionhandlers/ResponseExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR;
+import static org.springframework.http.ResponseEntity.status;
+
+@ControllerAdvice
+public class ResponseExceptionHandler extends ResponseEntityExceptionHandler {
+
+    private static final Logger log = LoggerFactory.getLogger(ResponseExceptionHandler.class);
+
+    @ExceptionHandler(UnableToGenerateSasTokenException.class)
+    protected ResponseEntity<String> handleJsonProcessingException() {
```
rkondratowicz: wrong method name

src/main/java/uk/gov/hmcts/reform/bulkscanning/exceptions/ServiceConfigNotFoundException.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.hmcts.reform.bulkscanning.exceptions;
+
+import uk.gov.hmcts.reform.logging.exception.AlertLevel;
+import uk.gov.hmcts.reform.logging.exception.UnknownErrorCodeException;
+
+/**
+ * SonarQube reports as error. Max allowed - 5 parents
+ */
```
rkondratowicz: This is not javadoc comment for this class, it should be a regular comment for the annotation (if it is actually really needed).

(I know this is done in other repos as well)

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+    ) {
+        this.cloudBlobClient = cloudBlobClient;
+        this.storageCredentials = storageCredentials;
+        this.accessTokenConfiguration = accessTokenConfiguration;
+    }
+
+    public String generateSasToken(String serviceName) {
+        //Based on the service name container reference would be returned for specific service
+        try {
+            log.info("SAS Token request received for service {} ", serviceName);
```
rkondratowicz: could be done before `try`

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
rkondratowicz: or dont use any foreach at all and use .filter .map

src/main/java/uk/gov/hmcts/reform/bulkscanning/config/AccessTokenConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public List<TokenConfig> getServiceConfig() {
+        return serviceConfig;
+    }
+
+    public void setServiceConfig(List<TokenConfig> serviceConfig) {
+        this.serviceConfig = serviceConfig;
+    }
+
+    public static class TokenConfig {
```
banderous: To be discussed tomorrow, but this PR would be simpler if we used a single container for all our services; we wouldn't need this configuration, or a parameterized endpoint or exceptions in case they ask for a container that doesn't exist.

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
```diff
@@ -0,0 +1,95 @@
...
+    private AccessTokenConfiguration.TokenConfig getTokenConfigForService(String serviceName) {
+        return accessTokenConfiguration.getServiceConfig().stream()
+            .filter(tokenConfig -> tokenConfig.getServiceName().equalsIgnoreCase(serviceName))
+            .findFirst()
+            .orElseThrow(
+                () -> new ServiceConfigNotFoundException("No service configuration found for service " + serviceName)
+            );
+    }
+
+    private EnumSet<SharedAccessBlobPermissions> getBlobPermissions(AccessTokenConfiguration.TokenConfig config) {
```
banderous: Do we need to make permissions configurable? Can't we hard code this?

AFAIK we should only ever grant Write and List permissions; making this configurable introduces a risk of misconfiguration where we accidentally grant the token permissions to download the files.

src/main/java/uk/gov/hmcts/reform/bulkscanning/config/AccessTokenConfiguration.java
nitinprabhu: Yes I agree it would be simpler.Current implementation is as per discussion with Salman.I will have a chat with him to see if we can make it as a single container.

src/main/java/uk/gov/hmcts/reform/bulkscanning/services/SasTokenGeneratorService.java
nitinprabhu: I initially thought to hardcode it but then thought when we want to download we will need to modify the code again so thought will make it configurable as it will then give us ability to configure it for download too.

banderous: When we download we don't need to use a SAS, we can just use the normal Azure credentials & blob API.

src/main/java/uk/gov/hmcts/reform/bulkscanning/exceptionhandlers/ResponseExceptionHandler.java
```diff
@@ -0,0 +1,29 @@
...
+    @ExceptionHandler(UnableToGenerateSasTokenException.class)
+    protected ResponseEntity<String> handleUnableToGenerateSasTokenException() {
+        return status(INTERNAL_SERVER_ERROR).body("Exception occurred while generating SAS Token");
+    }
+
+    @ExceptionHandler(Exception.class)
+    protected ResponseEntity<Void> handleInternalException(Exception exception) {
+        log.error(exception.getMessage(), exception);
+        return status(INTERNAL_SERVER_ERROR).build();
+    }
```
rkondratowicz: missing handler for `UnableToGenerateSasTokenException`?

src/main/resources/application.yaml
```diff
@@ -16,5 +16,11 @@ info:
...
+  serviceConfig:
+    - serviceName: sscs
+      permissions: WRITE,LIST
```
rkondratowicz: should this actually be configurable?

src/main/java/uk/gov/hmcts/reform/bulkscanning/controllers/SasTokenController.java
```diff
@@ -0,0 +1,30 @@
...
+@RestController
+public class SasTokenController {
+
+    private final SasTokenGeneratorService tokenGeneratorService;
+
+    public SasTokenController(SasTokenGeneratorService tokenGeneratorService) {
+        this.tokenGeneratorService = tokenGeneratorService;
+    }
+
+    @GetMapping(path = "/sas/{serviceName}", produces = MediaType.APPLICATION_JSON_VALUE)
```
doncem: I think @KarolJastrzebski meant prefix in `@RestController`. And I vote for `token`, `sas` is kind of DSL-y. But that's just a vote

src/main/java/uk/gov/hmcts/reform/bulkscanning/config/AccessTokenConfigurationProperties.java
```diff
@@ -0,0 +1,48 @@
+package uk.gov.hmcts.reform.bulkscanning.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import java.util.List;
+
+@ConfigurationProperties("accesstoken")
+public class AccessTokenConfigurationProperties {
```
doncem: Now that I look at this I think of a shorter acceptable naming: `AccessTokenProperties`. But it's not that important really

src/main/resources/application.yaml
nitinprabhu: spring boot throws exception if we use camel case/underscores

src/main/resources/application.yaml
nitinprabhu: I had a chat with Salman about this and he thinks it's better to make it configurable as we can remove/add permissions through config file rather than changing source code.Also a PR need to be created and approved before anyone updates the permissions so that way it won't go wrong.

nitinprabhu: I know we had a chat about it in the morning but what I got to know is LIST permission could be removed later that is one of the requirement.Alternatively what we are gaining hardcoding it ? I think config makes it easier to change rather than code.

banderous: Because this service should never dispense a SAS with READ permissions; that's a security requirement.

If we hard code the permissions and add comments explaining why, then we make this clear.

If we instead make these configurable, the implication is that they can be changed arbitrarily. 

When I see something is configurable, I assume there must be a legitimate scenario that would justify changing it, or why would it be configurable?

banderous: Also, we cannot remove LIST permissions since Exela would not be able to use Azure's utilities to upload files.

nitinprabhu: They can't be changed arbitrarily.You will need to update the config and deploy isn't it ? How come it can be changed arbitrarily.

nitinprabhu: You only need write permissions to upload files.

banderous: Bulk copy tools need to be able to LIST files to see which files they need to upload, and which have already been uploaded.

The fact that the config has to be changed by PR is irrelevant; these permissions are constants and should be coded as such - it is misleading to code anything else; you wouldn't put PI in a configuration file, so why would you put these permissions?

src/main/java/uk/gov/hmcts/reform/bulkscanning/exceptionhandlers/ResponseExceptionHandler.java
rkondratowicz: oops, I meant `ServiceConfigNotFoundException` (which should result in 4xx) I assume

PR #8, opened 2018-06-20
Use product variable in storage account name

KarolJastrzebski:
Builds are failing due to error returned from Azure with code
`StorageAccountAlreadyExists`. After deleting the resources the build
can be re-ran.

Each build implicitly is given information about pull request number
that is included in `product` variable. It follows `pr-{number}-rpe`
scheme.

Solution to the error is to include `product` value in the storage
account name. To make it more challenging, Azure has certain limitations
for account naming:

 - length between 3 and 24 characters
 - alphanumerics only

Because of that, any dashes in `product` value have to be removed, and
the final value has to be trimmed down to 24 characters.

```
[x] Yes
[ ] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -6,6 +6,8 @@ provider "vault" {
 
 locals {
   app = "bulk-scanning"
+  base_account_name = "${var.product}bulkscan${var.env}"
+  account_name = "${replace(local.base_account_name, "-", "")}"
 }
 
 module "backend" {
@@ -40,7 +42,7 @@ resource "azurerm_resource_group" "rg" {
 }
 
 resource "azurerm_storage_account" "provider" {
-  name                      = "bulkscanning${var.env}"
+  name                      = "${substr(local.account_name, 0, 24)}"
   resource_group_name       = "${azurerm_resource_group.rg.name}"
   location                  = "${var.location}"
   account_tier              = "Standard"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -42,7 +42,7 @@ resource "azurerm_resource_group" "rg" {
 }
 
 resource "azurerm_storage_account" "provider" {
-  name                      = "${substr(local.account_name, 0, 24)}"
+  name                      = "${substr(local.account_name, 0, min(length(local.account_name), 24))}"
   resource_group_name       = "${azurerm_resource_group.rg.name}"
   location                  = "${var.location}"
   account_tier              = "Standard"


infrastructure/main.tf
```diff
@@ -6,6 +6,8 @@ provider "vault" {
 
 locals {
   app = "bulk-scanning"
+  base_account_name = "${var.product}bulkscan${var.env}"
```
lgonczar: Why no hyphens? We separate name components with hyphens by convention.

nitinprabhu: hyphens are not allowed for storage account

infrastructure/main.tf
```diff
@@ -40,7 +42,7 @@ resource "azurerm_resource_group" "rg" {
 }
 
 resource "azurerm_storage_account" "provider" {
-  name                      = "bulkscanning${var.env}"
+  name                      = "${substr(local.account_name, 0, 24)}"
```
lgonczar: I think it would be best to see an error when the name is too long. Otherwise, we risk having environment component taken out of the name.

infrastructure/main.tf
lgonczar: Fair enough

infrastructure/main.tf
nitinprabhu: this will be only for preview I believe so yes it would be good if we can do this only for preview environment and not trim for any other.

infrastructure/main.tf
KarolJastrzebski: This is one of the limitations for account naming: alphanumerics only, no dashes or underscores

infrastructure/main.tf
```diff
@@ -40,7 +42,7 @@ resource "azurerm_resource_group" "rg" {
 }
 
 resource "azurerm_storage_account" "provider" {
-  name                      = "bulkscanning${var.env}"
+  name                      = "${substr(local.account_name, 0, min(length(local.account_name), 24))}"
```
lgonczar: This is still equivalent to `${substr(local.account_name, 0, 24))}`

KarolJastrzebski: I have described the issue in the c9a6c3d8f9570dfa770db850b3273758682296e4's message:
>Terraform's substr() expects that offset + length cannot be larger than the length of the string.

infrastructure/main.tf
lgonczar: I should have read the description

infrastructure/main.tf
lgonczar: I see. I thought it was the reaction to our previous comments on this line:
```
@lgonczar
lgonczar 22 minutes ago  Member
I think it would be best to see an error when the name is too long. Otherwise, we risk having environment component taken out of the name.

  @nitinprabhu
nitinprabhu 20 minutes ago  Member
this will be only for preview I believe so yes it would be good if we can do this only for preview environment and not trim for any other.
```

lgonczar: Why was my comment not addressed? We have the same limitation with Vault name and never truncate it, as that could result in name conflicts. We exclude environment name in preview, as the product prefix already provides the environment information.

KarolJastrzebski: Sorry for that, will do it today.

PR #10, opened 2018-06-21
Let account name overflow 24-character limit

KarolJastrzebski:
In the spirit of failing fast, our team had decided to not trim down
lengthy names. Implementing this principle to keep things consistent.

This also addresses the comment made by @lgonczar  in #8 

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -6,13 +6,14 @@ provider "vault" {
 
 locals {
   app = "bulk-scanning"
+  is_preview = "${(var.env == "preview" || var.env == "spreview")}"
   preview_account_name = "${var.product}bulkscan"
   default_account_name = "${var.product}bulkscan${var.env}"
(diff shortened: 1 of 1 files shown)


infrastructure/main.tf
```diff
@@ -49,7 +49,7 @@ resource "azurerm_resource_group" "rg" {
 }
 
 resource "azurerm_storage_account" "provider" {
-  name                      = "${substr(local.account_name, 0, min(length(local.account_name), 24))}"
+  name                      = "${local.account_name}"
```
lgonczar: Could you do it like this: https://github.com/hmcts/draft-store/blob/master/infrastructure/main.tf#L29 ?

When in preview, we don't need to append the environment bit, which is most likely to make the name too long, given the product gets the `PR_123` prefix.

PR #11, opened 2018-06-21
Add functional tests for SAS Token endpoint

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-493

### Change description ###

- Added functional tests for SAS token.

- Tested by configuring test url pointing to local environment.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -14,7 +14,6 @@ locals {
   previewVaultName     = "${var.product}-bulk-scan"
   nonPreviewVaultName  = "${var.product}-bulk-scan-${var.env}"
   vaultName            = "${local.is_preview ? local.previewVaultName : local.nonPreviewVaultName}"
-  tokenValidity        = "${var.token_validity}"
 }
 
 module "bulk-scan" {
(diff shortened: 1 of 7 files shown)


infrastructure/main.tf
```diff
@@ -27,9 +28,9 @@ module "bulk-scan" {
   capacity     = "${var.capacity}"
 
   app_settings = {
-    STORAGE_ACCOUNT_NAME = "${azurerm_storage_account.provider.name}"
-    STORAGE_KEY          = "${azurerm_storage_account.provider.primary_access_key}"
-
+    STORAGE_ACCOUNT_NAME    = "${azurerm_storage_account.provider.name}"
+    STORAGE_KEY             = "${azurerm_storage_account.provider.primary_access_key}"
+    SSCS_SAS_TOKEN_VALIDITY = "${local.tokenValidity}"
```
KarolJastrzebski: The name suggests that the variable is related to SSCS only. Is that the case?

infrastructure/main.tf
```diff
@@ -14,6 +14,7 @@ locals {
   previewVaultName     = "${var.product}-bulk-scan"
   nonPreviewVaultName  = "${var.product}-bulk-scan-${var.env}"
   vaultName            = "${local.is_preview ? local.previewVaultName : local.nonPreviewVaultName}"
+  tokenValidity        = "${var.token_validity}"
```
KarolJastrzebski: This variable copies the original value and brings no added value. Why not to use `${var.token_validity}` instead?

infrastructure/main.tf
nitinprabhu: yes will update variable name.

PR #13, opened 2018-06-28
Update dependencies

doncem:
### Change description ###

Notes:

- hystrix library was incompatibly used with spring boot version
- `azure-storage` currently has v8. not updating major jump and this can be done (suggesting to do) in separate pr together with reading task

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -125,45 +125,21 @@ dependencyCheck {
 
 repositories {
   jcenter()
-  // remove once spring sorts out their cloud artifact to proper version
+
   maven {
-    url "https://repo.spring.io/libs-milestone"
+    url 'https://dl.bintray.com/palantir/releases' // docker-compose-rule is published on bintray
```
nitinprabhu: I think we could remove this docker compose rule and Azurite test later. Not sure what benefit we are getting from Azurite? Not part of this PR just commenting here so we don't miss it later.

doncem: would be good. locally integration tests did not pass 😕

PR #16, opened 2018-06-29
Feature: parse metadata file

doncem:

[Create metadata records in database](https://tools.hmcts.net/jira/browse/RPE-554)

### Change description ###

Converting json to entities

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EntityParser.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+@Service
+public class EntityParser {
```
nitinprabhu: Do you think this can be more of utility class than a service ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EntityParserTest.java
```diff
@@ -0,0 +1,25 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class EntityParserTest {
```
nitinprabhu: Could we add some more scenarios when say payment section is missing or non scalable item is missing ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EntityParser.java
doncem: Started like that but then wanted to have the benefit of already assigned `ObjectMapper` in spring boot bean collection avoiding the `Autowire` annotation

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
```diff
@@ -0,0 +1,90 @@
...
+
+@Entity
+@Table(name = "envelopes")
+public class Envelope {
+
+    @Id
+    private UUID id;
+
+    @JsonProperty("po_box")
+    private String poBox;
```
rkondratowicz: what do those annotations do on private fields?

doncem: Works on private fields as well. Is not part of the work but used all annotations to prove locally the parsing does what it does. Simply left as a no harmer since proxy (?) object is already with json property constructors. Getters and setters will simply fall into same json object

PR #19, opened 2018-07-02
Read PDFs from azure container

KarolJastrzebski:
https://tools.hmcts.net/jira/browse/RPE-553

This is an MVP implementation where things are to be added, like parsing zipped files that is done in #16 .

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -44,3 +44,6 @@ accesstoken:
   serviceConfig:
     - serviceName: sscs
       validity: ${SAS_TOKEN_VALIDITY:300} #In seconds
+
+scan:
+  interval: ${SCAN_INTERVAL:30000}
```
rkondratowicz: worth adding unit of time imho

nitinprabhu: You will need to define `SCAN_INTERVAL` in terraform else code cannot be deployed.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Test
+    public void connects_without_exceptions() throws URISyntaxException, StorageException {
+        CloudBlobContainer abc = cloudBlobClient.getContainerReference("abc");
+        abc.createIfNotExists();
+
+        BlobStorageRead blobStorageRead = new BlobStorageRead(cloudBlobClient, pdfsConsumer);
+        blobStorageRead.readBlobs();
+    }
```
rkondratowicz: can you also add a test that checks whether `readBlobs` does what it should do?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/PdfsConsumer.java
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Consumes list of PDFs from an envelope (single zip file).
+ */
+@Component
+public class PdfsConsumer implements Consumer<List<PDF>> {
```
nitinprabhu: Why we need a consumer here ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+        this.cloudBlobClient = cloudBlobClient;
+        this.pdfsConsumer = pdfsConsumer;
+    }
+
+    @Scheduled(fixedDelayString = "${scan.interval}")
+    public void readBlobs() {
+        cloudBlobClient.listContainers().forEach(cloudBlobContainer -> {
+            try {
+                String containerName = cloudBlobContainer.getName();
+                CloudBlobContainer container = cloudBlobClient.getContainerReference(containerName);
```
rkondratowicz: confusing names imho. what's the type of `cloudBlobContainer` variable passed in this lambda?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+                readBlobsFromContainer(container);
+            } catch (URISyntaxException e) {
+                LOGGER.warn("Invalid URL", e);
```
nitinprabhu: may be put TO-DO comments here that it's going to be logged into app insights

nitinprabhu: same for others

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+            } catch (StorageException e) {
+                LOGGER.warn("Could not obtain container reference", e);
```
rkondratowicz: are you sure those should be only warnings?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+    public BlobStorageRead(CloudBlobClient cloudBlobClient, Consumer<List<PDF>> pdfsConsumer) {
+        this.cloudBlobClient = cloudBlobClient;
+        this.pdfsConsumer = pdfsConsumer;
+    }
+
+    @Scheduled(fixedDelayString = "${scan.interval}")
+    public void readBlobs() {
+        cloudBlobClient.listContainers().forEach(cloudBlobContainer -> {
+            try {
+                String containerName = cloudBlobContainer.getName();
```
nitinprabhu: Please log some info stating which container you are processing.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
```diff
@@ -0,0 +1,49 @@
...
+        cloudBlobClient = account.createCloudBlobClient();
+    }
+
+    @Test
+    public void connects_without_exceptions() throws URISyntaxException, StorageException {
+        CloudBlobContainer abc = cloudBlobClient.getContainerReference("abc");
+        abc.createIfNotExists();
+
+        BlobStorageRead blobStorageRead = new BlobStorageRead(cloudBlobClient, pdfsConsumer);
+        blobStorageRead.readBlobs();
```
nitinprabhu: No asserts ? May be put something that no exception occurred.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.stream.StreamSupport;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+@Component
+public class BlobStorageRead {
```
nitinprabhu: Can it be renamed to `DocumentProcessor` or something else ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
rkondratowicz: extra points for javadoc comment ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
rkondratowicz: extra points for javadoc comment ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+
+                    return processZipFile(blobInputStream);
+                } catch (StorageException e) {
+                    LOGGER.warn("Could not download data", e);
+                } catch (URISyntaxException e) {
+                    LOGGER.warn("Invalid URL", e);
+                }
+                return null;
+            })
+            .filter(Objects::nonNull)
```
rkondratowicz: you may want to filter out empty lists as well

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+            .filter(cloudBlob -> cloudBlob.getName().toLowerCase().endsWith(".zip"))
+            // TODO: notify supplier about non-zip files
+            .map(cloudBlob -> {
+                try {
+                    BlobInputStream blobInputStream = container.getBlockBlobReference(cloudBlob.getName())
+                        .openInputStream();
+
+                    return processZipFile(blobInputStream);
+                } catch (StorageException e) {
+                    LOGGER.warn("Could not download data", e);
```
lgonczar: Same here - why are these only warnings?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+                    LOGGER.warn("Invalid URL", e);
+                }
+                return null;
+            })
+            .filter(Objects::nonNull)
+            .forEach(pdfsConsumer);
+    }
+
+    private List<PDF> processZipFile(BlobInputStream blobInputStream) {
+        List<PDF> pdfs = new ArrayList<>();
```
lgonczar: Alternatively, you could use Stream.generate() for a more declarative approach. Up to you.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+
+        try (ZipInputStream zis = new ZipInputStream(blobInputStream)) {
```
lgonczar: Do we know how big those zips are going to be?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -0,0 +1,92 @@
...
+            } catch (StorageException e) {
+                LOGGER.warn("Could not obtain container reference", e);
+            }
+        });
+    }
+
+    private void readBlobsFromContainer(CloudBlobContainer container) {
+        StreamSupport.stream(container.listBlobs().spliterator(), false)
+            .filter(CloudBlob.class::isInstance)
+            .map(CloudBlob.class::cast)
```
rkondratowicz: would be worth adding a comment on those two lines imho

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
lgonczar: Are there unit tests for that class?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.hmcts.reform.bulkscanprocessor.services.BlobStorageRead;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.PDF;
+
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+import java.util.List;
+import java.util.function.Consumer;
+
+@RunWith(MockitoJUnitRunner.class)
+public class BlobStorageReadTest {
```
lgonczar: Where will these tests be executed?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
KarolJastrzebski: Not at all and would be keen to escalate to error level.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
```diff
@@ -43,7 +59,35 @@ public void connects_without_exceptions() throws URISyntaxException, StorageExce
         CloudBlobContainer abc = cloudBlobClient.getContainerReference("abc");
         abc.createIfNotExists();
 
+        Throwable throwable = catchThrowable(() -> {
+            BlobStorageRead blobStorageRead = new BlobStorageRead(cloudBlobClient, pdfsConsumer);
+            blobStorageRead.readBlobs();
+        });
+
+        assertThat(throwable).doesNotThrowAnyException();
```
lgonczar: You have to put the lambda as the argument, not the result. And shouldn't it be `assertThatCode(...`?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/BlobStorageReadTest.java
lgonczar: We still need unit tests, covering all the edge cases

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+            .forEach(pdfsConsumer);
+    }
+
+    private List<PDF> processZipFile(BlobInputStream blobInputStream) {
+        List<PDF> pdfs = new ArrayList<>();
+
+        try (ZipInputStream zis = new ZipInputStream(blobInputStream)) {
+            ZipEntry ze;
+            while ((ze = zis.getNextEntry()) != null) {
+                PDF pdf = new PDF(ze.getName(), ByteStreams.toByteArray(zis));
```
rkondratowicz: will the zip file contain only PDFs?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
```diff
@@ -42,12 +42,15 @@ public void readBlobs() {
         cloudBlobClient.listContainers().forEach(cloudBlobContainer -> {
             try {
                 String containerName = cloudBlobContainer.getName();
+                LOGGER.info("Processing {} container", containerName);
                 CloudBlobContainer container = cloudBlobClient.getContainerReference(containerName);
                 readBlobsFromContainer(container);
             } catch (URISyntaxException e) {
                 LOGGER.warn("Invalid URL", e);
```
lgonczar: Why is this exception not logged as error?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
KarolJastrzebski: In this ticket we are only reading PDFs.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/DocumentProcessorTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        verify(pdfsConsumer, times(1)).accept(pdfListCaptor.capture());
+
+        assertThat(pdfListCaptor.getAllValues()).hasSize(1);
+
+        assertThat(pdfListCaptor.getValue())
+            .hasSize(3)
+            .anyMatch(pdf -> "abc.pdf".equals(pdf.getFilename()))
+            .anyMatch(pdf -> "def.pdf".equals(pdf.getFilename()))
+            .anyMatch(pdf -> "ghi.pdf".equals(pdf.getFilename()));
```
rkondratowicz: you can replace the 4 lines above with `.containsExactlyInAnyOrder(...)`.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
lgonczar: In my opinion, the benefit would be simplicity. Not that important, though.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
lgonczar: Do you know how big individual zip files can be? If they're too big, we can't extract the whole file at once in memory.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/DocumentProcessorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    public void connects_without_exceptions() throws URISyntaxException, StorageException {
+        CloudBlobContainer abc = cloudBlobClient.getContainerReference("abc");
+        abc.createIfNotExists();
+
+        Throwable throwable = catchThrowable(() -> {
+            DocumentProcessor documentProcessor = new DocumentProcessor(cloudBlobClient, pdfsConsumer);
+            documentProcessor.readBlobs();
+        });
+
+        assertThat(throwable).doesNotThrowAnyException();
```
lgonczar: You have to put the lambda as the argument, not the result. And shouldn't it be assertThatCode(...?

Or expect the throwable to be null.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Scheduled(fixedDelayString = "${scan.interval}")
+    public void readBlobs() {
+        cloudBlobClient.listContainers().forEach(cloudBlobContainer -> {
+            try {
+                String containerName = cloudBlobContainer.getName();
+                LOGGER.info("Processing {} container", containerName);
+                CloudBlobContainer container = cloudBlobClient.getContainerReference(containerName);
+                readBlobsFromContainer(container);
+            } catch (URISyntaxException e) {
```
lgonczar: Why is this exception not logged as error?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
lgonczar: Oh, and if those zips (or their whole decompressed content) turn out to be too large to be kept in memory, current way won't work, whereas the streamed one will allow for processing the PDFs one by one, without having to extract them all at once.

KarolJastrzebski: Stream.generate() is an infinite stream and requires limiting. I have had a peek at `ZipFile.streams()` which utilises `Spliterators.spliterator()` and `ZipEntryIterator` where both of them require passing exact number of elements.

Of course, there is always `Stream.builder()` option, but it just another way for building the same array of elements. This brings another idea of converting `BlobInputStream` into `Iterator<PDF>`.

I am keen to hear your thoughts on this.

lgonczar: We can just terminate the stream when there's nothing left.

Whether it's a stream or an iterator, doesn't matter. What's important is that we don't read the whole content of the zip file into memory if we don't have to. The agreed limit is already quite big (4*75 MB). If someone uploads a file that's too big to fit in memory, the job will keep failing as long as that file is there.

infrastructure/main.tf
```diff
@@ -51,6 +51,7 @@ module "bulk-scan" {
     FLYWAY_USER                   = "${module.bulk-scan-db.user_name}"
     FLYWAY_PASSWORD               = "${module.bulk-scan-db.postgresql_password}"
     //
+    SCAN_INTERVAL                 = "30000"
```
lgonczar: Currently, this is the delay between the end of the last job execution and the start of the next. It would be best to either rename it to `SCAN_DELAY` (or sth similar) or set `fixedRateString` in `@Scheduled` annotation.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private void readBlobsFromContainer(CloudBlobContainer container) {
+        StreamSupport.stream(container.listBlobs().spliterator(), false)
+            .filter(CloudBlob.class::isInstance)
+            .map(CloudBlob.class::cast)
+            .filter(cloudBlob -> cloudBlob.getName().toLowerCase().endsWith(".zip"))
+            // TODO: notify supplier about non-zip files
+            .map(cloudBlob -> {
+                try {
+                    BlobInputStream blobInputStream = container.getBlockBlobReference(cloudBlob.getName())
```
lgonczar: This should go to try-with-resources, as the stream always has to be closed. Sorry for not mentioning earlier - didn't spot it before.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/PdfsConsumer.java
lgonczar: I think this class is not used anymore

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/BlobStorageRead.java
KarolJastrzebski: I like this but it looks like premature optimisation. Let's address this in separate ticket.

lgonczar: Not premature optimisation - we're reading potentially large files into memory and we can see how some input can cause this job to fail. I'm ok with addressing it in another pull request, though.

infrastructure/main.tf
```diff
@@ -51,6 +51,7 @@ module "bulk-scan" {
     FLYWAY_USER                   = "${module.bulk-scan-db.user_name}"
     FLYWAY_PASSWORD               = "${module.bulk-scan-db.postgresql_password}"
     //
+    SCAN_DELAY                    = "30000"
```
nitinprabhu: we can configure this in `variables.tf` ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/HandleZipTestDummy.java
```diff
@@ -0,0 +1,35 @@
...
+        assertThat(pdfNames).containsExactlyInAnyOrder(fileNames);
+    }
+
+    void jurisdictionNameIs(String name) {
+        assertThat(jurisdiction).isEqualTo(name);
+    }
+
+    void isInvoked() {
+        assertThat(invoked).isTrue();
+    }
```
rkondratowicz: As mentioned before, let's keep assertions in test. All those method are only called in one place.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/DocumentProcessorTest.java
```diff
@@ -0,0 +1,75 @@
...
+        CloudBlockBlob blockBlobReference = abc.getBlockBlobReference(testZip);
+        blockBlobReference.uploadFromFile(new File("src/integrationTest/resources/" + testZip).getAbsolutePath());
+
+        HandleZipTestDummy dummy = new HandleZipTestDummy();
+
+        DocumentProcessor documentProcessor = new DocumentProcessor(cloudBlobClient, dummy);
+        documentProcessor.readBlobs();
+
+        dummy.isInvoked();
+        dummy.jurisdictionNameIs("abc");
```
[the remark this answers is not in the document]
lgonczar: I agree. Naturally you would expect `dummy.isInvoked()` to just return a boolean, not that an assertion will be made. The straight assertion equivalents are short and readable.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.function.BiConsumer;
+import java.util.stream.StreamSupport;
+import java.util.zip.ZipFile;
+
+/**
+ * Main component that is run in intervals.
+ *
+ * Looks up all available containers for zip files and extracts pdf files from them.
```
nitinprabhu: This class is a processor which reads file from blob storage, extracts zip and stores them in document storage.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DocumentProcessor.class);
+
+    private final CloudBlobClient cloudBlobClient;
+    private final BiConsumer<String, ZipFile> handleZip;
+
+    /**
+     * Create new document processor for zip files stored on a blob storage.
+     *
+     * @param cloudBlobClient connection to Azure Blob Storage
```
nitinprabhu: Do we need these comments ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param handleZip callback function invoked for every downloaded zip file.
+     *          See {@link ZipFileHandler#accept(String, ZipFile)}.
+     */
+    @Autowired
+    public DocumentProcessor(CloudBlobClient cloudBlobClient, BiConsumer<String, ZipFile> handleZip) {
```
nitinprabhu: Not sure why we need a BiConsumer.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+     */
+    @Scheduled(fixedDelayString = "${scan.delay}")
+    public void readBlobs() {
+        cloudBlobClient.listContainers().forEach(cloudBlobContainer -> {
+            try {
+                String containerName = cloudBlobContainer.getName();
+                LOGGER.info("Processing {} container", containerName);
+                CloudBlobContainer container = cloudBlobClient.getContainerReference(containerName);
+                readBlobsFromContainer(container);
+            } catch (URISyntaxException e) {
```
nitinprabhu: Can we use Exceptions with Improved Type Checking here ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private void readBlobsFromContainer(CloudBlobContainer container) {
+        StreamSupport.stream(container.listBlobs().spliterator(), false)
+
+            // checking if it is safe to cast to CloudBlob
+            .filter(CloudBlob.class::isInstance)
+            .map(CloudBlob.class::cast)
+
+            // so that we could get method for retrieving its name
+            .filter(cloudBlob -> cloudBlob.getName().toLowerCase().endsWith(".zip"))
```
nitinprabhu: we should not filter zip if we get a non zip exception should be thrown and record should be created in db.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+            // TODO: notify supplier about non-zip files
+            .forEach(cloudBlob -> {
+                File tempFile = null;
+                try {
+                    // let's download every zip file using temporary file
+                    tempFile = File.createTempFile(cloudBlob.getContainer().getName(), "");
```
nitinprabhu: We could use streams instead of storing it to disk and fetching it again back before sending it to doc store. IO intensive operation. Do we really need to do this?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ZipFileHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    public void accept(String jurisdiction, ZipFile zipFile) {
+        try {
+            // extract data from manifest file
+            ZipEntry entry = zipFile.getEntry("manifest.json");
+            InputStream inputStream = zipFile.getInputStream(entry);
+            Envelope envelope = EntityParser.parseEnvelopeMetadata(inputStream);
+
+            // check if all files in the zip are present in the manifest
+            zipFile.stream()
+                .filter(file -> !file.getName().toLowerCase().equals("manifest.json"))
```
nitinprabhu: file will be called `metadata.json`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * Looks up all available containers for zip files and extracts pdf files from them.
+ */
+@Component
+public class DocumentProcessor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DocumentProcessor.class);
+
+    private final CloudBlobClient cloudBlobClient;
+    private final BiConsumer<String, ZipFile> handleZip;
```
lgonczar: Not that we've got more arguments and everything, maybe it would be better to use a problem-specific interface? By looking at the code you don't know what the string parameter represents.

Also, do we know what the handler will do and need?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
lgonczar: but that would mean the presence of a non-zip file will prevent the processing of other files

nitinprabhu: exception should be thrown while processing a non zip file and other files should continue processing. I would expect exception to be thrown while creating ZipFile object ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
```diff
@@ -0,0 +1,104 @@
...
+                LOGGER.error("Invalid URL", e);
+                // TODO: track exception in AppInsights Telemetry
+            } catch (StorageException e) {
+                LOGGER.error("Could not obtain container reference", e);
+                // TODO: track exception in AppInsights Telemetry
+            }
+        });
+    }
+
+    private void readBlobsFromContainer(CloudBlobContainer container) {
```
lgonczar: Worth extracting some logic out of this method - it's grown quite big.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ZipFileHandler.java
```diff
@@ -0,0 +1,48 @@
...
+public class ZipFileHandler implements BiConsumer<String, ZipFile> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZipFileHandler.class);
+
+    @Override
+    public void accept(String jurisdiction, ZipFile zipFile) {
+        try {
+            // extract data from manifest file
+            ZipEntry entry = zipFile.getEntry("manifest.json");
+            InputStream inputStream = zipFile.getInputStream(entry);
```
lgonczar: How about `manitestStream` or sth?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ZipFileHandler.java
```diff
@@ -0,0 +1,48 @@
...
+            Envelope envelope = EntityParser.parseEnvelopeMetadata(inputStream);
+
+            // check if all files in the zip are present in the manifest
+            zipFile.stream()
+                .filter(file -> !file.getName().toLowerCase().equals("manifest.json"))
+                .forEach(file -> {
+                    if (!envelope.hasReferenceTo(file.getName())) {
+                        throw new NotImplementedException();
```
lgonczar: What's expected here?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/DocumentProcessor.java
lgonczar: True. Provided it won't throw a RuntimeException (which it shouldn't)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ZipFileHandler.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+/**
+ * Instance of this class is automatically injected by Spring into
+ * {@link DocumentProcessor#DocumentProcessor(CloudBlobClient, BiConsumer)} as it implements
+ * {@link BiConsumer}.
+ */
+@Component
+public class ZipFileHandler implements BiConsumer<String, ZipFile> {
```
lgonczar: Needs to be tested

KarolJastrzebski:
Closing this one as @nitinprabhu is getting his take with the code.

PR #27, opened 2018-07-05
Add vault properties for db migration and updated DB name/user

nitinprabhu:
### Change description ###

- Vault name and micro service both needs to set as an environment variable as it is used by DB migration script.
-Also updated incorrect db name and user.

```
[ ] Yes
[ x] No
```


infrastructure/main.tf
```diff
@@ -29,8 +29,8 @@ module "bulk-scan-db" {
   product             = "${var.product}-${var.component}-db"
   location            = "${var.location_db}"
   env                 = "${var.env}"
-  database_name       = "feature_toggle"
-  postgresql_user     = "feature_toggler"
+  database_name       = "bulk_scan"
+  postgresql_user     = "bulk_scanner"
```
doncem: 😇

infrastructure/output.tf
```diff
@@ -1,4 +1,11 @@
...
   value = "${var.deployment_namespace}"
 }
+
+output "vaultName" {
+  value = "${local.vaultName}"
+}
+
+output "microserviceName" {
+  value = "${var.component}"
+}
```
doncem: 🤦‍♂️

PR #30, opened 2018-07-05
Distributed lock

KarolJastrzebski:
This is an integration with [ShedLock](https://github.com/lukas-krecan/ShedLock/) to prevent concurrency problems when running the job.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ShedLockConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ShedLockConfiguration.java
@@ -1,0 +1,29 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.config;
+
+import net.javacrumbs.shedlock.core.LockProvider;
+import net.javacrumbs.shedlock.provider.jdbc.JdbcLockProvider;
+import net.javacrumbs.shedlock.spring.ScheduledLockConfiguration;
+import net.javacrumbs.shedlock.spring.ScheduledLockConfigurationBuilder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import java.time.Duration;
+import javax.sql.DataSource;
+
+@Configuration
+public class ShedLockConfiguration {
+
+    @Bean
+    public LockProvider lockProvider(DataSource dataSource) {
+        return new JdbcLockProvider(dataSource);
+    }
+
+    @Bean
+    public ScheduledLockConfiguration taskScheduler(LockProvider lockProvider) {
+        return ScheduledLockConfigurationBuilder
+            .withLockProvider(lockProvider)
+            .withPoolSize(10)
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ShedLockConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+    @Bean
+    public LockProvider lockProvider(DataSource dataSource) {
+        return new JdbcLockProvider(dataSource);
+    }
+
+    @Bean
+    public ScheduledLockConfiguration taskScheduler(LockProvider lockProvider) {
+        return ScheduledLockConfigurationBuilder
+            .withLockProvider(lockProvider)
+            .withPoolSize(10)
```
nitinprabhu: Would be nice if we make this configurable.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ShedLockConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+@DependsOn({"flyway", "flywayInitializer"})
+public class ShedLockConfiguration {
+
+    @Bean
+    public LockProvider lockProvider(DataSource dataSource) {
+        return new JdbcLockProvider(dataSource);
+    }
+
+    @Bean
+    public ScheduledLockConfiguration taskScheduler(LockProvider lockProvider) {
```
nitinprabhu: Could we add a test case to validate this ?

src/main/resources/application.yaml
```diff
@@ -57,6 +57,8 @@ dm:
 
 scheduling:
   enabled: ${SCHEDULING_ENABLED:false}
+  pool: 10
+  lockAtMostFor: PT10M # 10 minutes
```
rkondratowicz: those 2 could be env vars (with defaults)

src/main/resources/application.yaml
```diff
@@ -61,6 +61,8 @@ dm:
 
 scheduling:
   enabled: ${SCHEDULING_ENABLED:false}
+  pool: ${SCHEDULING_POOL:10}
+  lockAtMostFor: ${SCHEDULING_LOCKATMOSTFOR:PT10M} # 10 minutes
```
rkondratowicz: it would be good to add a comment on what the format here is (unless sample `PT10M` is clear to everyone...) imho

rkondratowicz: how about `LOCK_AT_MOST_FOR` instead of `LOCKATMOSTFOR`?

KarolJastrzebski: I don't mind either way. Taken that approach after reading `org.springframework.core.env.SystemEnvironmentPropertySource#checkPropertyName` implementation so that `scheduling.lockAtMostFor` gets resolved to `SCHEDULING_LOCKATMOSTFOR`.

True that, perhaps too defensive that ended up too obscure. Adding underscores.

PR #35, opened 2018-07-09
Bugfix: JPA entities

doncem:

[Persist child entities correctly](https://tools.hmcts.net/jira/browse/RPE-572)

### Change description ###

Notes:
- Envelope ID was never saved due to entity not being assigned to each child element [major fix]
- Moving transactional scope to source code [minor fix]. It allows to trust test cases are using DB for retrieving entities rather than in-memory built-up proxies

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItemTest.java
```diff
@@ -37,7 +37,7 @@ public void should_update_document_url_of_scannable_item() throws IOException {
 
         // when
         scannableItemRepository.saveAll(items.stream()
-            .peek(item -> item.setDocumentUrl("localhost/document/" + item.getId()))
+            .peek(item -> item.setDocumentUrl("localhost/document/" + item.hashCode()))
```
nitinprabhu: Document Id is UUID and not hashcode :)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItemTest.java
```diff
@@ -50,7 +50,7 @@ public void should_update_document_url_of_scannable_item() throws IOException {
 
         // and
         dbItems.forEach(item ->
-            assertThat(item.getDocumentUrl()).isEqualTo("localhost/document/" + item.getId())
+            assertThat(item.getDocumentUrl()).isEqualTo("localhost/document/" + item.hashCode())
```
nitinprabhu: same goes here

PR #37, opened 2018-07-11
Clean up token generation

rkondratowicz:
Tiny improvements:
- 1 bean instead of 3 in configuration
- simpler expiry date setting

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasTokenGeneratorService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasTokenGeneratorService.java
@@ -2,8 +2,6 @@ package uk.gov.hmcts.reform.bulkscanprocessor.services;
 
 import com.microsoft.azure.storage.StorageException;
 import com.microsoft.azure.storage.blob.CloudBlobClient;
-import com.microsoft.azure.storage.blob.CloudBlobContainer;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasTokenGeneratorService.java
```diff
@@ -55,13 +56,9 @@ public String generateSasToken(String serviceName) {
...
 
-        Calendar cal = new GregorianCalendar(TimeZone.getTimeZone("UTC"));
-        cal.setTime(new Date());
-        cal.add(Calendar.SECOND, config.getValidity());
-
         SharedAccessBlobPolicy policy = new SharedAccessBlobPolicy();
-        policy.setPermissions(getBlobPermissions());
-        policy.setSharedAccessExpiryTime(cal.getTime());
+        policy.setPermissions(EnumSet.of(WRITE, LIST));
+        policy.setSharedAccessExpiryTime(Date.from(now().plusSeconds(config.getValidity()).toInstant(ZoneOffset.UTC)));
```
doncem: I think `LocalDateTime` should also have `.toInstant` method so `Date.from` becomes irrelevant here ;)

rkondratowicz: `.toInstant` **is** called on LocalDateTime.
There's a lot of parentheses in this line...

PR #38, opened 2018-07-11
Feature: envelope status

doncem:
### JIRA link (if applicable) ###

[Introduce envelope state](https://tools.hmcts.net/jira/browse/RPE-573)

### Change description ###

Creating envelope statuses table with single happy scenario

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #39, opened 2018-07-12
Add `UploadFailure` scenario

doncem:

### Change description ###

Notes:
- integration tests cover pretty much all the additions of this PR
- include message in exception of upload failure
- state integration was done in #38 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
@@ -1,0 +1,170 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
+
+import com.google.common.collect.ImmutableList;
+import com.microsoft.azure.storage.CloudStorageAccount;
+import com.microsoft.azure.storage.blob.CloudBlobClient;
+import com.microsoft.azure.storage.blob.CloudBlobContainer;
+import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import com.palantir.docker.compose.DockerComposeRule;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.util.ObjectUtils;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeRepository;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeState;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeStateRepository;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -108,10 +108,18 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
...
-            envelopeProcessor.markAsUploaded(envelope, container.getName(), zipFilename);
-
-            cloudBlockBlob.delete();
+            try {
+                documentProcessor.processPdfFiles(pdfFiles, envelope.getScannableItems());
+                envelopeProcessor.markAsUploaded(envelope, container.getName(), zipFilename);
+
+                cloudBlockBlob.delete();
+            } catch (Exception exception) {
+                // catching any exception originated from document processor
```
nitinprabhu: It would be nice if we could log envelope name along with exception just in case update to DB fails ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
```diff
@@ -0,0 +1,170 @@
...
+        assertThat(envelopeState)
+            .extracting("container", "zipFileName", "status")
+            .hasSameElementsAs(ImmutableList.of(testContainer.getName(), ZIP_FILE_NAME, DOC_UPLOAD_FAILURE));
+        assertThat(envelopeState.getId()).isNotNull();
+        assertThat(envelopeState.getReason()).isNotBlank();
+        assertThat(envelopeState.getEnvelope().getId()).isEqualTo(actualEnvelope.getId());
+    }
+
+    @Test
+    public void should_record_failure_of_upload_when_document_management_throws_exception() throws IOException {
```
nitinprabhu: Could we have a scenario where even DB update fails ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    private void uploadZipToBlobStore() throws Exception {
+        byte[] zipFile = toByteArray(getResource(ZIP_FILE_NAME));
+
+        CloudBlockBlob blockBlobReference = testContainer.getBlockBlobReference(ZIP_FILE_NAME);
+        blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length);
+    }
+
+    private List<Pdf> getUploadResources() throws IOException {
```
nitinprabhu: I think method like these could be extracted to a common base class instead of repeating it in each class ? Could be a separate PR for refactoring though.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: logging should come in together with exception handler up in another level. this additional "exception wrapping" is for recording states. left logging out of it intentionally

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
doncem: True. best to leave for later once all the work around it is done and you are able to see bigger picture. This function will also change in the next PR 😉

PR #40, opened 2018-07-13
Update: envelope pre upload failure status

doncem:

### Change description ###

Last set of tests on "new" generic failure state. Line of work:

- #38 - state groundbase
- #39 - upload failure scenario

Next to come:

- Migrate to new data structure of having process events instead of envelope states
- Clean up envelope states

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
@@ -35,13 +35,14 @@ import static com.google.common.io.Resources.getResource;
 import static com.google.common.io.Resources.toByteArray;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.BDDMockito.given;
+import static uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeStatus.DOC_FAILURE;
 import static uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeStatus.DOC_UPLOAD_FAILURE;
 
 @RunWith(SpringRunner.class)
 @SpringBootTest
 public class BlobProcessorTaskTestForFailedStatus {
 
-    private static final String ZIP_FILE_NAME = "1_24-06-2018-00-00-00.zip";
+    private static final String ZIP_FILE_NAME_SUCCESS = "1_24-06-2018-00-00-00.zip";
 
     @ClassRule
     public static DockerComposeRule docker = DockerComposeRule.builder()
@@ -87,8 +88,6 @@ public class BlobProcessorTaskTestForFailedStatus {
 
         testContainer = cloudBlobClient.getContainerReference("test");
         testContainer.createIfNotExists();
-
-        uploadZipToBlobStore();
     }
 
     @After
@@ -99,8 +98,11 @@ public class BlobProcessorTaskTestForFailedStatus {
     }
 
     @Test
-    public void should_record_failure_of_upload_when_document_management_returns_empty_response() throws IOException {
+    public void should_record_failure_of_upload_when_document_management_returns_empty_response() throws Exception {
         // given
+        uploadZipToBlobStore(ZIP_FILE_NAME_SUCCESS);
+
+        // and
         given(documentManagementService.uploadDocuments(getUploadResources())).willReturn(Collections.emptyMap());
 
         // when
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -108,20 +110,31 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
                 }
             }
 
-            Envelope envelope = envelopeProcessor.processEnvelope(metadataStream);
+            envelope = envelopeProcessor.processEnvelope(metadataStream);
+            isUploadFailure = true;
```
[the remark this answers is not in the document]
doncem: Initial investigation was checking out exceptions. But there might be a lot more in play - failing while using document management client, failing when interacting with db. Now whether it saved or not in db - another matter. Could be not allocating connection slot and throw/catch magically does. Nevertheless, not only `UnableToUploadDocumentException` is in play here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -108,20 +110,31 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
...
+        String zipFileName,
+        String message,
+        Envelope envelope
+    ) {
+        EnvelopeProcessor.FailureMarker marker = isUploadFailure
+            ? envelopeProcessor::markAsUploadFailed
+            : envelopeProcessor::markAsGenericFailure;
+
+        marker.markAsFailure(message, envelope, container, zipFileName);
+    }
```
rkondratowicz: imho it would be clearer just do:
```
if (isUploadFailure) {
  envelopeProcessor.markAsUploadFailure(...)
} else {
  envelopeProcessor.markAsGenericFailure(...)
}
```
where this function is called (and remove the functional interface).

Or use a `switch` statement if there are more failure types in the future.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -71,4 +77,15 @@ public void markAsUploadFailed(String reason, Envelope envelope, String containe
...
+    public void markAsGenericFailure(String reason, Envelope envelope, String container, String zipFileName) {
+        envelopeStateRepository.save(
+            EnvelopeStatusBuilder
+                .newEnvelopeStatus(container, zipFileName)
+                .withEnvelope(envelope)
+                .withStatus(DOC_FAILURE)
+                .withReason(reason)
+                .build()
+        );
+    }
```
rkondratowicz: Looks like we've got 3 almost identical methods now:
- `markAsUploaded`
- `markAsUploadFailed`
- `markAsGenericFailure`

Perhaps the common part could be extracted to a private method?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
```diff
@@ -145,16 +150,94 @@ public void should_record_failure_of_upload_when_document_management_throws_exce
...
         assertThat(envelopeState.getReason()).isEqualTo(throwable.getMessage());
         assertThat(envelopeState.getEnvelope().getId()).isEqualTo(actualEnvelope.getId());
     }
 
-    private void uploadZipToBlobStore() throws Exception {
-        byte[] zipFile = toByteArray(getResource(ZIP_FILE_NAME));
+    @Test
+    public void should_record_generic_failure_when_zip_does_not_contain_metadata_json() throws Exception {
+        // given
+        String noMetafileZip = "2_24-06-2018-00-00-00.zip";
```
rkondratowicz: does the file name need to be in this particular format?
ie, can the file be called something like `no-metafile.zip`?

doncem: it is most likely taken from the documentation spec:

> Zip file format : unqiueId_datetimecreated.zip

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
doncem: All generalisation can be done once we have a bigger scope of what's coming where from. Premature refactor is the root of evil. I am fully aware of similarity

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: No it is not. You have more than clear extraction of functional interface of what it's doing/what it's responsible of. Generally we mark the failure hence the method call. Signature is exactly the same and duplicating that in simple if/else call is simply introducing too many characters to read to clarify the correctness.

The correct implementation is in place disregarding the switch case. Ideally we would like to have general container of failure report and get rid of code branching entirely. Again - best to leave for later once bigger picture in mind

PR #42, opened 2018-07-14
New events table replacing state

doncem:

### Change description ###

As discussed on Friday, the `envelope state` sort of introduces confusion to developers. Renaming the entity to `process event` as it is recording events during processing, hence `bulk_scan_processing`. Next PR will drop the retired table and default value in envelopes

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
@@ -5,8 +5,8 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeRepository;
-import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeState;
-import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeStateRepository;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ProcessEvent;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ProcessEventRepository;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.MetadataNotFoundException;
 import uk.gov.hmcts.reform.bulkscanprocessor.util.EntityParser;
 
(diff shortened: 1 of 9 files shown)


src/main/resources/db/migration/V005__Process_events.sql
```diff
@@ -0,0 +1,25 @@
...
+CREATE INDEX process_events_container_event_idx ON process_events (container, event);
+CREATE INDEX process_events_zip_idx ON process_events (zipFileName);
+
+INSERT INTO process_events (container, zipFileName, createdAt, envelope_id, event, reason) (
+  SELECT container, zipFileName, createdAt, envelope_id, status, reason
+  FROM envelope_state
+);
+
+ALTER TABLE envelopes
+  ADD COLUMN lastEvent VARCHAR(100) NOT NULL DEFAULT 'ENVELOPE_CREATED';
```
nitinprabhu: When we are creating an envelope we are creating it with status ENVELOPE_CREATED but we are not putting any entries in process_events table?

src/main/resources/db/migration/V005__Process_events.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE process_events (
```
nitinprabhu: We are creating process_events table and then copying all values from envelope_state into process_events.Any reason why we are not just renaming envelope_state to process_events in migratiion script ?

src/main/resources/db/migration/V005__Process_events.sql
```diff
@@ -0,0 +1,25 @@
...
+  id BIGSERIAL PRIMARY KEY,
+  container VARCHAR(50) NOT NULL,
+  zipFileName VARCHAR(255) NOT NULL,
+  createdAt TIMESTAMP NOT NULL,
+  envelope_id VARCHAR(50) NULL,
```
nitinprabhu: Do we need envelope_id in process_events table ? Say we fail to process zip(generic failure) or we are creating event for AV failed envelope then when we create a record in process_events table we won't have envelope id. It will look like an orphan record ?

If we create a record in envelope table for above scenarios then it is fine to have envelope id as you can find them in both tables.

nitinprabhu: Is there a scenario where we will be retrieving records from event table based on envelope id ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -53,28 +53,37 @@ public Envelope processEnvelope(byte[] metadataStream) throws IOException {
...
+
+        event.setEnvelope(envelope);
+        event.setReason(reason);
 
-        state.setEnvelope(envelope);
-        state.setReason(reason);
+        processEventRepository.save(event);
 
-        envelopeStateRepository.save(state);
+        envelope.setLastEvent(DOC_UPLOAD_FAILURE);
```
nitinprabhu: If last state in envelope table for a envelope is `DOC_UPLOAD_FAILURE` then how would the Job scheduler task know how many times it needs to retry as envelope table doesn't have a date from which task can derive when it was first created.

src/main/resources/db/migration/V005__Process_events.sql
doncem: not backwards compatible. firstly additions, then deletion

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
nitinprabhu: As you mentioned in above comment there would be createdAt column in envelope so it should be derive the logic.

PR #45, opened 2018-07-17
 Integrated envelope controller with service auth

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-563

### Change description ###

- This PR integrates envelope retrieval with s2s.

- Using s2s service name we would retrieve jurisdiction based on the configuration mapping between service name and jurisdiction.

- Updated unit and integration tests.

- Functional tests would be in a separate PR.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
@@ -2,26 +2,51 @@ package uk.gov.hmcts.reform.bulkscanprocessor.services;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
```diff
@@ -2,26 +2,51 @@
...
+    }
 
-        return envelopeRepository.findAll();
+    private String getJurisdictionByServiceName(String serviceName) {
+        String jurisdiction = serviceJurisdictionMappingConfig
+            .getServicesJurisdiction()
+            .getOrDefault(serviceName, null);
+
+        if (jurisdiction == null) {
+            throw new ServiceJuridictionConfigNotFoundException(
```
doncem: This exception is not dealt with in controller advice

PR #46, opened 2018-07-17
Test deletion after successful processing

KarolJastrzebski:
https://tools.hmcts.net/jira/browse/RPE-514

```
[ ] Yes
[x] No
```


Jenkinsfile_CNP
```diff
@@ -8,10 +8,26 @@ def component = "bulk-scan-processor"
 
 def channel = '#rpe-build-notices'
 
+List<LinkedHashMap<String, Object>> secrets = [
+  secret('test-scan-delay', 'TEST_SCAN_DELAY'),
```
nitinprabhu: Do we really need these secrets as  you are using SAS token to upload to blob storage.

infrastructure/test_config.tf
```diff
@@ -0,0 +1,17 @@
+resource "azurerm_key_vault_secret" "test-scan-delay" {
```
nitinprabhu: Again same as above do we need these secrets ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,84 @@
...
+        JsonNode sasToken = objectNode.get("sas_token");
+        if (sasToken != null) {
+            return sasToken.asText();
+        }
+        throw new RuntimeException("Empty SAS token");
+    }
+
+    private void waitForBlobProcess() {
+        try {
+            Thread.sleep(scanDelay + 1000); // extra second for a good measure...
```
nitinprabhu: If there are no blobs then previous jobs should not take much time time to execute . We can just wait for say 5 seconds then check blob storage ?

src/functionalTest/resources/application.properties
```diff
@@ -1 +1,4 @@
 test-url=${TEST_URL}
+scan.delay=${TEST_SCAN_DELAY}
```
nitinprabhu: Same as above for all three properties

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/document/InMemoryMultipartFile.java
```diff
@@ -79,7 +79,9 @@ public InputStream getInputStream() throws IOException {
 
     @Override
     public void transferTo(File dest) throws IOException, IllegalStateException {
-        new FileOutputStream(dest).write(payload);
+        try (FileOutputStream stream = new FileOutputStream(dest)) {
```
nitinprabhu: Is this part of this PR ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,84 @@
...
+        CloudBlockBlob blockBlobReference = testContainer.getBlockBlobReference(zipName);
+        blockBlobReference.uploadFromFile(zipPath);
+
+        waitForBlobProcess();
+
+        assertThat(blockBlobReference.exists()).isFalse();
+    }
+
+    private String obtainSASToken() {
+        RestTemplate restTemplate = new RestTemplate();
```
nitinprabhu: I think you will need to relax https. Better to use Rest Assured as it will be consistent with other repos functional tests.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {StorageConfiguration.class})
+public class WhenProcessingEnvelopeTest {
```
nitinprabhu: Could we add another zip which doesn't get processed and hence blob won't be deleted ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void should_delete_zip_file_after_successful_ingestion()
+        throws URISyntaxException, StorageException, IOException {
+
+        // upload the file
+        String zipName = "4_24-06-2018-00-00-00.zip";
+        String zipPath = new File("src/integrationTest/resources/" + zipName).getAbsolutePath();
+
+        CloudBlockBlob blockBlobReference = testContainer.getBlockBlobReference(zipName);
+        blockBlobReference.uploadFromFile(zipPath);
```
nitinprabhu: we can do this as below 
```  
 byte[] zipFile = toByteArray(getResource(fileName));
 CloudBlockBlob blockBlobReference = testContainer.getBlockBlobReference(fileName);
 blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length);

```

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: Or, if there's a way of polling with an eventual timeout, that would be ideal.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: Why `When...`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/document/InMemoryMultipartFile.java
KarolJastrzebski: Sonar failed the build because of that.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
KarolJastrzebski: Thanks, changing to Rest Assured.

Jenkinsfile_CNP
KarolJastrzebski: True, thanks for spotting that.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
nitinprabhu: I think you will also need to configure your test container so that you could obtain SAS for that particular container else it will throw an error. Check https://github.com/hmcts/bulk-scan-processor/blob/2d0ace735c63a044b63ec226c95b37c4cb50bd4f/src/main/resources/application.yaml#L49

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
KarolJastrzebski: By default `scan.delay` is set to 30 seconds. This is a period that begins after the task was finished.

1. file is uploaded
2. here is the rest of the delay after last executio
3. blob processor is run
4. after successful processing, the file is deleted
5. fixed scan delay is reset

It is possible that the file gets uploaded during task execution or a second after it, and the file will be processed with the next run. Hence waiting for a fixed duration may give false-negative results, i.e. the file was not deleted because the test wasn't "patient" enough and didn't give application a chance to run the process.

I think it would be safe to assume that when file is deleted after a period shorter than `scan.delay`, the file was successfully processed and deleted as part of processing flow.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/document/InMemoryMultipartFile.java
rkondratowicz: it would be great to fix any sonar issues that are already there in the repo on a separate PR (and merge it first)

src/functionalTest/resources/application.properties
```diff
@@ -1 +1,4 @@
...
+storage.account_name=${TEST_STORAGE_ACCOUNT_NAME}
+storage.key=${TEST_STORAGE_KEY}
```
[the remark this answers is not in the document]
KarolJastrzebski: They are used by `StorageConfiguration` to create `CloudBlobClient`.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: What I meant was that in order to save those tests time, instead of waiting for the maximum possible amount of time, do it in the "wait at most N seconds for this condition to be met, polling every N milliseconds" style, which is possible here.

rkondratowicz: to clarify, here's a sample code that Leszek is talking about:
```
await()
  .atMost(30, SECONDS)
   .pollDelay(500, MILLISECONDS)
   .until(() -> checkWhetherSomethingExpectedHappened());
```
https://github.com/hmcts/private-beta-invitation-service/blob/master/src/functionalTest/java/uk/gov/hmcts/reform/pbis/e2e/SmokeTest.java#L57

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/document/InMemoryMultipartFile.java
KarolJastrzebski: Agreed #49

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
KarolJastrzebski: `getResources()` uses `functional` context, however, the zip file is in `integration` resources.

build.gradle
```diff
@@ -187,6 +187,7 @@ dependencies {
   integrationTestCompile group: 'com.revinate', name: 'assertj-json', version: '1.2.0'
 
   functionalTestCompile sourceSets.main.runtimeClasspath
+  functionalTestCompile sourceSets.test.runtimeClasspath
```
rkondratowicz: tests for tests?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,79 @@
...
+        testContainer = new CloudBlobContainer(PathUtility.addToQuery(containerUri, sasToken));
+    }
+
+    @Test
+    public void should_delete_zip_file_after_successful_ingestion() throws Exception {
+        CloudBlockBlob blockBlobReference = uploadZipFile("4_24-06-2018-00-00-00.zip");
+
+        await()
+            .atMost(scanDelay + 1000, TimeUnit.MILLISECONDS)
+            .pollDelay(5, TimeUnit.SECONDS)
```
rkondratowicz: Depending on `scanDelay` this could be longer than the `atMost` value.
IMHO it wouldn't hurt to check more often.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: given/when/then is typically represented by various clauses or sometimes even comments. Class names by convention represent the subject of testing, i.e. class/component or a piece of its functionality. This is the convention we're using everywhere.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: How about every second?

infrastructure/test_config.tf
```diff
@@ -0,0 +1,15 @@
...
+  depends_on = ["azurerm_storage_account.provider"]
+}
+
+resource "azurerm_key_vault_secret" "test-storage-account-name" {
+  name      = "test-storage-account-name"
+  value     = "${azurerm_storage_account.provider.name}"
+  vault_uri = "${module.bulk-scan-key-vault.key_vault_uri}"
+
+  depends_on = ["azurerm_storage_account.provider"]
+}
```
rkondratowicz: 'scan delay' and 'storage account name' are not really secret information.
Consider setting env vars for tests directly.
See https://github.com/hmcts/private-beta-invitation-service/blob/master/infrastructure/output.tf#L13 for reference

src/functionalTest/resources/application.properties
```diff
@@ -1 +1,3 @@
 test-url=${TEST_URL}
+scan.delay=${TEST_SCAN_DELAY}
+storage.account_name=${TEST_STORAGE_ACCOUNT_NAME}
```
rkondratowicz: no nesting/grouping here, so maybe just `-` instead of `.`?

infrastructure/test_config.tf
KarolJastrzebski: Thanks, I'll give it a go!

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,79 @@
...
+            .until(blockBlobReference::exists, is(false));
+    }
+
+    @Test
+    public void should_keep_zip_file_after_failed_processing() throws Exception {
+        CloudBlockBlob blockBlobReference = uploadZipFile("2_24-06-2018-00-00-00.zip");
+
+        await()
+            .timeout(scanDelay + 1000, TimeUnit.MILLISECONDS)
+            .until(blockBlobReference::exists, is(true));
```
lgonczar: If I understand correctly, this condition is met at the beginning, i.e. the blob exists. If that's the case, this test will always pass, regardless of the outcome of processing.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+import static com.jayway.awaitility.Awaitility.await;
+import static org.hamcrest.Matchers.is;
+
+@RunWith(SpringRunner.class)
+@TestPropertySource("classpath:application.properties")
+public class WhenProcessingEnvelopeTest {
+
+    @Value("${test-url}")
+    private transient String testUrl;
```
lgonczar: Why are these things transient?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
KarolJastrzebski: We upload invalid zip file and wait long enough to be sure that it was processed. Then we check that the file was not deleted.

lgonczar: We wait for this file to exist. As it exists since the beginning, the tests will pass without any waiting, no matter what the processing does with it later.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private transient int scanDelay;
+
+    @Value("${storage-account-name")
+    private transient String accountName;
+
+    private transient CloudBlobContainer testContainer;
+
+    @Before
+    public void setUp() throws Exception {
+        String serviceName = "test";
```
nitinprabhu: This is the container name ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,78 @@
...
+        testContainer = new CloudBlobContainer(PathUtility.addToQuery(containerUri, sasToken));
+    }
+
+    @Test
+    public void should_delete_zip_file_after_successful_ingestion() throws Exception {
+        CloudBlockBlob zipFile = uploadZipFile("4_24-06-2018-00-00-00.zip");
+
+        await()
+            .atMost(scanDelay + 1000, TimeUnit.MILLISECONDS)
+            .pollDelay(5, TimeUnit.SECONDS)
```
nitinprabhu: If we are using polling then do we really need scan delay ?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
```diff
@@ -0,0 +1,78 @@
...
+            .until(zipFile::exists, is(false));
+    }
+
+    @Test
+    public void should_keep_zip_file_after_failed_processing() throws Exception {
+        CloudBlockBlob zipFile = uploadZipFile("2_24-06-2018-00-00-00.zip");
```
nitinprabhu: worth specifying what's in the zip so when you look at the test case you know why it will fail

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
lgonczar: We still need to know when to time out.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
KarolJastrzebski: Yes. Renamed to `containerName`.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/WhenProcessingEnvelopeTest.java
nitinprabhu: The transient keyword in Java is used to indicate that a field should not be serialized. Why do we need it here that was the question and nothing else 😉

lgonczar: It's probably one of the annotations making it considered a bean. An beans need to be serializable. We might want to start using jUnitRunner instead. Currently, our tests are starting a Spring app, which for funtional/smoke tests is completely unnecessary.

In some of our projects we have already started using https://github.com/lightbend/config to read configuration without Spring's help.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
```diff
@@ -0,0 +1,78 @@
...
+    public void setUp() throws Exception {
+        String containerName = "test";
+        String sasToken = RestAssured
+            .given()
+            .relaxedHTTPSValidation()
+            .when()
+            .get(testUrl + "/token/" + containerName)
+            .thenReturn()
+            .jsonPath()
+            .getString("sas_token");
```
nitinprabhu: Will retrieving SAS token work without confusing test service in config ?

KarolJastrzebski: You should know better :grin:

KarolJastrzebski: Added `test` service to configuration.

src/main/resources/application.yaml
```diff
@@ -48,6 +48,8 @@ accesstoken:
   serviceConfig:
     - serviceName: sscs
       validity: ${SAS_TOKEN_VALIDITY:300} #In seconds
+    - serviceName: test
```
lgonczar: Should this value come from terraform, where test container is defined?

KarolJastrzebski: Agree, that would be ideal to have it declared only once. Refactoring for another PR?

lgonczar: It's actually a tiny change - all you need to add is an output variable that points to your azurerm_storage_container's name. And then in applycation.yaml refet the environment variable that Jenkins built from that output var.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
```diff
@@ -0,0 +1,84 @@
...
+            .until(() -> storageHasFile(zipFile.getName()), is(false));
+    }
+
+    @Test
+    public void should_keep_zip_file_after_failed_processing() throws Exception {
+        CloudBlockBlob zipFile = uploadZipFile("2_24-06-2018-00-00-00.zip"); // invalid due to missing json file
+
+        await()
+            .timeout(scanDelay + 1000, TimeUnit.MILLISECONDS)
+            .until(() -> storageHasFile(zipFile.getName()), is(true));
```
lgonczar: This `await()` still needs to be changed, as this test will always pass, no matter what happens during the processing. We only check if the zip file exists directly after uploading it. The `await()` part has nothing to wait for and will return instantly.

src/main/resources/application.yaml
KarolJastrzebski: If it goes for `test` service, the same should be applied for `sscs` and it becomes refactoring.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
KarolJastrzebski: We upload the file and wait at least `scanDelay + 1000 milliseconds` before we make any assertions. By then the file should be processed. If it was valid, it would have been deleted. But we know it was invalid and therefore should exist.

lgonczar: If the condition we are waiting for was not met, we would wait at most `scanDelay + 1000 milliseconds`, but as the condition is met at the beginning, we don't wait for anything. Replace the input file with a valid file and the test will still pass. `timout()` is the same as `atMost()`

KarolJastrzebski: Yes, you are right. Changing to `Thread.sleep()`. Thanks!

PR #50, opened 2018-07-19
Clean-up db [revisited]

doncem:
### Change description ###

Applying changes as per discussion earlier today:

- rename last event to state
- drop envelope from event

Adding container to envelope as it is conveniently helpful to track down duplicates for future status and in general - querying envelopes instead of events

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V007__Envelope_upgrade.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE process_events DROP COLUMN envelope_id;
+
+ALTER TABLE envelopes
+  ADD COLUMN container VARCHAR(50) NOT NULL DEFAULT 'sscs', -- absolutely not null. can drop default in next migration
```
rkondratowicz: I would normally do this:
1. add nullable column
2. update existing rows
3. make column NOT NULL

doncem: if you have something to update - perhaps. now it's enough to assign default. only one container is present too

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
```diff
@@ -125,15 +131,22 @@ public String getZipFileName() {
...
     }
 
     private void assignSelfToChildren(List<? extends EnvelopeAssignable> assignables) {
         assignables.forEach(assignable -> assignable.setEnvelope(this));
     }
+
+    @PrePersist
+    public void containerMustBePresent() {
+        if (container == null) {
+            throw new InvalidEnvelopeException("Missing required container");
```
nitinprabhu: I think we should not do this because if the container is not set that is problem with the code and not the envelope itself. Can we just log instead ? If we find this in DB that could be reconciled anytime but if we do not process the document than that won't be correct.

src/main/resources/db/migration/V007__Envelope_upgrade.sql
rkondratowicz: there should never be a container assigned by default if not provided by the app

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
```diff
@@ -46,8 +49,7 @@
     private Timestamp zipFileCreatedDate;
     @JsonProperty("zip_file_name")
     private String zipFileName;
-    @JsonProperty("last_event")
-    private Event lastEvent = Event.ENVELOPE_CREATED;
+    private Event status = Event.ENVELOPE_CREATED;
```
nitinprabhu: Should status should marked as JsonIgnore as we do not want to send this back to services ?

PR #63, opened 2018-07-26
Add service to validate if given envelope can be updated by given service

rkondratowicz:
it will be used by the service that makes the updates in next PR.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessService.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.services;
 
+import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.EnvelopeAccessProperties;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ForbiddenException;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ServiceConfigNotFoundException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessService.java
```diff
@@ -0,0 +1,42 @@
...
+public class EnvelopeAccessService {
+
+    private final EnvelopeAccessProperties access;
+
+    public EnvelopeAccessService(EnvelopeAccessProperties accessProps) {
+        this.access = accessProps;
+    }
+
+    /**
+     * Checks whether envelope from given jurisdiction can be update by given service.
```
nitinprabhu: update -> updated ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessService.java
```diff
@@ -0,0 +1,42 @@
...
+     * Throws an exception if service is not allowed to make and update
```
nitinprabhu: make and update -> make an update ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessService.java
```diff
@@ -0,0 +1,42 @@
...
+     * or configuration for the jurisdiction is not found.
+     */
+    public void assertCanUpdate(String envelopeJurisdiction, String serviceName) {
```
nitinprabhu: Do we want to log this info envelopeJurisdiction and serviceName ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ForbiddenException;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ServiceConfigNotFoundException;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+@RunWith(MockitoJUnitRunner.class)
+public class EnvelopeAccessServiceTest {
```
nitinprabhu: Should we add a test case which verifies that no exception is thrown when service is allowed to update envelopes ?

PR #75, opened 2018-07-30
Validate status transition

rkondratowicz:
Check whether we can change envelope's status to `X` when it is currently in status `Y`

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeStatusChangeValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeStatusChangeValidator.java
@@ -28,8 +28,7 @@ public class EnvelopeStatusChangeValidator {
         boolean ok =
             allowedTransitions
                 .getOrDefault(from, emptyList())
-                .stream()
-                .anyMatch(s -> s == to);
+                .contains(to);
 
         if (!ok) {
             throw new InvalidStatusChangeException("Cannot change from status " + from + " to status " + to);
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeStatusChangeValidator.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public void assertCanUpdate(Status from, Status to) {
+        boolean ok =
+            allowedTransitions
+                .getOrDefault(from, emptyList())
+                .stream()
+                .anyMatch(s -> s == to);
+
+        if (!ok) {
+            throw new InvalidStatusChangeException("Cannot change from status " + from + " to status " + to);
```
doncem: Would be nice if service (or any other component) did not care about message construct instead the exception itself requires what is needed in order to construct it. Simply said: maybe just pass `fromStatus` and `toStatus`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeStatusChangeValidator.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Checks whether it's legal to transition from given status to another.
+     */
+    public void assertCanUpdate(Status from, Status to) {
+        boolean ok =
+            allowedTransitions
+                .getOrDefault(from, emptyList())
+                .stream()
+                .anyMatch(s -> s == to);
```
doncem: We could just check whether `list.contains(element)`

rkondratowicz: good point! fixed.

PR #76, opened 2018-07-30
Fix an envelope deletion test

lgonczar:
### Change description ###

- Fix an always-passing tests that didn't wait for processing
- Set a polling interval in a test to 1 second (default is 100 ms)
- Make sure blobs are deleted after tests

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
@@ -8,12 +8,14 @@ import com.microsoft.azure.storage.blob.CloudBlobContainer;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.UUID;
@@ -36,6 +38,8 @@ public class EnvelopeDeletionTest {
 
     private transient CloudBlobContainer testContainer;
 
+    private List<String> filesToDeleteAfterTest = new ArrayList<>();
+
     @Before
     public void setUp() throws Exception {
         Config conf = ConfigFactory.load();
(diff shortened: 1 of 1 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
```diff
@@ -36,6 +38,8 @@
 
     private transient CloudBlobContainer testContainer;
 
+    private List<String> zipFilesToDelete = new ArrayList<>();
```
rkondratowicz: a comment (or different name) would be good, to clarify that this is for the cleanup after tests only.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
```diff
@@ -74,25 +86,24 @@ public void should_keep_zip_file_after_failed_processing() throws Exception {
         uploadZipFile(srcZipFilename, destZipFilename); // invalid due to missing json file
 
         // ensure that processing has happened
-        await()
-            .atMost(scanDelay + 15_000, TimeUnit.MILLISECONDS)
-            .pollDelay(scanDelay * 2, TimeUnit.MILLISECONDS)
```
doncem: `pollDelay` was the reason to wait for processing to happen. If you intend to sleep instead for X amount, why not sticking X amount to `pollDelay` function instead? Won't need a `TODO` either 🙃

lgonczar: What we're effectively doing here is we wait for a specific amount of time and then make the assertion. This isn't really waiting until the condition is met, as you'd think by reading this chain. `Thread.sleep` makes it much more obvious. The TODO is about waiting for the processing of the file to happen (by polling some endpoint) and then making an assertion about the existence of the file. That would save us a lot of time on average.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
lgonczar: Will do

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
doncem: Ah, _endpoint polling-based solution_ 😉 . Probably would be enough. Still recommend using library instead of `Thread.sleep`

lgonczar: Can be an endpoint or any other way of checking if the file has been processed.

The statement reads "await at most ... until", when in fact it's not like this at all. We're just waiting for a specific amount of time and then make an assertion. There's also a conflict between the values in `atMost` and `pollDelay`. Depending on the configuration, pollDelay may be greater than atMost, which leads to an error.

PR #77, opened 2018-07-30
Add endpoint for making updates to envelope status

rkondratowicz:
plus unit tests.

integration and functional tests to come in next PR(s).

using `envelopes/{id}/status` route as suggested by Neil during planning 
(just `envelopes/{id}` would imply that this endpoint expects an entire envelope payload)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
@@ -65,10 +65,13 @@ public class EnvelopeController {
     ) {
         String serviceName = authService.authenticate(serviceAuthHeader);
 
+        // TODO: move to service
         switch (statusUpdate.status) {
             case CONSUMED:
                 envelopeUpdateService.markAsConsumed(id, serviceName);
                 break;
+            default:
+                throw new IllegalStateException("No code for handling status: " + statusUpdate.status);
         }
 
         return noContent().build();
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
@@ -3,26 +3,39 @@ package uk.gov.hmcts.reform.bulkscanprocessor.controllers;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiResponse;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.in.StatusUpdate;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.EnvelopeMetadataResponse;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.AuthService;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.EnvelopeRetrieverService;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.EnvelopeUpdateService;
+
+import java.util.UUID;
+
+import static org.springframework.http.ResponseEntity.noContent;
 
 @RestController
 @RequestMapping(path = "envelopes", produces = MediaType.APPLICATION_JSON_VALUE)
 public class EnvelopeController {
 
     private final EnvelopeRetrieverService envelopeRetrieverService;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/NewStatus.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.model.in;
+
+public enum NewStatus {
+    CONSUMED,
+}
```
rkondratowicz: Separate enums we only allow changing to some (well, one particular) statuses.
This will result in rejecting any requests with status that is not defined here.
Plus the swagger schema will be nicer (will only list valid statuses).

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
```diff
@@ -41,4 +54,26 @@ public EnvelopeMetadataResponse getProcessedEnvelopesByJurisdiction(
...
+    ) {
+        String serviceName = authService.authenticate(serviceAuthHeader);
+
+        // TODO: move to service
+        switch (statusUpdate.status) {
+            case CONSUMED:
+                envelopeUpdateService.markAsConsumed(id, serviceName);
+                break;
+            default:
+                throw new IllegalStateException("No code for handling status: " + statusUpdate.status);
```
doncem: Due to _new_ enum this branch is now not covered

rkondratowicz: checkstyle complain if `default` is missing

nitinprabhu: Is IllegalStateException the right kind of exception to be used here ? Could we use InvalidStatusException or something similar ?

doncem: `switch` statement has to have `default` - no checkstyle report needed 😝 . for simple usage as it is here, block can be replaced with `if` statement.
since there is only one new state present, there can be zero branches in general and just `markAsConsumed` instantly. but that will introduce risk for future.

so for simple condition test and full branch coverage - just use `if` 👍 . should be all good

doncem: this branch is never reached due to single enum being present in enum class

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/UpdateStatusEnvelopesControllerTest.java
```diff
@@ -0,0 +1,121 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
+
+
+@RunWith(SpringRunner.class)
+@WebMvcTest(EnvelopeController.class)
+public class UpdateStatusEnvelopesControllerTest {
```
nitinprabhu: Lots of tests nice one 👏

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/NewStatus.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.model.in;
+
+public enum NewStatus {
+    CONSUMED,
```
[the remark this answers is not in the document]
rkondratowicz: It's there so that when you add a new one you don't have to modify this line as well.
Checkstyle is ok with it btw ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/NewStatus.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.model.in;
+
+public enum NewStatus {
```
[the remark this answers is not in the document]
rkondratowicz: `Status` already exists

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/NewStatus.java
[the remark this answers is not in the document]
doncem: For enums Java accepts comma suffixed values as long as there is no extra code blocks present

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeController.java
rkondratowicz: with `if-elseif-...-else` the question still remains what should go in the last `else`

doncem: There is no else atm.

In case you want this pattern - re-use existing status and cover this extra branch with unit test.

rkondratowicz: there is no `if` atm. let's talk.

PR #78, opened 2018-07-31
Move all status update logic from controller to service

rkondratowicz:
Also allow all status in the controller


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeUpdateService.java
```diff
@@ -31,24 +32,25 @@ public EnvelopeUpdateService(
...
             envelopeRepo
                 .findById(envelopeId)
                 .orElseThrow(() -> new EnvelopeNotFoundException());
 
         accessService.assertCanUpdate(envelope.getJurisdiction(), serviceName);
-        statusChangeValidator.assertCanUpdate(envelope.getStatus(), Status.CONSUMED);
 
-        envelope.setStatus(Status.CONSUMED);
-        envelopeRepo.save(envelope);
+        if (newStatus == NewStatus.CONSUMED) {
```
nitinprabhu: We could avoid if else ?

```

if (newStatus != NewStatus.CONSUMED) {
 throw new IllegalStateException("No code for handling status: " + newStatus);
}

statusChangeValidator.assertCanUpdate(envelope.getStatus(), Status.CONSUMED);

envelope.setStatus(Status.CONSUMED);
envelopeRepo.save(envelope);

eventRepo.save(new ProcessEvent(envelope.getContainer(), envelope.getZipFileName(), Event.DOC_CONSUMED));

```

nitinprabhu: Not sure if throwing is IllegalStateException  appropriate here
> Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

rkondratowicz: I consider `if-else` better than `if` in general.
It's more explicit.

PR #83, opened 2018-08-01
[PREVIEW]RPE-628 Fixed Envelope field definition

luigibk:
https://tools.hmcts.net/jira/browse/RPE-628

### Change description ###
Updated envelope field definition to fix issue with serialization/deserialization

```
[ ] Yes
[ x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeTest.java
```diff
@@ -57,13 +57,12 @@ public void should_insert_into_db_and_validate_data_correctness_when_retrieved()
...
             Properties to ignore because:
             - id - autogenerated by us
-            - zip_file_created_date - because we fix wording incorrectly provided in metafile
             - amount / amount_in_pence - converting to integers/doubles and tested in unit case already
             - configuration/json - auto-included fields by JsonPath/assertj when merging inner objects
              */
             assertThat(JsonPath.parse(actualEnvelope)).isEqualToIgnoringGivenFields(
                 JsonPath.parse(originalMetaFile),
-                "id", "zip_file_created_date", "amount", "amount_in_pence", "configuration", "json"
+                "id", "amount", "amount_in_pence", "configuration", "json"
```
nitinprabhu: Any reason we need to remove the created date field ?

doncem: removing from ignore list as field name now matches as per spec

luigibk: They were being ignored (because of the field name mismatch I think) but they don't need to be anymore.

PR #58, opened 2018-07-25
Feature: Custom task scheduler

doncem:
### JIRA link (if applicable) ###

[Automated exception handling for failed documents](https://tools.hmcts.net/jira/browse/RPE-562)

### Change description ###

Effect-less implementation of automated exception handling since each exception in bulk scan processor task is caught and logged only.

[Proposal commit](https://github.com/hmcts/bulk-scan-processor/commit/8e99b24c99be007820b450dca39ea73193f44c53) is the heaviest part of this PR and not related to core feature implementation. Just bringing up the coverage diff, test the component at it's core

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #84, opened 2018-08-01
[PREVIEW] Refactor: processor task

doncem:

### Change description ###

Pending on #58 

Not breaking the current blob processor work flow in case of an error after the knowledge of a zip file. Breaking the task in case access to blob storage itself is faulty. In that case custom task scheduler logs failure using defined `ErrorHandler`.

```
[x] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.Resources;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.boot.test.context.SpringBootTest;
@@ -10,7 +11,10 @@ import org.springframework.test.context.junit4.SpringRunner;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Event;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.ProcessEvent;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocFailureGenericException;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocUploadFailureGenericException;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocumentNotFoundException;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.MetadataNotFoundException;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.document.output.Pdf;
 
 import java.nio.charset.Charset;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/DocFailureGenericException.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Event;
+
+public class DocFailureGenericException extends RuntimeException implements EventRelatedThrowable {
+
+    private static final Event EVENT = Event.DOC_FAILURE;
```
luigibk: I'd give it a more descriptive name than simply "EVENT"

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -150,40 +132,6 @@ public void should_process_other_zip_files_if_previous_zip_fails_to_process() th
             .uploadDocuments(ImmutableList.of(new Pdf("1111001.pdf", test1PdfBytes)));
     }
 
-    @Test
```
luigibk: Are these removed integration tests going to be replaced by other tests?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipEntryProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import static com.google.common.io.ByteStreams.toByteArray;
+
+public class ZipEntryProcessor {
```
luigibk: I'd say ZipEntriesProcessor or ZipFileProcessor. This does not process a single entry as far as I can see.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
doncem: Those where moved to `BlobProcessorTaskTestForFailedStatus` suite. Forgot to remove in the first place

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/DocFailureGenericException.java
doncem: For public access makes sense. But I'll rename

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/NoPdfFileFoundException.java
```diff
@@ -1,10 +1,36 @@
...
-public class NoPdfFileFoundException extends RuntimeException {
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Event;
+
+public class NoPdfFileFoundException extends RuntimeException implements EventRelatedThrowable {
 
     private static final long serialVersionUID = 9143161748679833084L;
 
-    public NoPdfFileFoundException(String message) {
-        super(message);
+    private static final Event EVENT = Event.DOC_FAILURE;
```
luigibk: I'd also rename this to DOC_FAILURE_EVENT. It really reads better without any need to navigate to the definition

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -46,97 +45,84 @@
...
-                }
-            }
+            Map<Envelope, List<Pdf>> envelopeMap = processZipFileEntry(zis, zipFilename, container.getName());
 
-            envelope = envelopeProcessor.processEnvelope(metadataStream, container.getName());
-            isUploadFailure = true;
-
-            documentProcessor.processPdfFiles(pdfFiles, envelope.getScannableItems());
-            envelopeProcessor.markAsUploaded(envelope, container.getName(), zipFilename);
+            for (Map.Entry<Envelope, List<Pdf>> entry : envelopeMap.entrySet()) {
```
nitinprabhu: There will be always one envelope and collection of pdf's. Any reason why we are using map here and then iterating through entry set ?

nitinprabhu: Also you have  a new class now processing of zip files can it be moved to that class ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/DocFailureGenericException.java
```diff
@@ -0,0 +1,34 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Event;
+
+public class DocFailureGenericException extends RuntimeException implements EventRelatedThrowable {
```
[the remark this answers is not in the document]
doncem: Kind of suggests you have envelope in there. And it's the opposite

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: Good point. Will simplify error wrapper

PR #89, opened 2018-08-02
[PREVIEW]RPE-564 Smoke test for Bulk scan component

luigibk:
https://tools.hmcts.net/jira/browse/RPE-564

### Change description ###
Smoke test for Bulk scan component

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/smokeTest/java/uk/gov/hmcts/reform/bulkscanprocessor/WhenApplicationDeployedTest.java
+++ b/src/smokeTest/java/uk/gov/hmcts/reform/bulkscanprocessor/WhenApplicationDeployedTest.java
@@ -1,13 +1,16 @@
 package uk.gov.hmcts.reform.bulkscanprocessor;
 
 import io.restassured.RestAssured;
+import io.restassured.response.Response;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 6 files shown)


src/smokeTest/java/uk/gov/hmcts/reform/bulkscanprocessor/WhenApplicationDeployedTest.java
```diff
@@ -29,4 +41,14 @@ public void storage_container_exists() {
             .then()
             .body("containers", not(emptyArray()));
     }
+
+    @Test
+    public void can_retrieve_envelopes() {
```
nitinprabhu: I would suggest instead of doing we have a method in Envelope Service like below

```
 public List<Envelope> getPageableEnvelopes() {
        Pageable pageable = PageRequest.of(0, 1);
        Page<Envelope> envelopes = envelopeRepository.findAll(pageable);
        return envelopes.getContent();
    }
```
And also have a new endpoint in controller which uses this method and finally for smoke we call this method.

This way we don't retrieve everything and also validate system is working fine :)

luigibk: I agree with the pagination, but I'd leave that for a different ticket.

PR #90, opened 2018-08-02
Handle InvalidTokenException

rkondratowicz:
So that we get 401 instead of 500 on both reads and updates


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ResponseExceptionHandler.java
```diff
@@ -59,6 +60,12 @@
         return status(HttpStatus.FORBIDDEN).body(exc.getMessage());
     }
 
+    @ExceptionHandler(InvalidTokenException.class)
+    protected ResponseEntity<String> handleInvalidToken(InvalidTokenException exc) {
```
nitinprabhu: Could you please add a test case ?

rkondratowicz: done

PR #93, opened 2018-08-03
Skip processing for failed uploads

doncem:

### Change description ###

In case latest envelope is marked as `UPLOAD_FAILURE` - skip the process by raising exception which will be logged appropriately by the error handler inside the wrapper and proceed with next blob inline.

As part of this PR envelopes needed to have `created_at` column to distinguish sequential order of when they were created. It is needed for business too so win win

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -35,15 +38,25 @@ public EnvelopeProcessor(
...
         }
         //TODO Perform json schema validation for the metadata file
         InputStream inputStream = new ByteArrayInputStream(metadataStream);
-        Envelope envelope = EntityParser.parseEnvelopeMetadata(inputStream);
-        envelope.setContainer(containerName);
 
+        return EntityParser.parseEnvelopeMetadata(inputStream);
+    }
+
+    public Optional<Envelope> hasEnvelopeFailedToUploadBefore(Envelope envelope) {
```
luigibk: Can't all this be done in sql/jpaql? That could be (much) more efficient depending on the dataset

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -107,10 +109,25 @@ private ZipFileProcessor processZipFileEntry(
...
-            Envelope envelope = envelopeProcessor.processEnvelope(zipFileProcessor.getMetadata(), containerName);
-            zipFileProcessor.setEnvelope(envelope);
+            Envelope envelope = envelopeProcessor.parseEnvelope(zipFileProcessor.getMetadata());
+            envelope.setContainer(containerName);
 
-            return zipFileProcessor;
+            Optional<Envelope> optionalEnvelope = envelopeProcessor.hasEnvelopeFailedToUploadBefore(envelope);
+
+            if (optionalEnvelope.isPresent()) {
+                throw new PreviouslyFailedToUploadException(
```
nitinprabhu: Do we need to throw an exception ? we can just log and skip the record

src/main/resources/db/migration/V008__Envelope_upgrade.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE envelopes
```
luigibk: Shouldn't you also set a timestamp for all the records already in the db?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: Error wrapper does that so you don't worry and continue the work

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -14,4 +14,13 @@
...
     List<Envelope> findByJurisdictionAndStatus(String jurisdiction, Status status);
+
+    /**
+     * Finds envelopes for a given container and zip file name.
+     *
+     * @param container from where container originated
+     * @param zipFileName of envelope
+     * @return A list of envelopes
+     */
+    List<Envelope> findByContainerAndZipFileName(String container, String zipFileName);
```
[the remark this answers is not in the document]
doncem: I'll move to jql w/e dsl as @luigibk suggested. Didn't want to go about on first go but better make it efficient and readable from the start

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -14,4 +17,24 @@
...
+
+    /**
+     * Finds envelope for a given container, zip file name and status.
+     *
+     * @param container from where container originated
+     * @param zipFileName of envelope
+     * @param status of envelope
+     * @return Optional envelope
+     */
+    @Query("select e from Envelope e"
```
nitinprabhu: You can do exists as you are not doing anything with the fetched envelope

nitinprabhu: something like below

```

 @Query("select CASE WHEN COUNT(e) > 0 THEN true ELSE false END from Envelope e"
         + "   and e.status = :status"
     )
     Optional<Envelope> existsByStatus(
         @Param("container") String container,
         @Param("zip") String zipFileName,
         @Param("status") Status status
     );

```

doncem: I'm using envelope to log ID and created at. Can skip the log to use exist query or leave as it is now

nitinprabhu: ohh I see `createdAt` might be useful while debugging so should be fine 👍

PR #96, opened 2018-08-06
Apply limit to the query

doncem:
### Change description ###

Final improvement in query - applying limit. Using `LIMIT` inside HQL is not supported. Instead need to provide `Pageable`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
@@ -20,8 +20,6 @@ public interface EnvelopeRepository extends JpaRepository<Envelope, UUID> {
 
     /**
      * Finds envelope for a given container, zip file name and status.
-     * Providing {@code PageRequest.of(0, 1)} should allow to return {@code Optional<Envelope>}
-     * but that is not the case even after following this notion in documentation.
      *
      * @param container from where container originated.
      * @param zipFileName of envelope.
@@ -34,7 +32,7 @@ public interface EnvelopeRepository extends JpaRepository<Envelope, UUID> {
         + "   and e.status = :status"
         + " order by e.createdAt desc"
     )
-    List<Envelope> checkLastEnvelopeByStatus(
+    List<Envelope> findRecentEnvelopes(
         @Param("container") String container,
         @Param("zip") String zipFileName,
         @Param("status") Status status,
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -47,12 +49,33 @@ public Envelope parseEnvelope(byte[] metadataStream) throws IOException {
...
+        List<Envelope> envelopes = envelopeRepository.checkLastEnvelopeByStatus(
             envelope.getContainer(),
             envelope.getZipFileName(),
-            UPLOAD_FAILURE
+            UPLOAD_FAILURE,
+            PageRequest.of(0, 1)
         );
+
+        if (envelopes.size() == 1) {
+            Envelope failedEnvelope = envelopes.get(0);
```
luigibk: Why not check if size() >= 1? I know it should not happen, but better safe than sorry...

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -21,20 +21,21 @@
...
+     * @return A list of envelopes.
      */
     @Query("select e from Envelope e"
         + " where e.container = :container"
         + "   and e.zipFileName = :zip"
         + "   and e.status = :status"
         + " order by e.createdAt desc"
     )
-    Optional<Envelope> checkLastEnvelopeStatus(
+    List<Envelope> findRecentEnvelopes(
```
luigibk: findRecentEnvelopesByStatus() or am I being pedantic again?

rkondratowicz: @luigibk by container, zipfilename AND status if anything ;)

doncem: Apparently having any kind of `By` message in there would confuse a lot since query have not only `status` to filter against. It was easier and still understandable to make just `recentEnvelopes` and you will have type-hint explicitly saying what you need: `container`, `zip`, `status`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
rkondratowicz: I'd personally go with `> 0` and leave logging as is.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
rkondratowicz: as longer as there's only single `find` were's good imho 👍

PR #98, opened 2018-08-07
Feature: failed upload task

doncem:

### Change description ###

Core set up of new scheduler task handling failed documents in an efficient way:

- Designate available thread pool size by choosing ~between maximum available from task scheduler and actual~ configured jurisdictions
- Instantiate processor prototype separately per jurisdiction so as per cloud blob storage.

Further implementation is coming later leaving this PR as an entry point.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTaskTest.java
@@ -2,7 +2,9 @@ package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 
 import com.microsoft.azure.storage.CloudStorageAccount;
 import com.microsoft.azure.storage.blob.CloudBlobClient;
+import com.palantir.docker.compose.DockerComposeRule;
 import org.junit.After;
+import org.junit.ClassRule;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -28,6 +30,11 @@ import static org.mockito.BDDMockito.willThrow;
 })
 @RunWith(SpringRunner.class)
 public class ReuploadFailedEnvelopeTaskTest {
+
+    @ClassRule
+    public static DockerComposeRule docker = DockerComposeRule.builder()
+        .file("src/integrationTest/resources/docker-compose.yml")
+        .build();
 
     @Rule
     public OutputCapture outputCapture = new OutputCapture();
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTaskTest.java
@@ -1,14 +1,27 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 
+import com.microsoft.azure.storage.CloudStorageAccount;
+import com.microsoft.azure.storage.blob.CloudBlobClient;
+import org.junit.After;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.context.TestConfiguration;
 import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.boot.test.rule.OutputCapture;
+import org.springframework.context.annotation.Bean;
+import org.springframework.orm.jpa.JpaObjectRetrievalFailureException;
 import org.springframework.test.context.junit4.SpringRunner;
-import uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor.FailedDocUploadProcessor;
+import uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor.EnvelopeProcessor;
+
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.BDDMockito.willThrow;
 
 @SpringBootTest(properties = {
     "scheduling.task.reupload.enabled=true"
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -38,4 +38,12 @@
...
         Pageable pageable
     );
+
+    /**
+     * Finds first 20 envelopes for a given status.
+     *
+     * @param status to filter upon
+     * @return A list of envelopes
+     */
+    List<Envelope> findFirst20ByStatus(Status status);
```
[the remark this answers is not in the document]
nitinprabhu: Not sure where this 20 came from ? why we are only fetching 20 envelopes at a time ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTask.java
```diff
@@ -0,0 +1,51 @@
...
+ * It will retrieve failed to upload envelopes from DB and try to re-upload.
+ * <ol>
+ *     <li>Get batch of failed to upload envelopes</li>
+ *     <li>Retrieve relevant zip file from Blob Storage</li>
+ *     <li>Try to upload pdfs</li>
+ * </ol>
+ */
+@Component
+@ConditionalOnProperty(value = "scheduling.enabled", matchIfMissing = true)
+public class ReuploadFailedEnvelopeTask {
```
nitinprabhu: Currently this ReuploadFailedEnvelopeTask fetches first 20 records at a time. It should fetch all configured jurisdiction and then process failed records for each jurisdiction in parallel for e.g we should process sscs and divorce failed records separately and parallely. Within each jurisdiction you can batch them i.e process few records at a time but that should be configurable and not hardcoded.

```
ReuploadFailedEnvelopeTask.java

private EnvelopeAccessProperties accessProps;

    ExecutorService executor = Executors.newFixedThreadPool(
            Runtime.getRuntime().availableProcessors()+1);
    
accessProps.getMappings().forEach(
    mapping -> {
 			executor.execute(
              new FailedDocUploadProcessor().process(mapping.getJurisdiction())
             );
    })

```

```
FailedDocUploadProcessor.java 
    1. Fetch envelopes for the given jurisdiction with status upload failure.You can fetch few records at a time or fetch all(we don't expect many failures unless doc upload service is down)
    2. Try uploading the pdf's
    3. Update scannable items 
```

luigibk: or you could also do this I believe

`accessProps.getMappings()
   .parallelStream()
   .forEach(m -> new FailedDocUploadProcessor().process(m.getJurisdiction()))`

With the caveat that it uses the fork/join thread pool which is shared. On the other hand I believe, the explicitly created thread pool above needs to be also explicitly shutdown.

nitinprabhu: You have to be very careful before using parallel stream. The problem is that all parallel streams use common fork-join thread pool and if you submit a long-running task, you effectively block all threads in the pool.

Yes executor needs to be shutdown explicitly.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTask.java
```diff
@@ -0,0 +1,102 @@
...
+    private final List<Mapping> accessMapping;
+
+    private final int poolSize;
+
+    public ReuploadFailedEnvelopeTask(
+        EnvelopeAccessProperties accessProperties,
+        TaskScheduler taskScheduler
+    ) {
+        this.accessMapping = accessProperties.getMappings();
+        poolSize = Math.max(((ThreadPoolTaskScheduler) taskScheduler).getPoolSize(), accessMapping.size());
```
nitinprabhu: We can just configure the thread pool size based on access mappings

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTask.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Lookup
```
nitinprabhu: Worth adding a comment here why is this configuration required.

src/main/resources/application.yaml
```diff
@@ -65,9 +65,12 @@ scheduling:
   enabled: ${SCHEDULING_ENABLED:false}
   pool: ${SCHEDULING_POOL:10}
   lock_at_most_for: ${SCHEDULING_LOCK_AT_MOST_FOR:PT10M} # 10 minutes in ISO-8601
-
-scan:
-  delay: ${SCAN_DELAY:30000} # In milliseconds
+  task:
+    scan:
+      delay: ${SCAN_DELAY:30000} # In milliseconds
```
nitinprabhu: Please do not forget to add those in app setting in tf config in next PR

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/FailedDocUploadProcessorTest.java
```diff
@@ -0,0 +1,69 @@
...
+import uk.gov.hmcts.reform.bulkscanprocessor.services.wrapper.ErrorHandlingWrapper;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+
+@RunWith(MockitoJUnitRunner.class)
+public class FailedDocUploadProcessorTest {
```
nitinprabhu: Please add failure test cases

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTask.java
```diff
@@ -0,0 +1,102 @@
...
+ * <ol>
+ *     <li>Get batch of failed to upload envelopes</li>
+ *     <li>Retrieve relevant zip file from Blob Storage</li>
+ *     <li>Try to upload pdfs</li>
+ * </ol>
+ */
+@Component
+@ConditionalOnProperty(value = "scheduling.enabled", matchIfMissing = true)
+@EnableConfigurationProperties(EnvelopeAccessProperties.class)
+public class ReuploadFailedEnvelopeTask {
```
nitinprabhu: Are we missing unit tests for this class ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/FailedDocUploadProcessorTest.java
doncem: The task itself is empty - there is nothing to fail against. Next batch of code has almost exact coverage as `BlobProcessorTask`. Can double check then

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTask.java
doncem: I'll think of something here. Since actual task deals with concurrency and processor - envelopes. Latter is/will be extensively tested in next PR

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ReuploadFailedEnvelopeTaskTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Test
+    public void should_return_new_instance_of_processor_when_rerequesting_via_lookup_method() {
+        // comparing instance hashes. .hashCode() just returns same one
+        assertThat(task.getProcessor().toString()).isNotEqualTo(task.getProcessor().toString());
+    }
+
+    @Test
+    public void should_await_for_all_futures_even_if_their_code_throw_exception() throws Exception {
```
nitinprabhu: throw -> throws ?

PR #100, opened 2018-08-08
[PREVIEW]External model for API

luigibk:
https://tools.hmcts.net/jira/browse/RPE-630

### Change description ###
Create external model for API

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
@@ -72,6 +72,8 @@ public class EnvelopeResponseMapper {
             .collect(Collectors.toList());
     }
 
+    
+
     private ScannableItemResponse toScannableItemResponse(ScannableItem scannableItem) {
         if (scannableItem == null) {
             return null;
@@ -128,6 +130,103 @@ public class EnvelopeResponseMapper {
             payment.getMethod(),
             Double.toString(payment.getAmount()),
             payment.getCurrency()
+        );
+    }
+
+
+
+    public List<Envelope> toEnvelopes(List<EnvelopeResponse> envelopesResponse) {
+        if (envelopesResponse == null) {
+            return null;
+        }
+        return envelopesResponse.stream()
+            .map(this::toEnvelope)
+            .collect(Collectors.toList());
+    }
+
+    public Envelope toEnvelope(EnvelopeResponse envelopeResponse) {
+        if (envelopeResponse == null) {
+            return null;
+        }
+        Envelope envelope = new Envelope(
+            envelopeResponse.getPoBox(),
+            envelopeResponse.getJurisdiction(),
+            envelopeResponse.getDeliveryDate(),
+            envelopeResponse.getOpeningDate(),
+            envelopeResponse.getZipFileCreateddate(),
+            envelopeResponse.getZipFileName(),
+            toScannableItems(envelopeResponse.getScannableItems()),
+            toPayments(envelopeResponse.getPayments()),
+            toNonScannableItems(envelopeResponse.getNonScannableItems())
+        );
+        return envelope;
+    }
+
+    private List<ScannableItem> toScannableItems(List<ScannableItemResponse> scannableItemsResponse) {
+        if (scannableItemsResponse == null) {
+            return null;
+        }
+        return scannableItemsResponse.stream()
+            .map(this::toScannableItem)
+            .collect(Collectors.toList());
+    }
+
+    private ScannableItem toScannableItem(ScannableItemResponse scannableItemResponse) {
+        if (scannableItemResponse == null) {
+            return null;
+        }
+        ScannableItem scannableItem = new ScannableItem(
+            scannableItemResponse.getDocumentControlNumber(),
+            scannableItemResponse.getScanningDate(),
+            scannableItemResponse.getOcrAccuracy(),
+            scannableItemResponse.getManualIntervention(),
+            scannableItemResponse.getNextAction(),
+            scannableItemResponse.getNextActionDate(),
+            scannableItemResponse.getOcrData(),
+            scannableItemResponse.getFileName(),
+            scannableItemResponse.getNotes()
+        );
+        scannableItem.setDocumentUrl(scannableItemResponse.getDocumentUrl());
+        return scannableItem;
+    }
+
+    private List<NonScannableItem> toNonScannableItems(List<NonScannableItemResponse> nonScannableItemsResponse) {
+        if (nonScannableItemsResponse == null) {
+            return null;
+        }
+        return nonScannableItemsResponse.stream()
+            .map(this::toNonScannableItem)
+            .collect(Collectors.toList());
+    }
+
+    private NonScannableItem toNonScannableItem(NonScannableItemResponse nonScannableItemResponse) {
+        if (nonScannableItemResponse == null) {
+            return null;
+        }
+        return new NonScannableItem(
+            nonScannableItemResponse.getItemType(),
+            nonScannableItemResponse.getNotes()
+        );
+    }
+
+    private List<Payment> toPayments(List<PaymentResponse> paymentsResponse) {
+        if (paymentsResponse == null) {
+            return null;
+        }
+        return paymentsResponse.stream()
+            .map(this::toPayment)
+            .collect(Collectors.toList());
+    }
+
+    private Payment toPayment(PaymentResponse paymentResponse) {
+        if (paymentResponse == null) {
+            return null;
+        }
+        return new Payment(
+            paymentResponse.getDocumentControlNumber(),
+            paymentResponse.getMethod(),
+            Double.toString(paymentResponse.getAmount()),
+            paymentResponse.getCurrency()
         );
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
@@ -30,19 +30,19 @@ public class EnvelopeResponseMapper {
             return null;
         }
         EnvelopeResponse response = new EnvelopeResponse(
+            envelope.getId(),
+            envelope.getContainer(),
             envelope.getPoBox(),
             envelope.getJurisdiction(),
             envelope.getDeliveryDate(),
             envelope.getOpeningDate(),
             envelope.getZipFileCreateddate(),
             envelope.getZipFileName(),
+            envelope.getStatus(),
             toScannableItemsResponse(envelope.getScannableItems()),
             toPaymentsResponse(envelope.getPayments()),
             toNonScannableItemsResponse(envelope.getNonScannableItems())
         );
-        response.setContainer(envelope.getContainer());
-        response.setId(envelope.getId());
-        response.setStatus(envelope.getStatus());
         return response;
     }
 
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
```diff
@@ -0,0 +1,233 @@
...
+            envelope.getOpeningDate(),
+            envelope.getZipFileCreateddate(),
+            envelope.getZipFileName(),
+            toScannableItemsResponse(envelope.getScannableItems()),
+            toPaymentsResponse(envelope.getPayments()),
+            toNonScannableItemsResponse(envelope.getNonScannableItems())
+        );
+        response.setContainer(envelope.getContainer());
+        response.setId(envelope.getId());
+        response.setStatus(envelope.getStatus());
```
rkondratowicz: can we move all fields to the constructor?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/NonScannableItemResponse.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.model.out;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class NonScannableItemResponse {
+
+    @JsonProperty("item_type")
+    private String itemType;
+    @JsonProperty("notes")
```
rkondratowicz: new line would be nice

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/EnvelopeResponse.java
```diff
@@ -0,0 +1,163 @@
...
+import java.sql.Timestamp;
+import java.util.List;
+import java.util.UUID;
+
+import static java.util.Collections.emptyList;
+
+public class EnvelopeResponse {
+
+    @JsonIgnore
+    private UUID id;
```
rkondratowicz: pretty sure id should be returned

rkondratowicz:
Please add the '[PREVIEW]' prefix to run functional tests. I think at least one may be broken (relying on ID being returned in the response)

doncem:
I think you need to close and at least re-open (if not create new one) the PR so that jenkins can register it as a `PREVIEW`. Renaming doesn't do the trick

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
```diff
@@ -22,23 +23,27 @@
...
     private final EnvelopeAccessProperties envelopeAccessProperties;
+    private final EnvelopeResponseMapper envelopeResponseMapper;
 
     public EnvelopeRetrieverService(
         EnvelopeRepository envelopeRepository,
         EnvelopeAccessProperties envelopeAccessProperties
     ) {
         this.envelopeRepository = envelopeRepository;
         this.envelopeAccessProperties = envelopeAccessProperties;
+        this.envelopeResponseMapper = new EnvelopeResponseMapper();
```
rkondratowicz: We may as well make all the methods in `EnvelopeResponseMapper`.
What do others think?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
```diff
@@ -0,0 +1,213 @@
...
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ScannableItem;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.EnvelopeResponse;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.NonScannableItemResponse;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.PaymentResponse;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.ScannableItemResponse;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class EnvelopeResponseMapper {
```
nitinprabhu: Could we use some standard model mapping library instead of this this class ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/ScannableItemResponse.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public String toString() {
+        return "ScannableItemResponse{"
+            + "documentControlNumber='" + documentControlNumber + '\''
+            + ", scanningDate=" + scanningDate
+            + ", ocrAccuracy='" + ocrAccuracy + '\''
+            + ", manualIntervention='" + manualIntervention + '\''
+            + ", nextAction='" + nextAction + '\''
+            + ", nextActionDate=" + nextActionDate
+            + ", ocrData='" + ocrData + '\''
```
nitinprabhu: ocrdata should not be outputted

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
```diff
@@ -40,15 +41,18 @@ public void setUp() {
...
 
         when(envelopeRepository.findByJurisdictionAndStatus("testJurisdiction", PROCESSED))
             .thenReturn(envelopes);
 
-        assertThat(envelopeRetrieverService.getProcessedEnvelopesByJurisdiction("testService"))
-            .containsOnly(envelopes.get(0));
+        List<EnvelopeResponse> retrievedResponses =
+            envelopeRetrieverService.getProcessedEnvelopesByJurisdiction("testService");
+        assertThat(retrievedResponses).hasSize(1);
+        assertThat(retrievedResponses.get(0)).isEqualToComparingFieldByFieldRecursively(envelopeResponses.get(0));
```
nitinprabhu: Instead of getting first element could we not do 

```
 assertThat(retrievedResponses).containsOnly(EnvelopeCreator.expectedRespons());

```

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
```diff
@@ -40,15 +41,18 @@ public void setUp() {
     @Test
     public void should_return_all_envelopes_successfully_for_a_given_jurisdiction() throws Exception {
         List<Envelope> envelopes = EnvelopeCreator.envelopes();
+        List<EnvelopeResponse> envelopeResponses = EnvelopeCreator.envelopeResponses();
```
nitinprabhu: Why this `EnvelopeCreator.envelopeResponses()` need to return list rather than a single response object ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
luigibk: I am not sure this simple task is worth a new library with the added complexity. I have thought about it and decided against. If the model becomes bigger it's probably worth reconsidering.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/ScannableItemResponse.java
luigibk: in toString()? why? is it potentially too big?

nitinprabhu: it's user data so definitely we should not log it :)

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
luigibk: No, because we do not implement equals() and hashcode() and the 2 objects need to be compared field by field

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
luigibk: Just for consistency with the envelopes() method as I like them to be interchangeable moving from the internal model to the external one.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
nitinprabhu: so you can do something similar to because get(0) is never good. What if you get mutliple objects ?

```
assertThat(retrievedResponses)
            .extracting("fileld1", "fileld2", "field3")
            .containsOnly(
                tuple("value1", "value2,value3)
             
     );
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
luigibk: I am not a big fan of static methods in object oriented languages even if in some cases they can be really useful. In this case it seems to me it's debatable which way is better.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
nitinprabhu: You don't need to explicitly import any library for this. There is spring bean utils for this.

``` BeanUtils.copyProperties(dot, entity);```

Comment (luigibk) [reply src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java]: But that copies everything without giving me the option to transform properties if I need to. I am generally against that. I prefer explicit copies as they are more flexible. Unless there is really much time to save, which is not the case here as the model is small

Comment (rkondratowicz) [reply src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java]: the less 'magic' the better imho.
It's easier to see what is exactly happening / spot reasons for errors if it's all explicit.

File: src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java (hunk 4 above)
Comment (rkondratowicz) [reply src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java]: if you prefer non static methods, perhaps we could at least inject an object of this instance instead of creating it inside?

File: src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java (hunk 7 above)
Comment (rkondratowicz) [reply src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java]: @luigibk I think you can also do this:
```
assertThat(retrievedResponses)
  .usingFieldByFieldElementComparator()  
  .containsOnly(EnvelopeCreator.expectedRespons())
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
luigibk: Yes, it could be injected but it's a plain old (boring) java object that holds no state. The reason why I did not make it a "component" is because I can avoid using the SpringRunner for my unit tests, but again that is really a matter of taste.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverServiceTest.java
luigibk: @rkondratowicz that looks good but (from the docs): 
"Note that the comparison is not recursive, if one of the fields/properties is an Object, it will be compared to the other field/property using its equals method. "

luigibk: @rkondratowicz  Actually this should work:
`assertThat(retrievedResponses)
.usingRecursiveFieldByFieldElementComparator()
.containsOnly(element);`

PR #108, opened 2018-08-09
[PREVIEW] RPE-610 model spec update

luigibk:
https://tools.hmcts.net/jira/browse/RPE-610

### Change description ###
Update envelope and scannable items to include new fields as per technical specifications.

See: https://tools.hmcts.net/confluence/display/RBS/Technical+Specification+-+Exela#TechnicalSpecification-Exela-1.Exelaapplication

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
```diff
@@ -38,6 +38,8 @@ public EnvelopeResponse toEnvelopeResponse(Envelope envelope) {
             envelope.getZipFileCreateddate(),
             envelope.getZipFileName(),
             envelope.getStatus(),
+            (envelope.getClassification() == null ? null : envelope.getClassification().toString()),
```
rkondratowicz: I wish Java had `?.` operator like Kotlin ;)

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
```diff
@@ -50,6 +51,35 @@ public static Envelope envelope(String jurisdiction, Status status) throws Excep
...
+        );
+
+        envelope.setStatus(status);
+        envelope.setContainer("SSCS");
+
+        return envelope;
+    }
+
+    public static Envelope envelopeUrgent() throws Exception {
+        return envelope("SSCS", Status.CREATED);
```
rkondratowicz: I think you should call `envelopeUrgent` here

luigibk: Well spotted. Thanks

src/main/resources/db/migration/V009__Model_spec_update.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE envelopes
+  ADD COLUMN classification VARCHAR(50) NOT NULL DEFAULT 'EXCEPTION',
```
doncem: Perhaps lowercased default?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/EntityParser.java
```diff
@@ -11,6 +12,7 @@
     private static final ObjectMapper mapper = new ObjectMapper();
 
     public static Envelope parseEnvelopeMetadata(InputStream metadataStream) throws IOException {
+        mapper.enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS);
```
doncem: Enabling features should be done globally. I believe with naming strategy you will solve both cases. So no need for feature.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
nitinprabhu: Envelope validation would needs to be done separately. Again when it is mandatory why we want to save null in db?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Classification.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.entity;
+
+import com.google.common.base.Strings;
+
+public enum Classification {
+
+    EXCEPTION("exception"),
+    NEW_APPLICATION("new_application"),
+    SUPPLEMENTARY_EVIDENCE("supplementary_evidence");
```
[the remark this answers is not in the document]
luigibk: It's a label. Not used just for json mapping. I prefer the transformation to be explicit as it works not just with son.

src/main/resources/db/migration/V009__Model_spec_update.sql
luigibk: No, because it's hibernate managed (and it will by default use uppercase)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/EntityParser.java
luigibk: See above, but I don't see a problem with this be enabled globally after the mapping of Status is refactored, but that is not a task related to this ticket, so it should be dealt with later

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapper.java
luigibk: You are right as we have a default value for classification now. This was from the initial implementation.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Classification.java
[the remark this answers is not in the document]
luigibk: Also the null ternary checks are meant to allow importing metadata with the old specification and assigning it the default values.

src/main/resources/db/migration/V009__Model_spec_update.sql
rkondratowicz: imho it would be best to just have `NOT NULL` here and have the logic on what value to set if it's not in the json file in application code (more explicit than java code saving null, but what ends up in db is some magic string defined in schema. but again, just my opinion).

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
```diff
@@ -74,6 +77,7 @@ public ScannableItem(
         this.ocrData = ocrData;
         this.fileName = fileName;
         this.notes = notes;
+        this.documentType = (documentType == null ? "Other" : documentType);
```
nitinprabhu: Document type should not be null and if it is null contract is broken. There will be validation in place even before it is passed to data layer

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
```diff
@@ -96,6 +103,8 @@ public Envelope(
         this.openingDate = openingDate;
         this.zipFileCreateddate = zipFileCreateddate;
         this.zipFileName = zipFileName;
+        this.classification = (classification == null ? Classification.EXCEPTION : classification);
+        this.urgent = (urgent == null ? false : urgent);
```
nitinprabhu: See below comment for classification field

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
```diff
@@ -50,6 +51,35 @@ public static Envelope envelope(String jurisdiction, Status status) throws Excep
...
+    public static Envelope envelopeUrgent(String jurisdiction, Status status) throws Exception {
+        Timestamp timestamp = getTimestamp();
+
+        Envelope envelope = new Envelope(
+            "SSCSPO",
+            jurisdiction,
+            timestamp,
+            timestamp,
+            timestamp,
+            UUID.randomUUID() + ".zip",
```
nitinprabhu: We should use proper format for filenames else looking at the test it might look like filename is a random uuid

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
luigibk: I started from the assumption that we still accept inputs using the "old" model, which does not have the new fields. This means accepting nulls and converting them to the default value. If not the case this can be changed. Please be mindful that "being permissive in what you accept and strict in what you send" is a standard principle while building an api.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
luigibk: See above comment about accepting documents using the "old" model.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
luigibk: I have not made any changes to this (ever). I suspect the semantics of the test would be changed either way, even if what you suggest would help with documenting the code.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Envelope.java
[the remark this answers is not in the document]
luigibk: SonarCloud suggestions are great if the fix does not break implementation ;-)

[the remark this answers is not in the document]
doncem: dang. these nulls

PR #88, opened 2018-08-02
Single point of usage of envelope access service

doncem:
### Change description ###

Envelope access related code was used in envelope retriever service. This piece of functionality got separate implementation since #63 

Summary:

- Reducing access config property inclusion down to single place
- Reformatting code indentation as short variable names where unnecessary put to knew line making consequent lines extra indented
- Made code by reference easier to apply/read rather than _trivial_ lambda function
- Fix unit tests using mocks of config properties

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #121, opened 2018-08-14
Move read permission checks to single service

rkondratowicz:
Extracted what seemed to be the main change in https://github.com/hmcts/bulk-scan-processor/pull/88/files which diverted quite a bit from master in the meantime.


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeAccessServiceTest.java
```diff
@@ -59,4 +62,18 @@ public void assertCanUpdate_should_not_throw_an_exception_if_service_can_update_
         assertThatCode(() -> service.assertCanUpdate("jur_A", "update_A"))
             .doesNotThrowAnyException();
     }
+
+    @Test
+    public void getReadJurisdictionForService_should_return_name_of_then_jurisdiction_from_which_service_can_read() {
```
doncem: Since when we started/decided to prefix with method name?

rkondratowicz: See other existing tests in this file.
Also good imho, as there are now 2 methods in this service, so it's clear which tests are for which method.
Alternatively I could create a separate class.

PR #125, opened 2018-08-15
Add service method to read envelope by id

rkondratowicz:
Just service part for now.
Controller usage + functional test in next PR.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
@@ -7,10 +7,14 @@ import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.EnvelopeAccessProperties;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeRepository;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Status;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ForbiddenException;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.mapper.EnvelopeResponseMapper;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.EnvelopeResponse;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
 
 @Service
 @EnableConfigurationProperties(EnvelopeAccessProperties.class)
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
```diff
@@ -7,10 +7,14 @@
...
 import uk.gov.hmcts.reform.bulkscanprocessor.model.mapper.EnvelopeResponseMapper;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.EnvelopeResponse;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
 
 @Service
 @EnableConfigurationProperties(EnvelopeAccessProperties.class)
```
doncem: just noticed that this annotation is redundant now

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeRetrieverService.java
```diff
@@ -43,4 +47,16 @@ public EnvelopeRetrieverService(
...
+            .findById(id)
+            .map(envelope -> {
+                String jurisdiction = envelopeAccessService.getReadJurisdictionForService(serviceName);
+                if (Objects.equals(envelope.getJurisdiction(), jurisdiction)) {
+                    return envelopeResponseMapper.toEnvelopeResponse(envelope);
+                } else {
+                    throw new ForbiddenException("Service " + serviceName + " cannot read envelope " + id);
+                }
+            });
+    }
```
doncem: It's not like I am trying to find a _better_ or _worse_ solution to this, but in general this looks like not using `Optionals` optimally. Plus this can and probably should return actual `EnvelopeResponse`. Perhaps better to just get rid of `Optional` wraps entirely and do smth as follows:

```java
String jurisdiction = accessService.getReadJurisdictionForService(serviceName);
Envelope dbEnvelope = repository
    .findById(id)
    .orElseThrow(EnvelopeNotFound::new);

if (envelope.getJurisdiction().isEqual(jurisdiction)) {
    return responseMapper.toResponse(envelope);
} else {
    throw new Forbidden();
}
```

rkondratowicz: I've actually originally started with throwing `EnvelopeNotFound` exception from this method, but then thought the service should just return what it found, and the controller will decide what to do with it (so the `.orElseThrow()` which results in 404, will happen in controller)

doncem: In that case it falls under strange usage of `Optional`: seeing such if/else inside `optional.map` suggests it is not used correctly.

How about

```java
String jurisdiction = accessService.getReadJurisdictionForService(serviceName);
Optional<Envelope> dbEnvelope = envelopeRepository.findById(id);

dbEnvelope
    .map(Envelope::getZipFileName)
    .filter(jurisdiction::equals)
    .orElseThrow(() -> new ForbiddenException("Service " + serviceName + " cannot read envelope " + id));

return dbEnvelope;
```

at the very least hit to access service can be extracted outside prior making a call to db

PR #129, opened 2018-08-17
[PREVIEW] Validate docs before uploading

doncem:
### JIRA ticket ###

[Validate zip file received from Exela](https://tools.hmcts.net/jira/browse/RPE-677)

### Change description ###

This PR introduces another possible FAILURE scenario when bulk scan processor is aware of envelope, but the data from zip file is not actually matching. This falls under failure before upload

```
[ ] Yes
[x] No
```


doncem:
We have few integration tests doing that for us already. But can play around with unit testing too. Especially as this component is purely unit-driven.

I may need to think about extraction in more detail

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
```diff
@@ -54,13 +60,36 @@ public void process(ZipInputStream zis) throws IOException {
...
                     // contract breakage
-                    throw new NoPdfFileFoundException(containerName, zipFileName);
+                    throw new NoPdfFileFoundException(containerName, zipFileName, zipEntry.getName());
             }
         }
 
         log.info("PDFs found in {}: {}", zipFileName, pdfs.size());
     }
 
+    public void assertEnvelopeHasPdfs() {
```
rkondratowicz: Looks like the method body has some extra logic beyond what the method suggests.
Please add javadoc.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -173,15 +162,18 @@ public void should_successfully_return_all_envelopes_with_processed_status_for_a
...
             .extracting("zipFileName", "status")
-            .containsExactlyInAnyOrder(tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
-                tuple("8_24-06-2018-00-00-00.zip", UPLOAD_FAILURE));
+            .containsExactlyInAnyOrder(
+                tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
+                tuple("8_24-06-2018-00-00-00.zip", FAILURE)
+            );
+
+        byte[] testPdfBytes1 = toByteArray(getResource("1111005.pdf"));
+        Pdf pdf1 = new Pdf("1111005.pdf", testPdfBytes1);
```
rkondratowicz: Is this the pdf file from `8_24-06-2018-00-00-00.zip` zip package?
If so, it would be good to use more meaningful variable names or add a comment.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -102,7 +103,7 @@ public void should_read_from_blob_storage_and_save_metadata_in_database_when_zip
     public void should_process_other_zip_files_if_previous_zip_fails_to_process() throws Exception {
         // given
         uploadZipToBlobStore("3_24-06-2018-00-00-00.zip"); //Zip with only pdf without metadata
-        uploadZipToBlobStore("4_24-06-2018-00-00-00.zip"); //Zip with pdf and metadata
+        uploadZipToBlobStore("7_24-06-2018-00-00-00.zip"); //Zip with pdf and metadata
```
rkondratowicz: did this have to change? ie, was `4_24-06-2018-00-00-00.zip` invalid?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
```diff
@@ -6,6 +6,7 @@
 
     CONSUMED, // client service handled the documents
     CREATED,
+    FAILURE, // when we are aware of envelope, but there are inconsistency among files and metadata info
```
rkondratowicz: We already have `UPLOAD_FAILURE`, `FAILURE` sounds very generic...
Perhaps something like 'metadata failure'???

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/NoPdfFileFoundException.java
```diff
@@ -12,8 +12,8 @@
 
     private final String zipFileName;
 
-    public NoPdfFileFoundException(String container, String zipFileName) {
-        super("Zip file '" + zipFileName + "' does not contain pdf files");
+    public NoPdfFileFoundException(String container, String zipFileName, String fileName) {
+        super("Zip '" + zipFileName + "' contains non-pdf file: " + fileName);
```
rkondratowicz: This changes the meaning of this exception completely.
Previously: there are **no** pdf files at all.
Now: there may be some pdfs but also a file that is not a pdf.

Is this indented?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
doncem: Zip file `4_...` was wrong from the very beginning. Since we did not check integrity before uploading, the error never appeared and mock always provided correct response

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -185,15 +174,18 @@ public void should_successfully_return_all_envelopes_with_processed_status_for_a
...
-
-        assertThat(envelopesFromDb)
+        assertThat(envelopeRepository.findAll())
+            .hasSize(2)
             .extracting("zipFileName", "status")
-            .containsExactlyInAnyOrder(tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
-                tuple("8_24-06-2018-00-00-00.zip", UPLOAD_FAILURE));
+            .containsExactlyInAnyOrder(
+                tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
+                tuple("8_24-06-2018-00-00-00.zip", METADATA_FAILURE)
```
rkondratowicz: Please update the comment in the setup stage of this test, at the moment it says:
```
uploadZipToBlobStore("8_24-06-2018-00-00-00.zip"); // Zip file with metadata and pdf
```
If I understand correctly, this zip file is actually invalid.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -185,15 +174,18 @@ public void should_successfully_return_all_envelopes_with_processed_status_for_a
...
+            );
+
+        byte[] testPdfBytes1 = toByteArray(getResource("1111005.pdf"));
+        Pdf pdf1 = new Pdf("1111005.pdf", testPdfBytes1);
```
rkondratowicz: Is this the pdf file from 8_24-06-2018-00-00-00.zip?
If so, it would be good to use more meaningful variable names or add a comment.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -27,15 +34,24 @@ protected Processor(
         this.documentProcessor = documentProcessor;
         this.envelopeProcessor = envelopeProcessor;
         this.errorWrapper = errorWrapper;
+
+        validator = Validation
+            .byProvider(HibernateValidator.class)
+            .configure()
+            .buildValidatorFactory()
+            .getValidator();
```
rkondratowicz: Please keep it simple.
There's no need for a custom validation annotation which is tightly coupled to `ZipFileProcessor` and is used only once.
Just create a method.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ZipFileProcessorValidationTest.java
```diff
@@ -0,0 +1,141 @@
...
+        null,
+        null,
+        new ErrorHandlingWrapper((Throwable t) -> {
+            throw new RuntimeException(t);
+        })
+    ) {
+    };
+
+    @Parameterized.Parameters(name = "Processing {0}. Should fail on custom validator? {2}")
+    public static Iterable<Object[]> data() {
```
nitinprabhu: May be it's just me but this now looks bit difficult to read. It would have been nice if we could have created individual test methods for each of the case such as 
1. Scannable items has less entries compared to PDFs.
2. Scannable items has more entries compared to PDFs.
3. Scannable items has same entries but different file names in comparison to PDFs.
I think other scenarios need not be handled by this validation such as can't call dm, no metafile, non-pdf files etc. Those would be handled even before execution comes at this point.

PR #130, opened 2018-08-17
[PREVIEW]Use test containers docker compose instead of palantir docker compose rule

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-640

### Change description ###

- Used test containers docker compose containers instead of existing palantir docker compose rule as it was not tearing down containing properly.

- Can't use `@ClassRule` to automatically start and stop the container due to issue
https://github.com/testcontainers/testcontainers-java/issues/765

- Test containers compatible with version 2 of docker compose hence version needs to updated.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -157,10 +157,6 @@ repositories {
   jcenter()
 
   maven {
-    url 'https://dl.bintray.com/palantir/releases' // docker-compose-rule is published on bintray
-  }
-
-  maven {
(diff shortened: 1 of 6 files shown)


doncem:
perhaps we can remove the

```gradle
maven {
  url 'https://dl.bintray.com/palantir/releases' // docker-compose-rule is published on bintray
}
```

PR #135, opened 2018-08-23
[PREVIEW] Increment upload failure count

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/TaskErrorHandler.java
```diff
@@ -71,4 +76,10 @@ private void registerEvent(EventRelatedThrowable exception, String reason) {
             processEvent.getEvent()
         );
     }
+
+    private void incrementUploadFailureCount(DocUploadFailureGenericException exc) {
```
doncem: My guess would be this method is entirely envelope related - nothing additionally will be needed for exception handling therefore it can accept `Envelope` straight away. Current set of methods accepts exception because of generalisation

Just a thought

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/TaskErrorHandler.java
```diff
@@ -71,4 +76,10 @@ private void registerEvent(EventRelatedThrowable exception, String reason) {
...
+        Envelope envelope = exc.getEnvelope();
+        envelope.setUploadFailureCount(envelope.getUploadFailureCount() + 1);
```
doncem: Maybe worth having `public void incrementFailureCount()` method inside the `Envelope`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/TaskErrorHandler.java
rkondratowicz: Taking exception as param here for consistency as well. That's 3rd `if` already. We may want to use a map from exception to function that handles it soon (before PMD starts complaining ;))

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/TaskErrorHandler.java
rkondratowicz: Considering we're leaning toward replacing jpa with jdbc, I'd rather keep envelope model as 'thin' as possible.

doncem: let's do this! i mean the replacing 😉

PR #138, opened 2018-08-28
Read blobs using acquire lease

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-646

### Change description ###

- Reading  blobs by acquiring lease.

-  Lease time should be between 15 and 60 seconds which is enforced by Azure.

-  `leaseTimeInSeconds` will be made configurable in next PR.

- Removed Scheduler lock for blob processor.

- Added integration test to verify acquire lease.

```
[ ] Yes
[x ] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuite.java
```diff
@@ -134,7 +134,18 @@ protected void uploadZipToBlobStore(String zipFileName) throws Exception {
...
+
+        //Blob need to be deleted as same blob may exists if previously uploaded blob was not deleted
+        //due to doc upload failure
+        if (blockBlobReference.exists()) {
+            blockBlobReference.breakLease(0);
+            blockBlobReference.delete();
+        }
+
+        //A Put Blob operation may succeed against a blob that exists in the storage emulator with an active lease,
+        //even if the lease ID has not been specified in the request.
```
rkondratowicz: missing spaces after `//` :)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -74,22 +73,36 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
...
+                }
             }
         }
     }
 
+    private CloudBlockBlob acquireLease(CloudBlockBlob cloudBlockBlob, String containerName, String zipFilename) {
+        // TODO - Lease time  needs to be made configurable
+        return errorWrapper.wrapAcquireLeaseFailure(containerName, zipFilename, () -> {
+            //The value of the parameter 'leaseTimeInSeconds' should be between 15 and 60 seconds.
+            cloudBlockBlob.acquireLease(15, null);
```
rkondratowicz: instead of blindly trying to lease and throwing an exception, can we check if the files is already locked and only then acquire a lease for it?
Or, even better, acquire lease for first available file?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -46,6 +46,8 @@ protected void processParsedEnvelopeDocuments(
             documentProcessor.processPdfFiles(pdfs, envelope.getScannableItems());
             envelopeProcessor.markAsUploaded(envelope);
 
+            //Lease needs to be broken before deleting the blob.0 implies lease is broken immediately
+            cloudBlockBlob.breakLease(0);
```
rkondratowicz: Should we not also break the lease when something went wrong?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskWithAcquireLeaseTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        //When
+        Future<Void> future = processBlobUsingExecutor();
+
+        processor.processBlobs();
+
+        future.get(); // wait for completion of processor.processBlobs()
+
+        //Then
+        //One thread should not be able to acquire lease and other should fail
```
rkondratowicz: some missing spaces in this file as well

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskWithAcquireLeaseTest.java
```diff
@@ -0,0 +1,93 @@
...
+        assertThat(outputCapture.toString())
+            .contains("Lease already acquired for container test and zip file 1_24-06-2018-00-00-00.zip");
+
+        // We expect only one envelope which was uploaded and other failed
+        List<Envelope> envelopes = envelopeRepository.findAll();
+        assertThat(envelopes.size()).isEqualTo(1);
+
+        // and
+        List<ProcessEvent> processEvents = processEventRepository.findAll();
+        assertThat(processEvents).hasSize(2);
```
rkondratowicz: is this related? if so, worth adding a comment why `2` imho

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskWithAcquireLeaseTest.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    private void givenValidZipFileUploadedAndDocStoreMocked() throws Exception {
+        uploadZipToBlobStore(ZIP_FILE_NAME_SUCCESS);
+
+        byte[] test1PdfBytes = toByteArray(getResource("1111001.pdf"));
+        byte[] test2PdfBytes = toByteArray(getResource("1111002.pdf"));
+
+        Pdf pdf1 = new Pdf("1111001.pdf", test1PdfBytes);
+        Pdf pdf2 = new Pdf("1111002.pdf", test2PdfBytes);
```
rkondratowicz: probably copied from some other tests, but we can just inline those `XXbytes` variables, they're not used anywhere else.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
```diff
@@ -50,4 +53,30 @@ public void wrapDocUploadFailure(Envelope envelope, Supplier<Void> supplier) {
...
+            boolean isLeaseAlreadyAcquired = Optional.of(exception)
+                .filter(e -> e instanceof StorageException)
+                .map(se -> (StorageException) se)
+                .filter(se -> se.getHttpStatusCode() == HttpStatus.CONFLICT.value())
+                .isPresent();
+
+            Exception exc = isLeaseAlreadyAcquired
+                ? new LeaseAlreadyPresentException(
+                "Lease already acquired for container " + containerName + " and zip file " + zipFileName)
+                : exception;
```
rkondratowicz: perhaps this would be more idiomatic
```
if (exception instanceof StorageException && ((StorageException) exception).getHttpStatusCode == CONFLICT) {
    errorHandler.handleError(new LeaseAlreadyPresentException(...));
} else {
    errorHandler.handleError(exc);
}

File: src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java [hunk 8]
```diff
@@ -74,22 +73,36 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
...
-                    zipFileProcessor.getPdfs(),
-                    cloudBlockBlob
-                );
+
+        CloudBlockBlob blobWithLeaseAcquired = acquireLease(cloudBlockBlob, container.getName(), zipFilename);
+
+        if (Objects.nonNull(blobWithLeaseAcquired)) {
+            BlobInputStream blobInputStream = blobWithLeaseAcquired.openInputStream();
+
+            //Zip file will include metadata.json and collection of pdf documents
```
rkondratowicz: another missing space

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/LeaseAlreadyPresentException.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+public class LeaseAlreadyPresentException extends RuntimeException {
+    private static final long serialVersionUID = 6136150918143302940L;
+
+    public LeaseAlreadyPresentException(String message) {
```
rkondratowicz: worth adding a cause if we're wrapping some existing exception, we'll know where it was thrown originally

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
doncem: there's no benefit of try/catch signature. i would still vote for ^ approach but let's consider this one too:

```java
try {
  // smth to return
} catch (StorageException exception) {
    // more functionally comparison but it doesn't matter
    boolean isLeaseAcquired = HttpStatus.valueOf(exception.getCode()).equals(CONFLICT));

    errorHandler.handleError(isLeaseAcquired ? new LeaseAlreadyPresentException(...) : exception);
} catch (Exception exception) {
    errorHandler.handleError(exception);
}

return null;
```

p.s. the event will not be registered via error handler since exceptions are not `EventAwareThrowable`s. I think it is intended so all good. Just a note here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -46,6 +46,8 @@ protected void processParsedEnvelopeDocuments(
             documentProcessor.processPdfFiles(pdfs, envelope.getScannableItems());
             envelopeProcessor.markAsUploaded(envelope);
 
+            // Lease needs to be broken before deleting the blob.0 implies lease is broken immediately
```
doncem: single space or new line between sentences

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
nitinprabhu: comment was not part of my PR changes any ways will update it :)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -0,0 +1,117 @@
...
+        // We expect only two events one for doc upload and another for doc processed
+        List<ProcessEvent> processEvents = processEventRepository.findAll();
+        assertThat(processEvents).hasSize(2);
+    }
+
+    @NotNull
+    private List<Future<Void>> processBlobUsingExecutor(int numberofThreads) throws Exception {
+        ExecutorService executorService = Executors.newFixedThreadPool(
+            numberofThreads,
+            ThreadFactories.withName("BSP-ACQUIRE-LEASE-%d")
```
doncem: yes! would make more sense to just have `BSP-TEST` but it really doesn't matter 😉

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -0,0 +1,117 @@
...
+        futureTasks.forEach(future -> {
+            try {
+                future.get(); // wait for completion of processor.processBlobs()
+            } catch (Exception e) {
+                fail("Failed while processing blobs with exception " + e);
+            }
+        });
+
+        // Then
+        // Only one thread should not be able to acquire lease and others should fail
```
luigibk: "Only one thread should not be able to..." should read: "Only one thread should be able to..."

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
luigibk: BSP-ACQUIRE-LEASE-TEST I would say

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -74,22 +73,36 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
...
+                }
             }
         }
     }
 
+    private CloudBlockBlob acquireLease(CloudBlockBlob cloudBlockBlob, String containerName, String zipFilename) {
+        // TODO - Lease time  needs to be made configurable
+        return errorWrapper.wrapAcquireLeaseFailure(containerName, zipFilename, () -> {
+            // The value of the parameter 'leaseTimeInSeconds' should be between 15 and 60 seconds.
+            cloudBlockBlob.acquireLease(15, null);
```
luigibk: What happens if for some reason (such as high load) processing takes longer than 15secs? Does another thread acquire the lease and processes the file? Have we got a test for this?

nitinprabhu: Yes after 15 secs another thread can acquire the lease. Waiting for 15 secs in a test would be too long imo. We can rely on Azure storage for this and need not test that.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -0,0 +1,117 @@
...
+        assertThat(envelopes.size()).isEqualTo(1);
+
+        // and
+        // We expect only two events one for doc upload and another for doc processed
+        List<ProcessEvent> processEvents = processEventRepository.findAll();
+        assertThat(processEvents).hasSize(2);
+    }
+
+    @NotNull
+    private List<Future<Void>> processBlobUsingExecutor(int numberofThreads) throws Exception {
```
rkondratowicz: capital 'O'

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
luigibk: So if processing takes longer than 15secs we could have a situation in which a blob is processed twice? That is:
- T1 acquires lease and starts (long) processing.
- Lease expires
- T2 acquires lease and starts processing.
 Or is there a way to prevent that?

nitinprabhu: Single Blob processing should not take too much time(15 secs is too much) and if we expect and we know that if might take too long upfront then we should configure lease timeout > 15 secs or keep renewing lease. Unless proven that it takes longer than 15 secs I would not complicate it at this point as we need to know when to renew lease and where it is taking time.

luigibk: The problem I see is that, even if most of the time it will take less than 15secs, occasionally (e.g.  because of high load) it could take more than 15secs. We should have something in place that prevents the lease to be acquired again in those cases.

nitinprabhu: Like I said earlier unless we know where it takes time for example it might take time for doc upload then we should fix that rather than fixing lease. Feel free to raise a ticket as this is beyond the scope of this ticket.

PR #145, opened 2018-08-30
[PREVIEW] Removed hardcoded leasetimeout and moved to config

nitinprabhu:

https://tools.hmcts.net/jira/browse/RPE-646

### Change description ###

- Removed hardcoded lease timeout and moved it to spring configuration file.

- Configured terraform to set the value as an environment variable.

- Please note that I have autowired `leaseTimeout` property rather than adding new constructor argument. If we go with constructor approach then we would also need to update constructor of 
`FailedDocUploadProcessor` as  `FailedDocUploadProcessorTest` uses same Construct as `BlobProcessorTask`

```
[ ] Yes
[x ] No
```


src/main/resources/application.yaml
```diff
@@ -82,3 +82,8 @@ envelope-access:
      - jurisdiction: TEST
        readService: bulk_scan_processor_tests
        updateService: bulk_scan_processor_tests
+
+# The value of the parameter 'leaseTimeInSeconds' should be between 15 and 60 seconds.
+blob:
+  lease:
+    timeout: ${BLOB_LEASE_TIMEOUT:15}
```
rkondratowicz: did you consider moving it to settings for the 'scan' task? just a thought.

doncem: ah very true. could you please apply to this pr please?

PR #159, opened 2018-09-05
[PREVIEW] RPE-693 Refactor upload process

luigibk:
https://tools.hmcts.net/jira/browse/RPE-693

### Change description ###
Move pdf files check to doc error wrapper

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
@@ -184,11 +184,10 @@ public class EnvelopeControllerTest {
             .andExpect(MockMvcResultMatchers.jsonPath("envelopes[0].id").exists());
 
         assertThat(envelopeRepository.findAll())
-            .hasSize(2)
+            .hasSize(1)
             .extracting("zipFileName", "status")
             .containsExactlyInAnyOrder(
-                tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
-                tuple("8_24-06-2018-00-00-00.zip", METADATA_FAILURE)
+                tuple("7_24-06-2018-00-00-00.zip", PROCESSED)
             );
 
         byte[] mismatchingPdfBytes = toByteArray(getResource("1111005.pdf"));
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
@@ -54,7 +54,6 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
-import static uk.gov.hmcts.reform.bulkscanprocessor.entity.Status.METADATA_FAILURE;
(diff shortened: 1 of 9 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ZipFileProcessorValidationTest.java
```diff
@@ -24,27 +24,25 @@
 @RunWith(SpringRunner.class)
 public class ZipFileProcessorValidationTest {
 
-    private static final Processor processor = new Processor(
+    private static final EnvelopeProcessor envelopeProcessor = new EnvelopeProcessor(
```
rkondratowicz: class name is `ZipFileProcessorValidationTest`.
Seems like it's actually `EnvelopeProcessor` being tested here (or `Processor` previously).

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -184,11 +184,10 @@ public void should_successfully_return_all_envelopes_with_processed_status_for_a
...
 
         assertThat(envelopeRepository.findAll())
-            .hasSize(2)
+            .hasSize(1)
             .extracting("zipFileName", "status")
             .containsExactlyInAnyOrder(
-                tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
-                tuple("8_24-06-2018-00-00-00.zip", METADATA_FAILURE)
+                tuple("7_24-06-2018-00-00-00.zip", PROCESSED)
             );
```
rkondratowicz: Not sure if this assertion belongs in a test for a controller.
Especially for tests for reading envelopes.

Perhaps something for another round of refactoring.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/FileNameIrregularitiesException.java
```diff
@@ -5,7 +5,7 @@
 
 import java.util.Collection;
 
-public class FileNameIrregularitiesException extends RuntimeException implements EnvelopeAwareThrowable {
+public class FileNameIrregularitiesException extends RuntimeException implements EventRelatedThrowable {
```
doncem: But we are aware of the envelope 🤔 . Perhaps we should still mark as failure and let the the retry limit allow re-parsing this envelope next time? Just need to be sure that counter is ++'ed

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -184,11 +183,10 @@ public void should_successfully_return_all_envelopes_with_processed_status_for_a
             .andExpect(MockMvcResultMatchers.jsonPath("envelopes[0].id").exists());
 
         assertThat(envelopeRepository.findAll())
-            .hasSize(2)
+            .hasSize(1)
             .extracting("zipFileName", "status")
             .containsExactlyInAnyOrder(
-                tuple("7_24-06-2018-00-00-00.zip", PROCESSED),
-                tuple("8_24-06-2018-00-00-00.zip", METADATA_FAILURE)
```
rkondratowicz: Looks like `METADATA_FAILURE` status can be remove as envelopes in such state will never be created after this change. (and such envelopes can be deleted from db).

Ok for next PR imho.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/FileNameIrregularitiesException.java
luigibk: The envelope is not created anymore when the files do not match the metadata. This is essentially treated as you'd treat a metadata invalid case as the envelope would be unusable anyway if created.

PR #160, opened 2018-09-06
[PREVIEW] Added new fields to the validation schema

aliveni:
### Added new fields to envelope, non_scannable_items and payments tables ###

```
[ ] Yes
[x ] No
```


src/main/resources/db/migration/V013_Add_new_fields.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE envelopes
+  ADD COLUMN caseNumber VARCHAR(100) NULL;
+
+ALTER TABLE envelopes
+  ADD COLUMN documentControlNumber VARCHAR(100);
+
+ALTER TABLE payments
+  ADD COLUMN paymentInstrumentNumber VARCHAR(100),
```
rkondratowicz: it would be nice to put `NULL` here as well for consistency imho

src/main/resources/metafile-schema.json
```diff
@@ -162,6 +169,21 @@
...
             "title": "ISO 4217 Currency CodeISO 42",
             "pattern": "^[A-Z]{3}$"
+          },
+          "payment_instrument_number": {
+            "id": "#/properties/payments/items/properties/payment_instrument_number",
+            "type": "string",
+            "title": "Cheque number in case of cheques ; postal order number in case of postal orders"
+          },
+          "sort_code": {
+            "id": "#/properties/payments/items/properties/sort_code",
```
nitinprabhu: I think we will need validation such that if payment type is cheque then sort code should be present ?

src/main/resources/metafile-schema.json
```diff
@@ -162,6 +169,21 @@
...
+            "type": "string",
+            "title": "The sort code in case of cheque payments otherwise empty"
+          },
+          "account_number": {
+            "id": "#/properties/payments/items/properties/account_number",
```
nitinprabhu: same here if payment type is cheque then sort code should be present ?

src/main/resources/metafile-schema.json
```diff
@@ -192,7 +220,11 @@
             "type": "string",
             "title": "contextual information describing this item"
           }
-        }
+        },
+        "required": [
+          "document_control_number",
+          "item_type"
```
nitinprabhu: not sure if item type is mandatory ?

src/main/resources/db/migration/V013_Add_new_fields.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE envelopes
```
nitinprabhu: If fields are mandatory then they should be not null ? We will also need to populate default values if field are mandatory as part of migration.

src/main/resources/metafile-schema.json
aliveni: updated schema with validation and added unit tests

src/main/resources/metafile-schema.json
aliveni: item type is added in required fields list in the confluence. we can discuss later

src/main/resources/db/migration/V013__Add_new_fields.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE envelopes
+  ADD COLUMN caseNumber VARCHAR(100) NULL;
+
+ALTER TABLE non_scannable_items
+  ADD COLUMN documentControlNumber VARCHAR(100) NOT NULL;
+
+ALTER TABLE payments
+  ADD COLUMN paymentInstrumentNumber VARCHAR(100),
```
rkondratowicz: let's add `NULL` / `NOT NULL` for consistency

src/main/resources/db/migration/V013__Add_new_fields.sql
```diff
@@ -0,0 +1,13 @@
...
+  ADD COLUMN sortCode VARCHAR(10) NULL DEFAULT 112233,
+  ADD COLUMN accountNumber VARCHAR(50) NULL DEFAULT 12345678;
```
rkondratowicz: why the defaults if column in nullable?

aliveni: I added it for "Cheque" payment methods.  but I removed Default values and added an update statement

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -105,7 +105,7 @@ public void should_process_zip_file_only_once_by_processor_thread_which_acquires
     }
 
     private void givenValidZipFileUploadedAndDocStoreMocked() throws Exception {
-        uploadZipToBlobStore(ZIP_FILE_NAME_SUCCESS);
+        uploadZipToBlobStore(ZIP_FILE_NAME_SUCCESS_WITHOUT_CASE_NUMBER);
```
nitinprabhu: Probably name it as `VALID_ZIP_FILE_WITHOUT_CASE_NUMBER`

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTest.java
```diff
@@ -54,6 +65,40 @@ public void should_parse_envelope_data_with_no_non_scannable_items_in() throws I
         assertThat(envelope.getPayments()).hasSize(1);
     }
 
+    @Test
+    public void should_parse_envelope_data_with_no_account_details_for_postal_payment() throws IOException {
+        Envelope envelope = getEnvelope("/metafiles/valid/postal-payment-method.json");
+
+        assertThat(envelope.getPayments()).hasSize(1);
+        assertThat(envelope.getPayments().get(0).getAmount()).isEqualTo(100.0);
```
nitinprabhu: Have you considered scenario where you may have two payment types for e.g solicitor may send cheque and postal order(worth checking with Ade)?
get(0) will fail if you have two payment methods in a single envelope.

src/test/resources/metafiles/invalid/invalid-scannable-items.json
```diff
@@ -0,0 +1,39 @@
...
+    {
+      "document_control_number": "1111002",
+      "method": "Cash",
+      "amount": "100.00",
+      "currency": "GBP"
+    }
+  ],
+  "non_scannable_items": [
+    {
+      "document_control_number": "1111002",
```
nitinprabhu: this number should be different than one used in payment and also there should be a section in scannable items referring to this number.

src/test/resources/metafiles/invalid/invalid-postal-order-payment.json
```diff
@@ -0,0 +1,40 @@
...
+    {
+      "document_control_number": "1111002",
+      "method": "Postal Order",
+      "amount": "100.00",
+      "currency": "GBP"
+    }
+  ],
+  "non_scannable_items": [
+    {
+      "document_control_number": "1111002",
```
nitinprabhu: Check below comment

src/test/resources/metafiles/invalid/use-of-numeric.json
```diff
@@ -36,11 +37,13 @@
...
       "method": "Cheque",
       "amount": 100.00,
-      "currency": "GBP"
+      "currency": "GBP",
+      "payment_instrument_number": "1000000000"
     }
   ],
   "non_scannable_items": [
     {
+      "document_control_number": "1111002",
```
nitinprabhu: use a different number and have a entry in scannable item

src/test/resources/metafiles/valid/cash-payment-method.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "document_control_number": "1111002",
+      "method": "Cash",
+      "amount": "100.00",
+      "currency": "GBP"
+    }
+  ],
+  "non_scannable_items": [
+    {
+      "document_control_number": "1111002",
```
nitinprabhu: same for this and all files `this number should be different than one used in payment and also there should be a section in scannable items referring to this number.`

src/main/resources/db/migration/V013__Add_new_fields.sql
```diff
@@ -0,0 +1,18 @@
...
+ALTER TABLE non_scannable_items
+  ADD COLUMN documentControlNumber VARCHAR(100) NOT NULL;
+
+ALTER TABLE payments
+  ADD COLUMN paymentInstrumentNumber VARCHAR(100) NULL,
+  ADD COLUMN sortCode VARCHAR(10) NULL,
+  ADD COLUMN accountNumber VARCHAR(50) NULL;
+
+ALTER TABLE scannable_items
+  ALTER COLUMN documentControlNumber SET NOT NULL;
```
rkondratowicz: 😨

src/main/resources/db/migration/V013__Add_new_fields.sql
```diff
@@ -0,0 +1,18 @@
...
+
+UPDATE payments
+  SET sortCode = '112233',
+      accountNumber = '12345678'
```
rkondratowicz: comment explaining why would be nice.
Also, if you set those value, should this column be `NOT NULL`?

aliveni: @rkondratowicz The mandatory fields are different for each payment method. sortcode and account number fields are mandatory only for 'Cheque' payment. So the above update statement is to provide default values for existing Cheque payments data.

PR #161, opened 2018-09-07
[PREVIEW] Retrieve connection string from shared infra

timja:
https://tools.hmcts.net/jira/browse/RPE-680

### Change description ###
Retrieve connection string from shared infra

```
[ ] Yes
[ ] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,6 @@ dependencyManagement {
 }
 
 repositories {
-  mavenCentral()
   jcenter()
 
   maven {
--- a/build.gradle
+++ b/build.gradle
@@ -154,6 +154,7 @@ dependencyManagement {
 }
 
 repositories {
+  mavenCentral()
   jcenter()
 
   maven {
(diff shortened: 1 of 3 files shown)


src/apiGatewayTest/java/uk/gov/hmcts/reform/bulkscanprocessor/api/GetSasTokenTest.java
```diff
@@ -59,6 +59,7 @@ public void should_reject_request_lacking_subscription_key() throws Exception {
         assertThat(response.body().asString()).contains("Access denied due to missing subscription key");
     }
 
+    @Test
```
nitinprabhu: ahh this was missing so it will be nice to see if the test works consistently now 😉

timja:
sure can do, its straight copy paste from orchestrator which worked fine on preview

build.gradle
```diff
@@ -154,6 +154,7 @@ dependencyManagement {
 }
 
 repositories {
+  mavenCentral()
```
rkondratowicz: ?

timja: trying it to get a green build, it means we try mavenCentral first for dependencies.
it doesn't help with plugins though which are only in bintray

PR #163, opened 2018-09-07
[PREVIEW] RPE-693 Refactor blob delete

luigibk:
https://tools.hmcts.net/jira/browse/RPE-693

### Change description ###
Refactor blob delete

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -257,6 +257,7 @@ public class BlobProcessorTaskTest extends ProcessorTestSuite<BlobProcessorTask>
     @After
     public void cleanUp() {
         envelopeRepository.deleteAll();
+        processEventRepository.deleteAll();
     }
 
 }
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Resources;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -12,8 +13,10 @@ import org.springframework.test.context.junit4.SpringRunner;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Event;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.ProcessEvent;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Status;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocumentNotFoundException;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.UnableToUploadDocumentException;
+import uk.gov.hmcts.reform.bulkscanprocessor.helper.EnvelopeCreator;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.document.output.Pdf;
 
 import java.nio.charset.Charset;
@@ -28,6 +31,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.tuple;
 import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
```diff
@@ -40,15 +41,16 @@ public ErrorHandlingWrapper(ErrorHandler errorHandler) {
         }
     }
 
-    public void wrapDocUploadFailure(Envelope envelope, Supplier<Void> supplier) {
+    public Boolean wrapDocUploadFailure(Envelope envelope, Supplier<Boolean> supplier) {
```
rkondratowicz: now that this function returns a vault, a javadoc comment would be nice

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
```diff
@@ -40,15 +41,16 @@ public ErrorHandlingWrapper(ErrorHandler errorHandler) {
...
         try {
-            supplier.get();
+            return supplier.get();
```
rkondratowicz: perhaps returning `true` could be here?
( operation succeeded -> always return true)

rkondratowicz: instead of always returning true in the supplier passed to the wrapper

luigibk: The supplier should return something anyway because of its contract for what I can see, so I am not sure what you suggest would make it any more readable

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
luigibk: That's done.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -232,4 +236,28 @@ public void should_increment_failure_count_when_unable_to_upload_file() throws E
         Envelope envelope = envelopes.get(0);
         assertThat(envelope.getUploadFailureCount()).isEqualTo(1);
     }
+
+    @Test
+    public void should_not_process_again_if_blob_delete_failure()
```
rkondratowicz: 'failed'

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -232,4 +236,28 @@ public void should_increment_failure_count_when_unable_to_upload_file() throws E
...
+        throws Exception {
+        //Given
+        uploadZipToBlobStore(ZIP_FILE_NAME_SUCCESS); //Zip file with metadata and pdfs
+
+        Envelope existingEnvelope = EnvelopeCreator.envelope("A", Status.DELETE_BLOB_FAILURE);
```
rkondratowicz: is it an existing envelope or a new one? surely the UUID will be different?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/wrapper/ErrorHandlingWrapper.java
```diff
@@ -40,15 +41,23 @@ public ErrorHandlingWrapper(ErrorHandler errorHandler) {
         }
     }
 
-    public void wrapDocUploadFailure(Envelope envelope, Supplier<Void> supplier) {
+    /**
+     * Wraps docs upload to handles related errors.
+     *
+     * @param envelope for the documents to upload
+     * @param supplier upload function
+     * @return true if upload successful
```
rkondratowicz: atm (sadly) it return whatever the passed supplier returns.
But it always returns `false` if running the `supplier` threw an exception.

luigibk: Yes, if the upload failed should definitely return false. Otherwise the supplier should know what to return

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
luigibk: Changed this to 'failed'

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
luigibk: It simulates an existing envelope by creating one before processing the blob. The check is done using the blob container + filename (not the uuid).

rkondratowicz: ok, in that case, imho, it would be more clear to save envelope in db first and then upload a zip to blob storage so that it's processed. (ideally with a comment)

luigibk: That is a good point. Actually the test should check the blob has been deleted and status has been updated. Which is what I've just added.

rkondratowicz:
coverage: 59.45% of diff hit (target 80%)

PR #176, opened 2018-09-17
[PREVIEW] Send envelope message

luigibk:
RPE-648

### Change description ###
Send envelope message to service bus for consumption by orchestrator.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Event.java
```diff
@@ -10,4 +10,5 @@
     DOC_CONSUMED, // client service handled the documents
     BLOB_DELETE_FAILURE, // when blob is not successfully deleted after processing is complete
```
rkondratowicz: not related to this PR, but is this still needed/used?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -33,18 +37,27 @@ protected Processor(
...
             return;
         }
-        markAsProcessed(envelope);
+        if (!markAsProcessed(envelope)) {
+            return;
+        }
+        if (!sendProcessedMessage(serviceBusHelper, envelope)) {
+            return;
+        }
+        markAsNotified(envelope);
```
rkondratowicz: `markAsProcessed` is wrapped, this is not.
Is this by design?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Event.java
luigibk: That's an event not an envelope status. It is still used as we might want to track how many times blob delete fails.

rkondratowicz: ok

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
luigibk: They are all wrapped by handleEvent() for what I can see.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -78,7 +78,7 @@ public void should_process_zip_file_only_once_by_processor_thread_which_acquires
         // and
         // We expect only two events one for doc upload and another for doc processed
         List<ProcessEvent> processEvents = processEventRepository.findAll();
-        assertThat(processEvents).hasSize(2);
+        assertThat(processEvents).hasSize(3);
```
rkondratowicz: Comment in line 80 needs updating.
Or consider updating this test to check the items, not just the count.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -53,6 +55,16 @@ public BlobProcessorTask(
...
 
+    /**
+     * Spring overrides the {@code @Lookup} method and returns an instance of bean.
+     *
+     * @return Instance of {@code ServiceBusHelper}
+     */
+    @Lookup
+    public ServiceBusHelper serviceBusHelper() {
+        return null;
+    }
```
rkondratowicz: can we inject this dependency the same way as all the other ones? (via constructor)

luigibk: This way a new service bus client is created for every run of the task. If we inject it like the other dependencies there will only be a single client for the entire lifetime of the application. For the time being I think this is safer as the client gets renewed.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
luigibk: It is definitely wrapped.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/DocProcessedNotificationFailureException.java
```diff
@@ -0,0 +1,43 @@
...
+    private final transient Envelope envelope;
+
+    public DocProcessedNotificationFailureException(Envelope envelope) {
+        super("Message send failure");
+
+        this.envelope = envelope;
+    }
+
+    public DocProcessedNotificationFailureException(Envelope envelope, Throwable cause) {
+        super(cause.getMessage(), cause);
```
lgonczar: I'd probably set a fixed message here, but don't feel strong about it.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: I agree - used to have a problem with the client not recovering after a broken connection.

PR #177, opened 2018-09-19
[PREVIEW] Fix blob delete bug

luigibk:
### Change description ###
Fix blob delete retry issue. A blob is currently deleted in some instances even if processing is not complete.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -14,6 +14,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Status;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -106,7 +107,7 @@ private void processZipFile(CloudBlobContainer container, String zipFilename)
 
     private void delete(CloudBlockBlob cloudBlockBlob, Envelope envelope) {
         try {
-            if (cloudBlockBlob != null) {
+            if (cloudBlockBlob != null && envelope.getStatus() == Status.PROCESSED) {
```
nitinprabhu: Delete method should not be called if envelope hasn' t been processed ?

luigibk: Not at the moment as we have moved from a model where all failures were upload failures to a model with different failures. So for the time being to avoid issues (such as the current build failure in master) we'd better be stricter. Later this can be modified as the file can be deleted earlier, but re-processing partially processed files needs also work.

nitinprabhu: I am failing to understand why we are doing this and that too because of build failure ? I think irrespective of the model we choose if previous step fails this step should not executed at all. I think we need to review this. Happy to discuss this in person or on slack if needed :)

PR #178, opened 2018-09-20
[PREVIEW] Introduce polling interval to speed up tests which wait for an event

luigibk:
### Change description ###
Introduce polling interval to speed up tests which wait for an event

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -202,7 +202,6 @@ public class BlobProcessorTaskTest extends ProcessorTestSuite<BlobProcessorTask>
         CloudBlockBlob blob = testContainer.getBlockBlobReference(zipFile);
         await("file should not be deleted")
             .timeout(2, SECONDS)
-            .pollInterval(1, TimeUnit.SECONDS)
             .until(blob::exists, is(true));
 
         // Verify envelope status is updated to UPLOAD_FAILED
@@ -303,7 +302,6 @@ public class BlobProcessorTaskTest extends ProcessorTestSuite<BlobProcessorTask>
         CloudBlockBlob blob = testContainer.getBlockBlobReference(VALID_ZIP_FILE_WITH_CASE_NUMBER);
         await("file should not be deleted")
             .atMost(5, SECONDS)
-            .pollInterval(1, TimeUnit.SECONDS)
             .until(blob::exists, is(true));
     }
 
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -200,6 +202,7 @@ public void should_keep_zip_file_after_unsuccessful_upload_and_not_create_doc_pr
         CloudBlockBlob blob = testContainer.getBlockBlobReference(zipFile);
         await("file should not be deleted")
             .timeout(2, SECONDS)
+            .pollInterval(1, TimeUnit.SECONDS)
```
rkondratowicz: Here we need to wait the full time (the background job may have not even checked the file)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -299,6 +303,7 @@ public void should_not_delete_blob_unless_processed()
         CloudBlockBlob blob = testContainer.getBlockBlobReference(VALID_ZIP_FILE_WITH_CASE_NUMBER);
         await("file should not be deleted")
             .atMost(5, SECONDS)
+            .pollInterval(1, TimeUnit.SECONDS)
```
rkondratowicz: same, here. Check file still exists after a while

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
luigibk: In fact it waits until the timeout expires unless the file is deleted before that. This way it just the condition more often...

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
luigibk: Same as above. It waits the full timeout unless there is no need for it.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
rkondratowicz: To clarify, here we check that something *didn't happen*, so we can't check immediately.
Other cases are fine.

PR #181, opened 2018-09-21
[PREVIEW] RPE-733 Updated EnvelopeMsg model to include envelope data

aliveni:
https://tools.hmcts.net/jira/browse/RPE-733

### Change description ###

Added envelope data to the Envelope Message model. 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
@@ -19,6 +19,7 @@ import uk.gov.hmcts.reform.bulkscanprocessor.entity.ScannableItem;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidMessageException;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.EnvelopeMsg;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.Msg;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.MsgLabel;
 
 import java.util.Arrays;
 import java.util.UUID;
@@ -88,6 +89,15 @@ public class ServiceBusHelperTest {
         when(envelope.getId()).thenReturn(null);
         Msg msg = new EnvelopeMsg(envelope);
         serviceBusHelper.sendMessage(msg);
+    }
+
+    @Test
+    public void should_add_test_label_to_test_message() {
+        when(envelope.getZipFileName()).thenReturn("zip-file.test.zip");
+        when(envelope.isTestOnly()).thenReturn(true);
+        Msg msg = new EnvelopeMsg(envelope);
+        Message busMessage = serviceBusHelper.mapToBusMessage(msg);
+        assertThat(busMessage.getLabel()).isEqualTo(MsgLabel.TEST.toString());
     }
 
     @Test
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
@@ -59,7 +59,7 @@ public class ServiceBusHelperTest {
         when(queueClientSupplier.get()).thenReturn(this.queueClient);
         serviceBusHelper = new ServiceBusHelper(queueClientSupplier, this.objectMapper);
 
-        mockEnevelopeData();
+        mockEnvelopeData();
     }
 
     @Test
@@ -127,7 +127,7 @@ public class ServiceBusHelperTest {
         assertThat(docUrls.get(1).asText()).isEqualTo(scannableItem2.getDocumentUrl());
     }
 
-    private void mockEnevelopeData() {
+    private void mockEnvelopeData() {
         when(envelope.getId()).thenReturn(envelopeId);
         when(envelope.getCaseNumber()).thenReturn("1111222233334446");
         when(envelope.getJurisdiction()).thenReturn("SSCS");
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
@@ -100,21 +100,21 @@ public class ServiceBusHelperTest {
     @Test
     public void should_send_message_with_envelope_data() throws Exception {
 
-        Msg message = new EnvelopeMsg(envelope);
+        EnvelopeMsg message = new EnvelopeMsg(envelope);
         Message busMessage = serviceBusHelper.mapToBusMessage(message);
 
         JsonNode jsonNode = objectMapper.readTree(busMessage.getBody());
 
-        assertThat(jsonNode.get("case_ref").textValue()).isEqualTo("1111222233334446");
-        assertThat(jsonNode.get("jurisdiction").textValue()).isEqualTo("SSCS");
-        assertThat(jsonNode.get("zip_file_name").textValue()).isEqualTo("zip-file-test.zip");
-        assertThat(jsonNode.get("classification").textValue()).isEqualTo(Classification.EXCEPTION.name());
+        assertThat(jsonNode.get("case_ref").textValue()).isEqualTo(message.getCaseNumber());
+        assertThat(jsonNode.get("jurisdiction").textValue()).isEqualTo(message.getJurisdiction());
+        assertThat(jsonNode.get("zip_file_name").textValue()).isEqualTo(message.getZipFileName());
+        assertThat(jsonNode.get("classification").textValue()).isEqualTo(message.getClassification().name());
 
         JsonNode docUrls = jsonNode.get("doc_urls");
         assertThat(docUrls.isArray()).isTrue();
         assertThat(docUrls.size()).isEqualTo(2);
-        assertThat(docUrls.get(0).asText()).isEqualTo("documentUrl1");
-        assertThat(docUrls.get(1).asText()).isEqualTo("documentUrl2");
+        assertThat(docUrls.get(0).asText()).isEqualTo(scannableItem1.getDocumentUrl());
+        assertThat(docUrls.get(1).asText()).isEqualTo(scannableItem2.getDocumentUrl());
     }
 
     private void mockEnevelopeData() {
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
@@ -10,15 +10,15 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Classification;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.entity.ScannableItem;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidMessageException;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.EnvelopeMsg;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.Msg;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -34,12 +64,36 @@ public boolean isTestOnly() {
...
+        return documentUrls;
+    }
+
     @Override
     public String toString() {
         return "EnvelopeMsg{"
-            + "envelopeId='" + envelopeId + '\''
-            + "testOnly='" + testOnly + '\''
-            + '}';
+            + "envelopeId='" + envelopeId + "'"
```
timja: maybe we could use this so we don't have to maintain toString fields?
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/builder/ReflectionToStringBuilder.html

Example of using it:
https://github.com/hmcts/cmc-claim-store/blob/40234e54341097e7c21473cb23138038bbe5032e/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/models/Timeline.java#L33
https://github.com/hmcts/cmc-claim-store/blob/812f7aef900c09be93d3ebc57b3fd21dc471e0a8/domain-model/src/main/java/uk/gov/hmcts/cmc/domain/utils/ToStringStyle.java#L10

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
```diff
@@ -27,7 +29,9 @@
 
     private final IQueueClient sendClient;
 
-    public ServiceBusHelper(QueueClientSupplier queueClientSupplier) {
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
timja: Could we inject the object mapper so that we get consistent jackson configuration?
I think that just instantiating it means the we get the default object mapper from jackson not the configured one from spring?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -1,30 +1,60 @@
...
+    @JsonProperty("doc_urls")
+    private List<String> documentUrls;
 
-    public EnvelopeMsg(String envelopeId, boolean testOnly) {
-        this.envelopeId = envelopeId;
+    @JsonIgnore
+    private final byte[] envelope;
+
+    @JsonIgnore
+    private final boolean testOnly;
```
nitinprabhu: I think this field is not required any more. Please check `bulk-scan-orchestrator`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
nitinprabhu: Let's not log all the details here for e.g caseNumber. For now we could just log zip filename and envelope id.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -1,30 +1,60 @@
...
+
+    public EnvelopeMsg(Envelope envelope, boolean testOnly) {
```
nitinprabhu: I think we will need to add zip filename to this model class.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -1,30 +1,60 @@
...
-    public EnvelopeMsg(String envelopeId) {
-        this(envelopeId, false);
-    }
+    @JsonProperty("doc_urls")
+    private List<String> documentUrls;
 
-    public EnvelopeMsg(String envelopeId, boolean testOnly) {
-        this.envelopeId = envelopeId;
+    @JsonIgnore
+    private final byte[] envelope;
```
nitinprabhu: Do we need this field ? can we not just set all the required fields from envelope object to envelope msg model object ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
nitinprabhu: metafile validator uses a local object mapper as it needs custom configuration but otherwise like @timja pointed out we can use default one provided by spring.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
luigibk: I think that field is still needed indeed. Better check bulk-scan-orchestrator.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/Msg.java
```diff
@@ -8,5 +11,13 @@
...
+
+    String getCaseNumber();
+
+    Classification getClassification();
+
+    String getJurisdiction();
+
+    List<String> getDocumentUrls();
+
 }
```
luigibk: These changes defeat the purpose of having a common interface among messages. The new fields should only be in EnvelopeMsg and getMsgBody() should take care of placing them in the byte[].

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
nitinprabhu: As far I know it was removed so yes better check bulk-scan-orchestrator 😉 https://github.com/hmcts/bulk-scan-orchestrator/blob/master/src/main/java/uk/gov/hmcts/reform/bulkscan/orchestrator/services/EnvelopeEventProcessor.java

luigibk: It was removed in its current implementation, as we are changing the messages substantially, but I believe there is still a need for a test message so that during tests, cases are not created in CCD. Rafal is taking care of this in the orchestrator so he's probably the best person to answer that question

nitinprabhu: Tests are executed on AAT so we would want them to get created on AAT. Why do you think we should not create cases ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/Msg.java
aliveni: @luigibk Sure, I'm assuming the receiving end will take care of unmarshalling the message to the correct subtype.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
aliveni: @nitinprabhu Sure, I'll add zip filename and the log message. Is the "zip file name" used just for logging or in the queue Message as well?

nitinprabhu: yes it is needed in queue message and also need to make sure it is in orchestrator as well :)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/Msg.java
luigibk: @aliveni yes that is right. What I am saying is different. The serviceBusHelper encapsulates (and isolates) the Azure-ServiceBus client dependency. The send and sendAsync methods there take a Msg, which is a generic interface that can be converted to a BusMessage, the Azure client representation of a message. If you introduce all those fields in the Msg interface, instead of just a body, all the implementors of the Msg interface, going forward, will have to have those fields, which is a rather strict requirement. What I would suggest is that the fields are only present in EnvelopeMsg and in that class you implement Msg#getBody(), which will convert those fields in the message body.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
luigibk: Because we will want to avoid hitting ccd in some test cases? Also this way we can implement syntethic transactions to monitor that the processor and orchestrator can talk while in prod (without creating cases in ccd prod).

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
aliveni: No, we don't need it anymore. I'll construct the queue message body in getMsgBody() as per Luigi's comment.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -1,45 +1,87 @@
...
+    private String caseNumber;
+
+    @JsonProperty("jurisdiction")
+    private String jurisdiction;
 
-    private final boolean testOnly;
+    @JsonProperty("classification")
+    private Classification classification;
 
+    @JsonProperty("zipFileName")
```
rkondratowicz: `zip_file_name`

rkondratowicz: (for consistency)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
```diff
@@ -61,22 +85,47 @@ public void should_send_message_with_messageId() throws Exception {
...
+    public void should_send_message_with_envelope_data() throws Exception {
+
+        Msg message = new EnvelopeMsg(envelope);
+        Message busMessage = serviceBusHelper.mapToBusMessage(message);
+
+        JsonNode jsonNode = objectMapper.readTree(busMessage.getBody());
+
+        assertThat(jsonNode.get("case_ref").textValue()).isEqualTo("1111222233334446");
+        assertThat(jsonNode.get("jurisdiction").textValue()).isEqualTo("SSCS");
+        assertThat(jsonNode.get("zipFileName").textValue()).isEqualTo("zip-file-test.zip");
```
rkondratowicz: perhaps we could use the values retrieved from `envelope`? Otherwise it's hard to tell, why exactly `1111222233334446` etc is the correct value.

aliveni: Sure

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
aliveni: renamed the property

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
```diff
@@ -61,22 +86,56 @@ public void should_send_message_with_messageId() throws Exception {
...
     @Test
     public void should_not_add_any_label_to_standard_message() {
-        Msg msg = new EnvelopeMsg("envelopeId");
+        Msg msg = new EnvelopeMsg(envelope);
         Message busMessage = serviceBusHelper.mapToBusMessage(msg);
         assertThat(busMessage.getLabel()).isNullOrEmpty();
     }
 
+    @Test
+    public void should_send_message_with_envelope_data() throws Exception {
```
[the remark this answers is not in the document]
rkondratowicz: you could check what was passed to `sendClient.send(...)`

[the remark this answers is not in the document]
aliveni: yeah, updated the test

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
```diff
@@ -61,22 +86,56 @@ public void should_send_message_with_messageId() throws Exception {
...
-        Msg msg = new EnvelopeMsg("");
+        when(envelope.getId()).thenReturn(null);
+        Msg msg = new EnvelopeMsg(envelope);
         serviceBusHelper.sendMessage(msg);
     }
 
     @Test
     public void should_add_test_label_to_test_message() {
-        Msg msg = new EnvelopeMsg("envelopeId", true);
+        when(envelope.getZipFileName()).thenReturn("zip-file.test.zip");
```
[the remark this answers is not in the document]
aliveni: I removed it now

PR #182, opened 2018-09-21
[PREVIEW] Error handling RFC

luigibk:
### Change description ###
More than a PR this is a request for comment.

This refactoring introduces a functional decorator pattern into error handling. All error handler methods can expose a common signature and can be composed as functions. It is not the simplest of implementations even if in my view it does increase clarity and maintainability. 

I'd like to get some feedback. Thanks for reading! 

```
[ ] Yes
[x] No
```


doncem:
I believe this now becomes obsolete after/since #276

PR #196, opened 2018-09-28
Clean up test

rkondratowicz:
- reorder fields so that they are grouped by type (mocks, autowired deps, actual tested class)
- remove not needed constant


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
```diff
@@ -37,28 +37,16 @@
 @SpringBootTest
 public class ServiceBusHelperTest {
 
-    @Mock
-    private IQueueClient queueClient;
+    @Autowired private ObjectMapper objectMapper;
```
KevinJCross: ok soooo ... which way is the standard?
inline Annotations or separate line annotations.
I really dont mind but lets be consistent ?

PR #201, opened 2018-10-03
[PREVIEW] Lease dest filename

luigibk:
### Change description ###
Fix for functional test GetSasTokenTest failures caused by leases acquired by other parts of the application while the test is running. See:

https://build.platform.hmcts.net/job/HMCTS_Platform/job/bulk-scan-processor/job/master/41/

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
@@ -5,7 +5,6 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Resources;
-import com.microsoft.azure.storage.AccessCondition;
 import com.microsoft.azure.storage.blob.CloudBlobContainer;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
 import com.microsoft.azure.storage.core.PathUtility;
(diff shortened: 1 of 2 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -77,6 +77,15 @@ public void uploadZipFile(
         blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length);
     }
 
+    public void uploadAndLeaseZipFile(
+        CloudBlobContainer container, String srcZipFilename, String destZipFilename
+    ) throws Exception {
+        byte[] zipFile = Resources.toByteArray(Resources.getResource(srcZipFilename));
+        CloudBlockBlob blockBlobReference = container.getBlockBlobReference(destZipFilename);
+        blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length);
+        blockBlobReference.acquireLease();
```
nitinprabhu: This is not going to work consistently . You are uploading the file and then acquiring the lease which sometime may work if the process is able to acquire the lock else it will fail as other processes(jobs) may be running.

You will need to do something as below

```

 CloudBlockBlob blockBlobReference = container.getBlockBlobReference(destZipFilename);
        blockBlobReference.acquireLease(15,"1234");
        blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length, AccessCondition.generateLeaseCondition("1234"),null,null);
....
```

luigibk: @nitinprabhu I agree with you on the race condition and I tried already what you suggest but I got an error as the blob had not been created when yet when requesting the lease. For now this is better than what we currently have.

nitinprabhu: ok makes sense to acquire lease after blob is uploaded. Have you tried below ?

```

CloudBlockBlob blockBlobReference = container.getBlockBlobReference(destZipFilename);
        blockBlobReference.uploadFromByteArray(zipFile, 0, zipFile.length, AccessCondition.generateLeaseCondition("1234"),null,null);
        blockBlobReference.acquireLease(15,"1234");```

luigibk:
@nitin your 2nd idea (i.e. using AccessCondition.generateLeaseCondition(leaseId)) looked promising but unfortunately it has the same issue (cannot be done before blob creation). See:
 
https://build.platform.hmcts.net/job/HMCTS_Platform/job/bulk-scan-processor/view/change-requests/job/PR-201/6/

PR #204, opened 2018-10-04
Clearer doc urls exception

rkondratowicz:
- rename
- inlude all files that were not found in response from document management service


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/DocumentProcessor.java
```diff
@@ -33,14 +35,18 @@ public void uploadPdfFiles(List<Pdf> pdfs, List<ScannableItem> scannedItems) {
...
-                throw new DocumentNotFoundException("Document metadata not found for file " + item.getFileName());
-            }
-        });
-
-        scannableItemRepository.saveAll(scannedItems);
+        List<String> filesWithoutUrl =
+            scannedItems
+                .stream()
+                .map(item -> item.getFileName())
+                .filter(fileName -> !response.keySet().contains(fileName))
```
KevinJCross: it would Probably better to get the response keyset out of the filter otherwise it could be quite expensive creating a new set for each filename
or create to sets and use some set maths something like :
```java
missingFiles = sets.diff( files, response.keys)
if ( missingfiles.isEmpty()){
   attachFiles()
}else{
  throw ExceptionWithMissingFiles(missingFiles);
}
```

http://bethecoder.com/applications/tutorials/tools-and-libs/google-guava/collections/sets-difference.html

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/DocumentProcessor.java
```diff
@@ -33,14 +35,18 @@ public void uploadPdfFiles(List<Pdf> pdfs, List<ScannableItem> scannedItems) {
...
-        scannedItems.forEach(item -> {
-            if (response.containsKey(item.getFileName())) {
-                item.setDocumentUrl(response.get(item.getFileName()));
-            } else {
-                throw new DocumentNotFoundException("Document metadata not found for file " + item.getFileName());
-            }
-        });
-
-        scannableItemRepository.saveAll(scannedItems);
+        List<String> filesWithoutUrl =
```
KevinJCross: I personally would put this in a method to simplify this method a bit and have a clear responsibility

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/DocumentProcessor.java
rkondratowicz: done

PR #206, opened 2018-10-05
AKS initial setup

luigibk:
https://tools.hmcts.net/jira/browse/RPE-694

### Change description ###
Initial Setup. Add AKS running of functional tests to bulk scanning

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -120,8 +120,8 @@ withPipeline(type, product, component) {
       }
 
       // get connection string for send access key
-      env.QUEUE_ENVELOPE_SEND = az "servicebus queue authorization-rule keys list --resource-group ${env.RESOURCE_GROUP} --namespace-name ${serviceBusNamespace} --queue-name ${queueName} --name SendSharedAccessKey --query primaryConnectionString"
-      echo "QES: ${env.QUEUE_ENVELOPE_SEND}"
+      env.QUEUE_SEND_CONN_STRING = az "servicebus queue authorization-rule keys list --resource-group ${env.RESOURCE_GROUP} --namespace-name ${serviceBusNamespace} --queue-name ${queueName} --name SendSharedAccessKey --query primaryConnectionString"
+      echo "QES: ${env.QUEUE_SEND_CONN_STRING}"
 
       echo sh(returnStdout: true, script: 'env')
     }
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -91,9 +91,8 @@ withPipeline(type, product, component) {
       //env.TEST_STORAGE_ACCOUNT_NAME = echo sh(returnStdout: true, script: "kubectl get secret  -n ${aksServiceName} ${storageSecret} -o=jsonpath='{.data.storageAccountName}'")
       //env.TEST_STORAGE_ACCOUNT_KEY = echo sh(returnStdout: true, script: "kubectl get secret  -n ${aksServiceName} ${storageSecret} -o=jsonpath='{.data.accessKey}'")
 
-      def serviceBusConnectionStrBase64 = kubectl.getSecret(serviceBusSecret, namespace, "{.data.connectionString}")
-      echo "SBC: ${serviceBusConnectionStrBase64}"
-      def serviceBusConnectionStr = new String(serviceBusConnectionStrBase64.decodeBase64())
+      def serviceBusConnectionStr = kubectl.getSecret(serviceBusSecret, namespace, "{.data.connectionString}")
+      echo "SBC: ${serviceBusConnectionStr}"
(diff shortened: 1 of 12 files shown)


kubernetes/deployment.template.yaml
```diff
@@ -0,0 +1,186 @@
...
+spec:
+  #
+  # Get the value of this from 'svcat get classes'
+  #
+  clusterServiceClassExternalName: azure-servicebus
+
+  #
+  # The 'SKU'. Get the value of this from 'svc describe class <class name>'
+  #
+  clusterServicePlanExternalName: basic
```
timja: should be able to use standard here I think

kubernetes/deployment.template.yaml
```diff
@@ -0,0 +1,186 @@
...
+kind: ConfigMap
+metadata:
+  name: shared-config
+  namespace: cnp
+data:
+  POSTGRES_USER: bspuser
+  POSTGRES_PASSWORD: bsppassword
+---
+#
+# Rhubarb Config
```
timja: not rhubarb ^.^

kubernetes/deployment.template.yaml
luigibk: Isn't that more expensive? I used basic for my test queues previously and that worked

kubernetes/deployment.template.yaml
luigibk: Are you splitting hair? :-)

kubernetes/deployment.template.yaml
timja: sure, if it works then all is good

src/smokeTest/java/uk/gov/hmcts/reform/bulkscanprocessor/WhenApplicationDeployedTest.java
```diff
@@ -18,7 +18,7 @@
 @TestPropertySource("classpath:application.yaml")
 public class WhenApplicationDeployedTest {
 
-    @Value("${test-url}")
+    @Value("http://${test-host}")
```
timja: an env var with:
TEST_URL should be exported to you and it will have https://full-url here
no need to change this

src/smokeTest/resources/application.yaml
```diff
@@ -1,4 +1,5 @@
 test-url: ${TEST_URL:http://localhost:8581}
+test-host: ${SERVICE_FQDN}
```
timja: Is this available? why add this?

Jenkinsfile_CNP
```diff
@@ -39,16 +44,89 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
...
+env.TEST_SCAN_DELAY = '4000'
+env.RESOURCE_GROUP = 'bulk-scan-aks-rg'
+
 GradleBuilder builder = new GradleBuilder(this, product)
 
 withPipeline(type, product, component) {
+  //after('checkout') {
+  //  echo sh(returnStdout: true, script: 'env')
+  //}
+  after('smoketest:aks') {
```
timja: after smoketest:aks seems too late shouldn't it be after deploy?

luigibk: Our smoke tests don't use that stuff, so that should work. It can be moved as well, it's not a big deal.

timja: to avoid surprises can we move it please?, i.e. our smoketests should be doing more
unless there's a reason not to?

Jenkinsfile_CNP
```diff
@@ -3,6 +3,10 @@
 @Library("Infrastructure")
 import uk.gov.hmcts.contino.GradleBuilder
 
+import uk.gov.hmcts.contino.Kubectl
+
```
timja: may as well remove all the lines between imports?

Jenkinsfile_CNP
```diff
@@ -39,16 +43,63 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
   ]
 }
 
+// Vars needed for AKS testing
+env.TEST_S2S_NAME = 'bulk_scan_processor_tests'
+env.TEST_S2S_URL = 'http://rpe-service-auth-provider-aat.service.core-compute-aat.internal'
+env.TEST_SCAN_DELAY = '4000'
+env.RESOURCE_GROUP = 'bulk-scan-aks-rg'
```
timja: remove rg from the name please, https://tools.hmcts.net/confluence/display/CNP/Naming+Conventions

Jenkinsfile_CNP
```diff
@@ -39,16 +43,63 @@ static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
...
+
 GradleBuilder builder = new GradleBuilder(this, product)
 
 withPipeline(type, product, component) {
+  before('smoketest:aks') {  // or should this be after('aksdeploy')?
```
timja: remove comment its fine where it is

PR #208, opened 2018-10-10
[PREVIEW] Non repudiation

luigibk:

### Change description ###
Signature verification implementation only. RuntimeException to be replaced + signature events recorded in the next PR as this is already getting way too big. 

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -44,6 +44,8 @@ storage:
   account_name: ${STORAGE_ACCOUNT_NAME:bulkscanning}
   key: ${STORAGE_KEY:dGVzdA==} #Base 64 encoded
   blob_lease_timeout: ${STORAGE_BLOB_LEASE_TIMEOUT:15}
+  signature_algorithm: ${STORAGE_BLOB_SIGNATURE_ALGORITHM:none}  # none or sha256withrsa
+  public_key_base64: ${STORAGE_BLOB_PUBLIC_KEY:dGVzdA==} #Base 64 encoded
```
timja: that public key is really short, how was it generated?

luigibk: It's not a public key. At this stage verification is disabled. For a real public key see the tests

timja: aah I see its `test` any reason its not empty?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -0,0 +1,125 @@
...
+    public static final String DOCUMENTS_ZIP = "documents.zip";
+    public static final String SIGNATURE_SIG = "signature.sig";
+
+    private ZipVerifiers() {
+    }
+
+    public static Function<ZipStreamWithSignature, ZipInputStream> getPreprocessor(
+        String signatureAlgorithm
+    ) {
+        if ("sha256withrsa".equalsIgnoreCase(signatureAlgorithm)) {
```
nitinprabhu: I think this implementation is not correct. Either you will have digital sig verification enabled or disabled. You won't be supporting multiple algorithms.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public static boolean verifySignature(String publicKeyBase64, byte[] data, byte[] signed) {
+        PublicKey publicKey = decodePublicKey(publicKeyBase64);
+        try {
+            Signature signature = Signature.getInstance("SHA256withRSA");
+            signature.initVerify(publicKey);
+            signature.update(data);
+            return signature.verify(signed);
+        } catch (NoSuchAlgorithmException | InvalidKeyException | SignatureException e) {
+            throw new RuntimeException(e);
```
nitinprabhu: Better to throw as custom user defined exception here which will include zip file name which failed verification.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private static PublicKey decodePublicKey(String publicKeyBase64) {
+        byte[] keyBytes;
+        try {
+            keyBytes = Base64.getDecoder().decode(publicKeyBase64);
+            X509EncodedKeySpec spec = new X509EncodedKeySpec(keyBytes);
+            KeyFactory kf = KeyFactory.getInstance("RSA");
+            return kf.generatePublic(spec);
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+            throw new RuntimeException(e);
```
nitinprabhu: same as above

src/main/resources/application.yaml
```diff
@@ -44,6 +44,8 @@ storage:
   account_name: ${STORAGE_ACCOUNT_NAME:bulkscanning}
   key: ${STORAGE_KEY:dGVzdA==} #Base 64 encoded
   blob_lease_timeout: ${STORAGE_BLOB_LEASE_TIMEOUT:15}
+  signature_algorithm: ${STORAGE_BLOB_SIGNATURE_ALGORITHM:none}  # none or sha256withrsa
```
nitinprabhu: flag should be to unable disable algorithm and not to give a choice for algorithm

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
```diff
@@ -39,7 +40,11 @@ public ZipFileProcessor(Envelope envelope) {
         this.envelope = envelope;
     }
 
-    public void process(ZipInputStream zis) throws IOException {
+    public void process(
```
nitinprabhu: Process method should ideally handle the extraction and if verification enabled call the zip verifier util class. 
Code would be much easier to read that way imo.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private String publicKeyBase64 =
+        "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDEXfjyDFFigzsmFvTe2cWZ45ggH/XoS/3C6Ur/"
+            + "V0egi8k5hnIIgPEOUqhrX5UcQorSX7bIlMped6TtPkYdGs/QI6S5m2uz+6Mjai7ZfACGhYxIs8"
+            + "35msqvRsDM0tIle/h3eZJb7iPE0anMWb8MkBYU3D3vAnPdBZxiEIwNMUNzqQIDAQAB";
+    
+    @Test
+    public void should_verify_signed_file_successfully() throws Exception {
+        byte[] test1PdfBytes = toByteArray(getResource("test1.pdf"));
+        byte[] test1SigPdfBytes = toByteArray(getResource("signature/test1.pdf.sig"));
```
nitinprabhu: This is incorrect. Signature is generated for zip file and not individual pdf's.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -0,0 +1,84 @@
...
+        ZipVerifiers.verifySignature(publicKeyBase64, test1PdfBytes, test1SigPdfBytes);
+    }
+
+    @Test
+    public void should_not_verify_other_file_successfully() throws Exception {
+        byte[] test2PdfBytes = toByteArray(getResource("test2.pdf"));
+        byte[] test1SigPdfBytes = toByteArray(getResource("signature/test1.pdf.sig"));
```
nitinprabhu: same as above.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void should_verify_2_valid_filenames_successfully() throws Exception {
+        Map<String, byte[]> files = new HashMap<>();
+        files.put("documents.zip", new byte[0]);
+        files.put("signature.sig", new byte[0]);
+        assertThat(ZipVerifiers.verifyFileNames(files)).isTrue();
+    }
+
+    @Test
+    public void should_not_verify_more_than_2_files_successfully() throws Exception {
```
nitinprabhu: it should be something more specific. something like `should_throw_file_content_verification_failed_when_zip_contains_multiple_sig_files`

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    public void should_not_verify_other_file_successfully() throws Exception {
+        byte[] test2PdfBytes = toByteArray(getResource("test2.pdf"));
+        byte[] test1SigPdfBytes = toByteArray(getResource("signature/test1.pdf.sig"));
+        ZipVerifiers.verifySignature(publicKeyBase64, test2PdfBytes, test1SigPdfBytes);
+    }
+
+    @Test
+    public void should_verify_2_valid_filenames_successfully() throws Exception {
```
nitinprabhu: please check comment in below method.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
luigibk: Please do read the comments on the PR and the ticket before commenting.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
luigibk: Yes, same as above.

src/main/resources/application.yaml
luigibk: Please do read the specs before commenting.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: That is just a signature test to verify the algorithm. And it is correct in its context (unit test).

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: Indeed same as above.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: I think your version is overly long and not more meaningful, so on this we'll have to disagree ;-)

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: Again same as above.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
nitinprabhu: well this gives the impression that you are generating signatures for pdf which is incorrect

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
nitinprabhu: same as above

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
nitinprabhu: `should_not_verify_more_than_2_files_successfully` doesn't gives correct message what 2 files you are talking about. are u saying 2 pdf's or 2 signature ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
luigibk: I could not disagree more :-) as you'd be creating a dependency between the extraction and all the extra activities which should just be pluggable.

src/main/resources/application.yaml
nitinprabhu: I don't agree with this one. Please can you confirm with Neil

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
luigibk: There is no extra code in implementing it this way and it lets you change algorithm in future if needed. I do not see why it's not correct. Which part of the spec does it violate?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
nitinprabhu: your verifier class is currently extracting the zip file. at least that logic should be in some separate class.

luigibk:
@timja Yes, docs will be added in the PR related to configuring this (it's currently disabled).

src/main/resources/application.yaml
luigibk: It's just to remind people that it needs to be base64 encoded. If you use that key you'll have failures anyway so it is as good as an empty one

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
nitinprabhu: whole thing is around enabling/disabling and not around algorithm choice. and definitely you should not code for future requirements as it might change.

luigibk: This allows enabling/disbaling + flexibility at no extra cost. You have a different opinion which I simply disagree with ;-)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
luigibk: The signature and zipped documents are in a zip. To verify the signature you need to extract that. If you put that logic in a separate class you have split the verifier and made it dependent on the zip documents processor. I really would not do that.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: That is not what I am testing. Sorry if it gives you the wrong impression. The test is useful as it is.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipFileProcessor.java
nitinprabhu: Verifier is a util class so why it would be a dependent ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: Any number of files in excess of 2 is a spec violation.

luigibk:
@nitinprabhu I am not sure what changes you are requesting. Can you please state clearly what they are?

luigibk:
@nitin Before commenting you should read the PR and the ticket descriptions. If you want an open discussion, please try to understand what a PR is about and then make comments.

luigibk:
@nitinprabhu I have split it correctly. As I said already, we'll have to agree to disagree on this.

PR #210, opened 2018-10-12
[PREVIEW] Non repudiation events

luigibk:

### Change description ###
Tests fixes + refactor processor instantiation for testing

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -82,4 +88,32 @@ public void should_not_verify_invalid_zip_successfully() throws Exception {
         ZipVerifiers.sha256WithRsaVerification(zipStreamWithSig);
     }
 
+    @Test
+    public void should_verify_valid_exela_test_zip_successfully() throws Exception {
```
rkondratowicz: is `should_verify_signed_file_successfully` test still needed?
This one seems a more 'real life' test, where zip files are signed (as opposed to single pdfs)

rkondratowicz:
Not sure if we should include third party company name in tests

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: Every kind of file should be signable and working. Using 2 different signatures (1 generated by us and 1 by the other party) gives more guarantees.

luigibk:
> Not sure if we should include third party company name in tests

We probably shouldn't :-)

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -82,4 +88,32 @@ public void should_not_verify_invalid_zip_successfully() throws Exception {
...
+    public void should_not_verify_invalid_xyz_signature_successfully() throws Exception {
+        byte[] test1XyzBytes = toByteArray(getResource("signature/xyz_test_envelope.zip"));
+        byte[] test1SigXyzBytes = toByteArray(getResource("signature/signature"));
+        assertThat(ZipVerifiers.verifySignature(xyzPublicKeyBase64, test1XyzBytes, test1SigXyzBytes)).isFalse();
+    }
+
+    @Test
+    public void should_not_verify_valid_xyz_test_zip_with_wrong_public_key_successfully() throws Exception {
+        byte[] test1XyzBytes = toByteArray(getResource("signature/xyz_test_envelope.zip"));
+        byte[] test1SigXyzBytes = toByteArray(getResource("signature/xyz_test_signature"));
```
rkondratowicz: I would think based on file names that this is the correct signature for the file above

luigibk: The signature is in fact correct, but the public key is not as it says on the "tin"

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -22,42 +22,48 @@
...
         assertThat(ZipVerifiers.verifyFileNames(files)).isFalse();
     }
 
     @Test
     public void should_not_verify_invalid_filenames_successfully() throws Exception {
         Map<String, byte[]> files = new HashMap<>();
-        files.put("documents.zip", new byte[0]);
-        files.put("signature2.sig", new byte[0]);
+        files.put("envelope.zip", new byte[0]);
+        files.put("signature.sig", new byte[0]);
```
rkondratowicz: took me a while to realise what was wrong with those file names. Perhaps we could use the constants defined in zip file validator in tests and for an invalid file do something like `"NOT" + SIGNATURE_FILE_NAME`?

rkondratowicz:
Quite a lot of different changes in this PR (not of them seem to be related to 'Non repudiation events'?), please make sure to set an appropriate git comment when merging.

luigibk:
> Quite a lot of different changes in this PR (not of them seem to be related to 'Non repudiation events'?), please make sure to set an appropriate git comment when merging.

That is true. The original idea was to deal with events, but it turned out that before doing that I needed to refactor a few things and add some extra tests. I'll try to clarify that in the description

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
luigibk: That is a good idea, I'll add it to the next PR

PR #211, opened 2018-10-12
Non-repudiation errors

luigibk:

### Change description ###
Non-repudiation error handling initial implementation

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -48,13 +53,24 @@ static ZipInputStream noOpVerification(ZipStreamWithSignature zipWithSignature)
...
         Map<String, byte[]> zipEntries = extractZipEntries(zipWithSignature.zipInputStream);
         if (!verifyFileNames(zipEntries)) {
-            throw new SignatureValidationException(
-                "Zip entries do not match expected file names. "
-                + "Actual names = " + zipEntries.keySet()
+            log.warn("Signature Failure. Zip entries do not match expected file names. "
+                + "Container = {} - File = {} - Actual names = {}",
+                zipWithSignature.container, zipWithSignature.zipFileName, zipEntries.keySet()
+            );
+            throw new DocSignatureFailureException(
```
timja: should we be doing something with these files?
won't be keep trying to process them?

luigibk: That is not included in this story, but we should definitely get all the "garbage" out of the way :-)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -51,17 +51,36 @@
...
+
+    @Test
+    public void should_read_blob_check_signature_and_save_metadata_in_database_when_zip_contains_metadata_and_pdfs()
+        throws Exception {
+        //Given
+        processor.signatureAlg = "sha256withrsa";
+        processor.publicKeyBase64 =
+            "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDEXfjyDFFigzsmFvTe2cWZ45ggH/XoS/3C6Ur/"
+                + "V0egi8k5hnIIgPEOUqhrX5UcQorSX7bIlMped6TtPkYdGs/QI6S5m2uz+6Mjai7ZfACGhYxIs8"
+                + "35msqvRsDM0tIle/h3eZJb7iPE0anMWb8MkBYU3D3vAnPdBZxiEIwNMUNzqQIDAQAB";
```
rkondratowicz: Looks like a magic string... I assume this is the public key corresponding to the private key with which `42_24-06-2018-00-00-00.test.zip` has been signed.

Can this be made clearer?
Perhaps both the zip file and public key could be put in the same folder in resources?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -51,17 +51,36 @@
...
         super.setUp(BlobProcessorTask::new);
     }
 
     @Test
-    public void should_read_from_blob_storage_and_save_metadata_in_database_when_zip_contains_metadata_and_pdfs()
+    public void should_read_blob_and_save_metadata_in_database_when_zip_contains_metadata_and_pdfs()
         throws Exception {
         //Given
         uploadZipToBlobStore(VALID_ZIP_FILE_WITH_CASE_NUMBER); //Zip file with metadata and pdfs
+        testBlobFileProcessed();
```
rkondratowicz: does `VALID_ZIP_FILE_WITH_CASE_NUMBER` contain signature? If not, do we support both signed an unsigned packages?

luigibk: Currently we have to support both (see 
https://tools.hmcts.net/jira/browse/RPE-678).

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
```diff
@@ -192,9 +193,22 @@ public void should_record_generic_failure_when_metadata_parsing_fails() throws E
...
+        processor.publicKeyBase64 =
+            "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDEXfjyDFFigzsmFvTe2cWZ45ggH/XoS/3C6Ur/"
+            + "V0egi8k5hnIIgPEOUqhrX5UcQorSX7bIlMped6TtPkYdGs/QI6S5m2uz+6Mjai7ZfACGhYxIs8"
+            + "35msqvRsDM0tIle/h3eZJb7iPE0anMWb8MkBYU3D3vAnPdBZxiEIwNMUNzqQIDAQAB";
+        checkFailureEvent("43_24-06-2018-00-00-00.test.zip", Event.DOC_SIGNATURE_FAILURE);
+    }
+
+    public void checkFailureEvent(String invalidZipFile, Event event) throws Exception {
         // given
-        String noPdfZip = "5_24-06-2018-00-00-00.zip";
```
rkondratowicz: At least we knew what the problem was originally.
A comment on line 196 could help.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
```diff
@@ -80,11 +80,12 @@ public void should_throw_exception_when_zip_file_has_mismatching_pdf() throws IO
     }
 
     private ZipFileProcessor getZipFileProcessor(String zipFileName) throws IOException, URISyntaxException {
-        ZipFileProcessor processor = new ZipFileProcessor("container", zipFileName);
+        String container = "container";
+        ZipFileProcessor processor = new ZipFileProcessor(container, zipFileName);
```
rkondratowicz: can `container` be inlined?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
rkondratowicz: Is this the correct exception? Was signature even checked?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -69,21 +69,23 @@ public void should_not_verify_invalid_filenames_successfully() throws Exception
 
     @Test
     public void should_verify_valid_zip_successfully() throws Exception {
-        byte[] test1Bytes = toByteArray(getResource("signature/1_24-06-2018-00-00-00.test.zip"));
+        String zipFileName = "1_24-06-2018-00-00-00.test.zip";
+        String container = "signature";
+        byte[] test1Bytes = toByteArray(getResource(container + "/" + zipFileName));
```
rkondratowicz: wouldn't it be clearer if `signature` remained to be a subfolder in resources for tests and something like `some_container` was used as a container parameter?

luigibk: No it wouldn't. That's why I did not do it that way :-)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
luigibk: It can but I don't like it, because it's used more than once in that method, so I'd rather declare it separately.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
luigibk: The method name is quite clear on what the problem is there. I am not sure I understand this comment. What comment would you add?

luigibk:
@rkondratowicz I am not sure why you think this PR should not be merged. Can you please it make it clear? Maybe you could mark the comments you made that you think are blockers?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
rkondratowicz: oops, didn't notice it's used twice

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
rkondratowicz: Also, looks like this string is used twice in this class.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
luigibk: That's part of the specification for signature verification. The file certainly cannot be verified so it's clearer to use the same exception.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
luigibk: I've moved the public key to a file

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -118,10 +134,19 @@ private static PublicKey decodePublicKey(String publicKeyBase64) {
     public static class ZipStreamWithSignature {
         public final ZipInputStream zipInputStream;
         public final String publicKeyBase64;
-
-        public ZipStreamWithSignature(ZipInputStream zipInputStream, String publicKeyBase64) {
+        public final String zipFileName;
```
KevinJCross: public ?

PR #218, opened 2018-10-17
Remove zip files from tests: part 1

rkondratowicz:
Background:

Instead of storing **zip files** in test resources (which can't be easily inspected or, perhaps more importantly, updated), store the contents of the zip files in **folders** with descriptive name.

This PR only adds a helper class to zip folder contents in memory and updates single test.

To do in next PR(s):
- update all tests to use directories instead of zip files.
- once it's clear what zip file contents are used in tests, check if we really need 12 different archives/directories in tests


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/helpers/DirZipper.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Stream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import static com.google.common.io.Resources.getResource;
+import static java.util.stream.Collectors.toList;
+
+public class DirZipper {
```
nitinprabhu: I would be happy if you could please rename this as `DirectoryZipper`

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/helpers/DirZipper.java
```diff
@@ -0,0 +1,43 @@
...
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        try (ZipOutputStream zos = new ZipOutputStream(outputStream)) {
+
+            // .listFiles() does not guarantee any order.
+            // Files need to be sorted as currently mocks rely on files order.
+            // TODO: improve mocks in tests so that order does not matter
+            List<File> filesToZip =
+                Stream.of(new File(getResource(dirName).getPath()).listFiles())
+                    .sorted()
+                    .collect(toList());
```
timja: could you just do a forEach here or is the order not guaranteed?

rkondratowicz: can't zip operations throw a checked exception 😢

timja: try catch / wrapper function?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/helpers/DirZipper.java
rkondratowicz: done

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/helpers/DirZipper.java
rkondratowicz: gets too bloated imho

PR #219, opened 2018-10-17
Remove zip files from tests (part 2)

rkondratowicz:
Updated a couple of tests.
More to come in future PRs...


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -139,26 +139,25 @@ private void testBlobFileProcessed() throws Exception {
     @Test
     public void should_process_other_zip_files_if_previous_zip_fails_to_process() throws Exception {
         // given
-        uploadZipToBlobStore("3_24-06-2018-00-00-00.zip"); //Zip with only pdf without metadata
-        uploadZipToBlobStore("7_24-06-2018-00-00-00.zip"); //Zip with pdf and metadata
+        uploadToBlobStorage("bad_24-06-2018-00-00-00.zip", zipDir("zipcontents/missing_metadata"));
```
nitinprabhu: do we need to append `bad_` or `good_ ` as the file name format would change and not look like a real one ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -139,26 +139,25 @@ private void testBlobFileProcessed() throws Exception {
...
         processor.processBlobs();
 
         // then
-        // We expect only one envelope 4_24-06-2018-00-00-00.zip which was uploaded
+        // We expect only one envelope from the valid zip file which was uploaded
         Envelope actualEnvelope = getSingleEnvelopeFromDb();
 
         String originalMetaFile = Resources.toString(
-            getResource("metadata_4_24-06-2018-00-00-00.json"),
+            getResource("zipcontents/ok/metadata.json"),
```
rkondratowicz: wrong metadata file was used here (`4...` instead of `7...`, which was uploaded), but those happen to be the same, so test passed...

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
rkondratowicz: Wanted to use something meaningful.
Does it matter if it's real if the application accepts it?

nitinprabhu: As of now it won't matter as we don't have any file format validation in place but if we have it in future then this will break

PR #221, opened 2018-10-17
Fix validation of pdf declaration in metadata

rkondratowicz:
Fixes:
- make method static
- exception message stated that there were missing pdf when there were too many of them
- fix sample zip file used in tests


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
```diff
@@ -25,21 +25,12 @@
...
-    };
-
-
     @Test
     public void should_throw_exception_when_zip_file_contains_fewer_pdfs() throws Exception {
         ZipFileProcessor zipFileProcessor = getZipFileProcessor("zipcontents/fewer_pdfs_than_declared");
 
         Throwable throwable = catchThrowable(() ->
-            envelopeProcessor.assertEnvelopeHasPdfs(
+            EnvelopeProcessor.assertEnvelopeHasPdfs(
```
KevinJCross: you could static import it to make it read better :D

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -109,6 +109,7 @@ public Envelope getEnvelopeByFileAndContainer(String container, String zipFileNa
         .orElse(null);
     }
 
+    // TODO: move to separate class.
```
KevinJCross: hmm ... I'm not sure what the policy with Todo's are.
In previous projects to avoid Tech dept going unnoticed we created a jira ticket involved in the todo ..
i.e. //TODO: RPE-XXX move to a separate class

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -117,26 +118,33 @@ public Envelope getEnvelopeByFileAndContainer(String container, String zipFileNa
      * @param envelope to assert against
      * @param pdfs     to assert against
      */
-    public void assertEnvelopeHasPdfs(Envelope envelope, List<Pdf> pdfs) {
+    public static void assertEnvelopeHasPdfs(Envelope envelope, List<Pdf> pdfs) {
```
KevinJCross: hmmm ... 
Is it really an assertion or more a validation ?
Surely this could be put on the Envelope instead....
Seems the encapsulation is a bit violated here.
maybe something like 
val missingFiles = envelope.getMissingPdfs( pdfs)

PR #223, opened 2018-10-18
[PREVIEW] Check if blob already leased before trying to acquire lease

luigibk:
https://tools.hmcts.net/jira/browse/RPE-770

### Change description ###
Check if blob already leased before trying to acquire lease. Randomise processing order as to create less contention.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
@@ -65,14 +65,13 @@ public class BlobProcessorTaskTestWithAcquireLease extends ProcessorTestSuite<Bl
         });
 
         // Then
-        // Only one thread should be able to acquire lease and others should fail
 
-        // We expect only one envelope which was uploaded and other failed
+        // Only one thread should be able to acquire lease and others should fail.
+        // Therefore we expect only one envelope which was uploaded and others failed
         List<Envelope> envelopes = envelopeRepository.findAll();
         assertThat(envelopes.size()).isEqualTo(1);
 
         // and
-        // We expect only 3 events: doc upload, doc processed and notification sent
         List<ProcessEvent> processEvents = processEventRepository.findAll();
         assertThat(processEvents)
             .extracting("event")
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
@@ -3,7 +3,6 @@ package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.netflix.servo.util.ThreadFactories;
-import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -66,20 +66,19 @@ public void should_process_zip_file_only_once_by_processor_thread_which_acquires
 
         // Then
         // Only one thread should be able to acquire lease and others should fail
```
rkondratowicz: looks like this comment was for the code that is now removed

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestWithAcquireLease.java
```diff
@@ -66,20 +66,19 @@ public void should_process_zip_file_only_once_by_processor_thread_which_acquires
...
-                "Lease already acquired for container test and zip file 10_24-06-2018-00-00-00.zip")
-        ).isEqualTo(4);
 
         // We expect only one envelope which was uploaded and other failed
         List<Envelope> envelopes = envelopeRepository.findAll();
         assertThat(envelopes.size()).isEqualTo(1);
 
         // and
-        // We expect only two events one for doc upload and another for doc processed
+        // We expect only 3 events: doc upload, doc processed and notification sent
```
rkondratowicz: imho comment no longer needed now when the test explicitly checks events (as opposed to just the number of them previously)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -96,9 +99,13 @@ private void processZipFiles(CloudBlobContainer container)
...
         ServiceBusHelper serviceBusHelper = serviceBusHelper();
 
-        for (ListBlobItem blobItem : container.listBlobs()) {
-            String zipFilename = FilenameUtils.getName(blobItem.getUri().toString());
-
+        // Randomise iteration order to minimise lease acquire contention
+        List<String> zipFilenames = new ArrayList<>();
+        container.listBlobs().forEach(
+            b -> zipFilenames.add(FilenameUtils.getName(b.getUri().toString()))
+        );
```
rkondratowicz: perhaps you could just map the blobs?
```
List<String> zipFilenames =
  container
    .listBlobs()
    .stream()
    .map(b -> FilenameUtils.getName(b.getUri().toString()))
    .collect(toList())
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -96,9 +99,13 @@ private void processZipFiles(CloudBlobContainer container)
...
+        Collections.shuffle(zipFilenames);
+        for (String zipFilename: zipFilenames) {
```
doncem: Apparently there's a more efficient way of randomly accessing list elements. Which is more/less extraction of `Collections.shuffle`:

```java
Random random = new Random();
int listSize = list.size();

for (int i = 0; i < listSize; i++) {
    int index = rand.nextInt(listSize));
    String element = list.get(index);
    // do stuff
}
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
luigibk: There are no guarantees this way on what list can be returned and I want an implementation of the RandomAccess interface as it makes shuffling more efficient. Passing a Supplier to collect() would work, but I decided to go "traditional".

rkondratowicz: may be worth adding a comment on this then before someone who didn't see this PR comes up with an idea of refactoring it in a couple of weeks ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
luigibk: Yes it seems right as I suppose this would avoid the penalty of rewriting the list (which is what shuffle does) before iterating. I am not sure in our case this would bring about much of a gain though.

PR #231, opened 2018-10-19
[PREVIEW] Cache base64 encoded public key

luigibk:

### Change description ###
Cache public key

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -196,10 +209,15 @@ public static ZipStreamWithSignature fromKeyfile(
...
-                        );
+                    if (cachedPublicKeyFile != null && publicKeyDerFile.equals(cachedPublicKeyFile.derFilename)) {
+                        publicKeyBase64 = cachedPublicKeyFile.publicKeyBase64;
+                    } else {
+                        publicKeyBase64 =
+                            Base64.getEncoder().encodeToString(
+                                Resources.toByteArray(getResource(publicKeyDerFile))
+                            );
+                        cachedPublicKeyFile = new PublicKeyFile(publicKeyDerFile, publicKeyBase64);
+                    }
```
rkondratowicz: IMHO it would be much better to avoid mutable static variable and keep this function [pure](https://en.wikipedia.org/wiki/Pure_function).
Perhaps it could take the actual key as parameter (not the file path) and caching could take place somewhere 'higher'. `Processor` has `publicKeyDerFilename` injected to it, so maybe there.

PR #240, opened 2018-10-24
[PREVIEW] Remove zip files from tests (part 11)

rkondratowicz:
Just some cleanup:
- remove zip file that was no longer used anyway
- remove unused helper method
- fix sample zip file name


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/GetSasTokenTest.java
```diff
@@ -37,7 +37,7 @@
 
     private String destZipFilename;
 
-    private static final String zipFilename = "2_24-06-2018-00-00-00.test.zip";
+    private static final String zipFilename = "24-06-2018-00-00-00.test.zip";
```
rkondratowicz: This is used as a base for generating unique zip file name.
The helper method adds a prefix.

PR #269, opened 2018-11-01
Fix env overriding issues for non-preview

luigibk:
https://tools.hmcts.net/jira/browse/RPE-694

### Change description ###
Master is broken as the aks preview setup overrode some environment variables.

```
[ ] Yes
[x] No
```


timja:
why is this preview? we shouldn't be using preview anymore

luigibk:
@timja True! :-)

PR #272, opened 2018-11-02
Add extra wait in test

rkondratowicz:
Sending to the queue now happens in a separate task, therefore file may be deleted from blob storage, but the envelope may not yet be on the queue....


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/UpdateStatusTest.java
```diff
@@ -76,16 +75,20 @@ public void should_allow_update_envelope_status() throws Exception {
 
         String s2sToken = testHelper.s2sSignIn(this.s2sName, this.s2sSecret, this.s2sUrl);
 
+        await("processing should end")
+            .atMost(scanDelay, TimeUnit.MILLISECONDS)
```
timja: is 4 seconds enough?
the other one does scanDelay + 40_000

PR #276, opened 2018-11-05
Feature/simplify error handling

lgonczar:
### Change description ###

Simplify exception handling. The reason behind this change is that current solution has become quite complicated, logic become dispersed and it's hard to track what's going on. We also started creating exception classes just to fit into this framework. Some of those exceptions only served the purpose of data containers and weren't even thrown. This would get worse after the implementation of story https://tools.hmcts.net/jira/browse/RPE-679, which moves rejected files to separate containers.

```
[ ] Yes
[x] No
```


doncem:
Deeecent. Few uncovered notes (reported by codecov):
- `DocSignatureFailureException`
- `PreviouslyFailedToUploadException`
- the catch around `handleBlobDeletionError` section

Others can be considered as minors or might even be automatically covered by those ^. First two I would consider as business / project related cases to be sure it's covered appropriately.

Massive PR though and can be addressed later if currently is too much

lgonczar:
Agreed. I'll create another PR addressing those issues. This one has a potential to create lots of conflicts.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -73,7 +72,7 @@
     private ZipVerifiers() {
     }
 
-    public static Function<ZipStreamWithSignature, ZipInputStream> getPreprocessor(
+    public static ZipPreprocessor getPreprocessor(
         String signatureAlgorithm
```
luigibk: I am not convinced about this change for 2 reasons:
1) It obfuscates a well known Function interface with an unknown one, introducing an extra indirection layer which is unnecessary,  so it reduces clarity even if it reduces verbosity.
2) It's got nothing to do with exception handling :-)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuite.java
```diff
@@ -108,7 +105,8 @@ public void setUp(Construct<T> processorConstruct) throws Exception {
             cloudBlobClient,
             documentProcessor,
             envelopeProcessor,
-            errorWrapper,
+            envelopeRepository,
+            processEventRepository,
             SIGNATURE_ALGORITHM,
```
luigibk: We seem to always inject: envelopeProcessor, envelopeRepository and processEventRepository together. However we also inject the same 2 repositories in EnvelopeProcessor already. Couldn't that be simplified?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -177,15 +185,38 @@ private CloudBlockBlob acquireLease(CloudBlockBlob cloudBlockBlob, String contai
             }
             cloudBlockBlob.acquireLease(blobLeaseTimeout, null);
             return cloudBlockBlob;
-        });
+        } catch (StorageException storageException) {
+            if (storageException.getHttpStatusCode() == HttpStatus.CONFLICT.value()) {
+                log.error(
+                    "Lease already acquired for container {} and zip file {}",
```
KevinJCross: keeping the ordering of the parameters in the logging the same could help read the logs.
Ever thought of putting this into a json object for logging ?
I guess that depends on if we catch our logs in json because then it becomes very easy to search and manipulate.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -177,15 +185,38 @@ private CloudBlockBlob acquireLease(CloudBlockBlob cloudBlockBlob, String contai
...
 
     private ZipFileProcessor processZipFileContent(
         ZipInputStream zis,
         String zipFilename,
         String containerName
     ) {
-        return errorWrapper.wrapDocFailure(containerName, zipFilename, () -> {
+        ZipFileProcessor processor = null;
+
+        try {
```
KevinJCross: I personally would pull out the bit in the try into a new method so this function is really just the error handling case.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -199,8 +230,16 @@ private ZipFileProcessor processZipFileContent(
 
             zipFileProcessor.setEnvelope(envelopeProcessor.saveEnvelope(envelope));
 
-            return zipFileProcessor;
-        });
+            processor = zipFileProcessor;
+        } catch (DocSignatureFailureException ex) {
```
KevinJCross: Ok im not saying use this but ... this is where pattern matching coming up in JEP 305 would help.
https://openjdk.java.net/jeps/305
https://cr.openjdk.java.net/~briangoetz/amber/pattern-match.html

but vavr library allows something like this already ... but it's not as readable.
https://www.baeldung.com/vavr-pattern-matching

KevinJCross: OR  For now we could do something like this ... it actually looks quite readable :D.... 
You might also want to return an Optional or vavr Option since null can be problematic
```java
return Try.of( ()->some expression))
.recover( DocSignatureFailureException.class, t -> handleEventRelatedError(Event.DOC_SIGNATURE_FAILURE, containerName, zipFilename, t) )
.recover( PreviouslyFailedToUploadException.class, t ->  handleEventRelatedError(Event.DOC_UPLOAD_FAILURE, containerName, zipFilename, t))
...
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -60,31 +69,76 @@ protected void processParsedEnvelopeDocuments(
         deleteBlob(envelope, cloudBlockBlob);
     }
 
+    protected void handleEventRelatedError(Event event, String containerName, String zipFilename, Exception exception) {
```
KevinJCross: this could be broken out to an error handler.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -60,31 +69,76 @@ protected void processParsedEnvelopeDocuments(
...
+    }
+
+    protected void registerEvent(Event event, String container, String zipFileName, String reason) {
+        ProcessEvent processEvent = new ProcessEvent(
+            container,
+            zipFileName,
+            event
+        );
+
+        processEvent.setReason(reason);
```
KevinJCross: why not put the reason on the constructor ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -60,31 +69,76 @@ protected void processParsedEnvelopeDocuments(
...
+        } catch (Exception ex) {
+            incrementUploadFailureCount(envelope);
+            updateEnvelopeLastStatus(envelope, Event.DOC_UPLOAD_FAILURE);
+            handleEventRelatedError(Event.DOC_UPLOAD_FAILURE, envelope.getContainer(), envelope.getZipFileName(), ex);
+            return Boolean.FALSE;
+        }
+    }
+
+    private void incrementUploadFailureCount(Envelope envelope) {
+        envelope.setUploadFailureCount(envelope.getUploadFailureCount() + 1);
```
KevinJCross: hmmm maybe we should consider this should be in the envelope.
Id also consider putting the envelope repository reference into the envelope too.
This way you would say 

envelope.failed(status)
envelope.save()

The envelope failure could increment a metric that is injected into the envelope.
This would require an envelope factory too.

KevinJCross: Or even hide the save altogether ...

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
KevinJCross: Yup ... 
Joshua Bloch agrees with this.
When possible use the standard functional interfaces ...
"Item 3: Favor the use of Standard Functional Interfaces"
https://medium.com/@yury.nino.roa/effective-java-for-lambda-functions-method-references-and-standard-interfaces-6648552f685c

PR #290, opened 2018-11-14
Amend SAS token response

doncem:

[SAS token generated should include ? prefix](https://tools.hmcts.net/jira/browse/BPS-251)

### Change description ###

- Prepend with `?`
- as discussed, renaming some files to represent what data exactly is

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasQueryParamsGenerator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasQueryParamsGenerator.java
@@ -24,14 +24,14 @@ import static java.time.LocalDateTime.now;
 
 @EnableConfigurationProperties(AccessTokenProperties.class)
 @Service
-public class SasTokenGeneratorService {
+public class SasQueryParamsGenerator {
 
-    private static final Logger log = LoggerFactory.getLogger(SasTokenGeneratorService.class);
+    private static final Logger log = LoggerFactory.getLogger(SasQueryParamsGenerator.class);
 
     private final CloudBlobClient cloudBlobClient;
     private final AccessTokenProperties accessTokenProperties;
 
-    public SasTokenGeneratorService(
+    public SasQueryParamsGenerator(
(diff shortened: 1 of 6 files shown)


rkondratowicz:
Are you sure this is actually a part of the token? or just what happens to be used to build the url (query params) in the shell script used for manual tests?
This endpoint was used before by our partner and it worked for them.
Will this change not break it?

doncem:
No it will not break. Since palo alto or whatever it's name is the requirement of `?` is mandatory otherwise everything fails. Library used in the service generates same structure token as the one in Azure portal just without preceding `?`. Azure accepts both versions. palo - with `?`. The situation has been explained throughout number of meetings and this change is less than trivial to address it

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasQueryParamsGenerator.java
```diff
@@ -24,30 +24,31 @@
...
 
         log.info("SAS Token request received for service {} ", serviceName);
 
         try {
-            return cloudBlobClient
+            String token = cloudBlobClient
                 .getContainerReference(serviceName)
                 .generateSharedAccessSignature(createSharedAccessPolicy(serviceName), null);
 
+            return token.charAt(0) == '?' ? token : "?" + token;
```
nitinprabhu: Are we covering test cases for both the scenarios ?

doncem: all tests written around sas token generator/controller are not using any mocks and token just comes back without `?`. Don't see the benefit of adding tests introducing mocks just to cover the branch despite the fact I'm in favour of having everything covered

doncem:
Partners came back with explanation they are appending `?` before sticking in sas query params retrieved from our endpoint. The ticket becomes obsolete and therefore PR too. QA should be aware of this as well so similar _false negatives_ can be avoided 👍

PR #298, opened 2018-11-19
Update document type validation

doncem:

[Validate document type in metafile](https://tools.hmcts.net/jira/browse/BPS-267)

### Change description ###

As part of validation it is enough to just add Enum. but for the purpose of getting one step ahead - introducing schema regex too since document type does not have too many options to choose from.

Change is backwards compatible according to DB but not in terms of REST responses. Will mark PR as `Yes` - it has breaking change.

Notes:
- bonus, exclude `.terraform`
- it doesn't matter what how comes in, the only **must** requirement is rest response is now capital case aka whatever is defined in enum
- instead of testing schema validation doing it with valid testing. schema validation regex can't be simply regexed: the one does not simply regex the regex

```
[x] Yes
[ ] No
```


src/integrationTest/resources/envelope.json
```diff
@@ -20,7 +20,7 @@
           "file_name": "1111002.pdf",
           "notes": "test",
           "document_url": "http://localhost:8080/documents/0fa1ab60-f836-43aa-8c65-b07cc9bebcbe",
-          "document_type": "Other"
+          "document_type": "OTHER"
```
lgonczar: Why is this one all capital?

doncem: This one is used in test for API response. Responses are directly converted from enum using `.toString()` which is capital. Therefore note in the description

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/DocumentType.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.entity;
+
+import javax.persistence.AttributeConverter;
+
+public enum DocumentType {
+
+    CHERISHED,
+    OTHER;
+
+    public static class Converter implements AttributeConverter<DocumentType, String> {
```
lgonczar: I'd rather extract this converter out of DocumentType, so that the enum itself can stay unaware of those conversions.

src/main/resources/metafile-schema.json
```diff
@@ -114,6 +114,7 @@
           "document_type": {
             "id": "#/properties/scannable_items/items/properties/document_type",
             "type": "string",
+            "pattern": "^([Cc][Hh][Ee][Rr][Ii][Ss][Hh][Ee][Dd]|[Oo][Tt][Hh][Ee][Rr])$",
```
lgonczar: How about picking uppercase or lowercase and enforcing that one?

doncem: I think the specs agree on `first capital` only. Kept it 100% case insensitive as those words form a "short" regex

doncem: completely forgot about flags. and i was googling something about it before starting to work. will check this tomorrow

doncem: so tried many different ways and looks like json schema does not support regex flags. nor does support case insensitive enums. this is "the most elegant" way

only two words is not that bad. what we could do is completely ignore this and just trust object mapper implementation of mapping to java enum

lgonczar: I think `(?i)` works in Java. But we should only use it if the spec wants us to be case insensitive. Otherwise we shouldn't be more liberal than the spec is.

lgonczar: Aaa, fair enough. Adding the restricting to schema was a good thing, though. How about just `$(Cherished|Other)` or something?

doncem: Then it is just an enum really. And it will receive same treatment as `classification`. Ideally don't want to bound to case sensitivity especially when [this feature is enabled](https://github.com/hmcts/bulk-scan-processor/blob/demo/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidator.java#L19). Makes json schema validation irrelevant. Plus it is not that smart to allow some more complex rules

src/integrationTest/resources/envelope.json
lgonczar: I see. Is the contract agreed somehow? If not, probably doesn't matter, but if it is, we could use a custom JSON serialiser for this enum.

doncem: I think we are not using this service as REST at all. Previously there was a plan but now everything got moved to ServiceBus. So most likely all web related stuff is redundant. Perhaps can be reviewed in next meeting

PR #300, opened 2018-11-20
Include OCR data in queue messages

lgonczar:

### Change description ###

Include OCR data, if present in the input metadata, in messages for the orchestrator.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/resources/metafiles/valid/from-spec.json
+++ b/src/test/resources/metafiles/valid/from-spec.json
@@ -15,6 +15,7 @@
       "next_action": "return",
       "next_action_date": "24-02-2017 00:00:00.000010",
       "file_name": "1111001.pdf",
+      "ocr_data": "VGVzdCBvY3IgZGF0YQo=",
       "notes": "The document had ink spilled on it",
       "document_type": "Passport"
     },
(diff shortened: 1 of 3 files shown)


lgonczar:
We can't add a functional test verifying the queue message, as it's not guaranteed that the message won't be consumed by the orchestrator first. As for unit and integration tests, OCR data was partly handled (it already existed in the envelope). Other than that, this field alone didn't create any edge cases, so I only had to add an assertion to a test or value to test data.

src/test/resources/metafiles/valid/from-spec.json
```diff
@@ -15,6 +15,7 @@
       "next_action": "return",
       "next_action_date": "24-02-2017 00:00:00.000010",
       "file_name": "1111001.pdf",
+      "ocr_data": "VGVzdCBvY3IgZGF0YQo=",
```
rkondratowicz: do we know whether this will be a base64 encoded json?

doncem: yes, we [do](https://tools.hmcts.net/confluence/display/RBS/Technical+Specification+-+Exela#TechnicalSpecification-Exela-3.4Metafilefields)

lgonczar:
As discussed, the processor will parse OCR data and send it as a String->String map. Closing.

PR #304, opened 2018-11-23
Set default document type for old db data

doncem:
### Change description ###

As there are a lot of invalid data in other testing environments present (check #298), this needs to be manually updated or just returned as OTHER disregarding the value. Latter option is most sensible

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/DocumentTypeConverter.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/DocumentTypeConverter.java
@@ -1,8 +1,13 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.entity;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.persistence.AttributeConverter;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/DocumentTypeConverter.java
```diff
@@ -1,16 +1,28 @@
...
     }
 
     @Override
     public DocumentType convertToEntityAttribute(String dbData) {
-        return DocumentType.valueOf(dbData.toUpperCase());
+        try {
+            return DocumentType.valueOf(dbData.toUpperCase());
+        } catch (IllegalArgumentException exception) {
+            // for backwards compatibility due to a lot of incorrect data in db
+            log.warn("Invalid document type found in DB: '{}'", dbData, exception);
```
lgonczar: I think we should write a migration script fixing DB data, but then we should always fail when it's invalid.

rkondratowicz: Agree, no reason to keep and support this invalid data.

doncem:
Will have to be simplified version

PR #309, opened 2018-11-27
Feature: document type validation

doncem:

[Validate document type in meta data file](https://tools.hmcts.net/jira/browse/BPS-267)

### Change description ###

Re-introducing almost same changes as were in #298 . I believe every1 is in favour of having enum in place. After few takes and thoughts adding validation just to schema didn't feel right. So to "perhaps" assure the "non-breaking" change - including the migration script in the PR to patch up `document_type` historically.

```
[x] Yes
[x] No
```


src/functionalTest/resources/1111006_2.metadata.json
```diff
@@ -17,7 +17,7 @@
...
         "file_name": "1111006.pdf",
         "notes": "",
-        "document_type": "SSC1"
+        "document_type": "Cherished"
```
nitinprabhu: Unfortunately this is going to change now and we are going to get document type(Other, Cherished and SSCS1) and document subtype with same value initially. Later Exela is going to send only Other and Cherished and everything else goes into subtype. This is awaiting confirmation from Exela.

doncem: Enum can be extended without a sweat. Let's get final confirmation even though task has already been set

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeResponseMapperTest.java
```diff
@@ -15,18 +16,21 @@
...
     @Test
-    public void should_map_envelope_to_envelope_response() throws Exception {
+    public void should_map_envelope_to_envelope_response() {
         EnvelopeResponse response = EnvelopeResponseMapper.toEnvelopeResponse(envelope);
 
         assertThat(response)
             .usingComparatorForFields(
                 new ToStringComparator<Classification>(), new String[]{"classification"}
             )
+            .usingComparatorForFields(
```
lgonczar: Is this needed?

doncem: yes. response is mapped to string value whereas assertion is done against domain model which has enum there and assertion does not know how to convert to string automatically. same treatment as `classification` here

PR #312, opened 2018-11-28
Add a new JSON column for OCR data

lgonczar:

### Change description ###

Create a DB migration script adding a new column to store OCR data as JSON. This is the first step to replace the `ocrData` column (storing Base64-encoded strings) with no downtime.

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V020__Add_ocrDataJson_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scannable_items
+ADD COLUMN ocrDataJson JSON null
```
nitinprabhu: might be worth just naming the column as ocrData rather than having type in the column name

doncem: ocrData is an existing column. @lgonczar is making non-breaking migration approach by just adding a new column

PR #315, opened 2018-12-03
Split the models into two sets, separately representing DB and metadata

lgonczar:
### Change description ###

Split the models into two sets, separately representing DB and metadata. What's stored in the DB is a bit different than what's in the metadata file and that difference keeps growing. Separating the models will make things clearer and easier to maintain.

The PR is large, but the change can't really be made in pieces.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    private static NonScannableItem toDbNonScannableItem(InputNonScannableItem nonScannableItem) {
+        return new NonScannableItem(
+            nonScannableItem.documentControlNumber,
+            nonScannableItem.itemType,
+            nonScannableItem.notes
+        );
+    }
```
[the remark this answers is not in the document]
lgonczar: What kind of simplification do you have in mind?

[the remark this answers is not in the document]
doncem: each case is a conversion from one class to another. so why not
```java
NonScannableItem(InputNonScannableItem input) {
    // set all final fields in this constructor
}
```

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTest.java
```diff
@@ -23,107 +24,107 @@ public void setUp() throws IOException, ProcessingException {
...
-        Envelope envelope = getEnvelope("/metafiles/valid/cheque-payment-method.json");
+        InputEnvelope envelope = getEnvelope("/metafiles/valid/cheque-payment-method.json");
 
-        assertThat(envelope.getPayments()).hasSize(1);
-        assertThat(envelope.getPayments().get(0).getAmount()).isEqualTo(100.0);
-        assertThat(envelope.getPayments().get(0).getPaymentInstrumentNumber()).isEqualTo("1000000000");
-        assertThat(envelope.getPayments().get(0).getAccountNumber()).isEqualTo("12345678");
-        assertThat(envelope.getPayments().get(0).getSortCode()).isEqualTo("112233");
+        assertThat(envelope.payments).hasSize(1);
+        assertThat(envelope.payments.get(0).amount).isEqualTo(new BigDecimal("100.00"));
```
nitinprabhu: instead of using get(0) can we use something similar to
```
assertThat(fellowshipOfTheRing).extracting("name", "age", "race.name")
                               .contains(tuple("Boromir", 37, "Man"),
                                         tuple("Sam", 38, "Hobbit"),
                                         tuple("Legolas", 1000, "Elf"));
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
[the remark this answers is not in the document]
rkondratowicz: imho those 2 models should be independent.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTest.java
lgonczar: Can be done like this. My change only replaced the use of getters with the use of public immutable fields. There are a few places where we can make this change and given the size of this PR, it would be best to have a separate one for that.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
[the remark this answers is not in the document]
lgonczar: I see the benefit of having one argument instead of many, but that would happen at the cost of making one set of models aware of the other. Like Rafał, I think it's best to avoid it.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTest.java
nitinprabhu: 👍

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeTest.java
```diff
@@ -40,44 +36,34 @@ public void cleanUp() {
...
 
         // and
         UUID envelopeId = repository.save(envelope).getId();
 
         // when
-        Envelope dbEnvelope = repository.getOne(envelopeId);
+        Envelope readEnvelope = repository.getOne(envelopeId);
 
         // then
         try (InputStream stream = EnvelopeCreator.getMetaFile()) {
```
rkondratowicz: looks like this can be removed as well

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelperTest.java
```diff
@@ -37,12 +37,17 @@
...
+    private ObjectMapper objectMapper;
+
+    @Mock
+    private IQueueClient queueClient;
+    @Mock
+    private Envelope envelope;
+    @Mock
+    private ScannableItem scannableItem1;
+    @Mock
+    private ScannableItem scannableItem2;
```
rkondratowicz: is there something new here? it's hard to tell...

PR #319, opened 2018-12-05
 Deserialise and process OCR data as a map of strings

lgonczar:

### Change description ###

Convert OCR data from metadata file to a String -> String map and store it like this. This includes transforming OcrDataParser into a JSON deserialiser.

```
[ ] Yes
[x] No
```


src/integrationTest/resources/zipcontents/fewer_pdfs_than_declared/metadata.json
```diff
@@ -14,7 +14,7 @@
...
...50X21vYmlsZSIsICJtZXRhZGF0YV9maWVsZF92YWx1ZSI6ICIifSwKICAgICB7ICJtZXRhZGF0YV9maWVsZF9uYW1lIjogImFwcGVsbGFudF9kb2IiLCAibWV0YWRhdGFfZmllbGRfdmFsdWUiOiAiMTEvMTAvMTk4MSJ9LAogICAgIHsgIm1ldGFkYXRhX2ZpZWxkX25hbWUiOiAiYXBwZWxsYW50X25pRmFsc2UiLCAibWV0YWRhdGFfZmllbGRfdmFsdWUiOiAiNDdCQVM3NjU0In0sCiAgICAgeyAibWV0YWRhdGFfZmllbGRfbmFtZSI6ICJpc19hcHBvaW50ZWUiLCAibWV0YWRhdGFfZmllbGRfdmFsdWUiOiAiVHJ1ZSJ9LAogICAgIHsgIm1ldGFkYXRhX2ZpZWxkX25hbWUiOiAiYXBwb2ludGVlX3RpdGxlIiwgIm1ldGFkYXRhX2ZpZWxkX3ZhbHVlIjogIk1yIn0sCiAgICAgeyAibWV0YWRhdGFfZmllbGRfbmFtZSI6ICJhcHBvaW50ZWVfZmlyc3ROYW1lIiwgIm1ldGFkYXRhX2ZpZWxkX3ZhbHVlIjogIlNvcGhpYSJ9LAogICAgIHsgIm1ldGFkYXRhX2ZpZWxkX25hbWUiOiAiYXBwb2ludGVlX2xhc3ROYW1lIiwgIm1ldGFkYXRhX2ZpZWxkX3ZhbHVlIjogIlNvbGUifSwKICAgICB7ICJtZXRhZGF0YV9maWVsZF9uYW1lIjogImFwcG9pbnRlZV9hZGRyZXNzIiwgIm1ldGFkYXRhX2ZpZWxkX3ZhbHVlIjogIk4vQSJ9LAogICAgIHsgIm1ldGFkYXRhX2ZpZWxkX25hbWUiOiAiYXBwb2ludGVlX3Bvc3Rjb2RlIiwgIm1ldGFkYXRhX2ZpZWxkX3ZhbHVlIjogIkQxNzFZNyJ9LAogICAgIHsgIm1ldGFkYXRhX2ZpZWxkX25hbWUiOiAiYXBwb2ludGVlX2RvYiIsICJtZXRhZGF0YV9maWVsZF92YWx1ZSI6ICIxNC8xMC8xOTgwIn0sCiAgICAgeyAibWV0YWRhdGFfZmllbGRfbmFtZSI6ICJhcHBvaW50ZWVfbmlGYWxzZSIsICJtZXRhZGF0YV9maWVsZF92YWx1ZSI6ICI0NDE3NjU0M0IifQogIF0KfQo=",
```
lgonczar: Same data, but in a valid JSON. Previous version has some double quotes missing and an extra comma.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
```diff
@@ -1,19 +1,35 @@
...
+    public OcrDataDeserializer() {
+        super(Map.class);
+    }
+
+    @Override
+    public Map<String, String> deserialize(
+        JsonParser jsonParser,
+        DeserializationContext deserializationContext
+    ) throws IOException {
+        return parseOcrData(jsonParser.getText());
```
doncem: Pretty pointless to have a method which just calls another method and there's no other use of that child method anywhere else. How about we get rid of try/catch block in the private method and wrap here by using built-in Jackson process of deserialising incoming string? This will use the library effectively.

```java
try {
   return parseOcrData(jsonParser.getText());
} catch (<list of exceptions> exception) {
    throw deserializationContext.instantiationException(_valueClass, exception);
}
```

Not entirely sure deserialization context actually needs throw part here

lgonczar: It's actually two statements happening here, not just a method call. I'm ok with moving try-catch to this method. Will do. Not sure about instantiationException, as it seems to have a more specific purpose (problem with constructor).

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/blob/InputScannableItem.java
```diff
@@ -15,7 +17,7 @@
     public final String manualIntervention;
     public final String nextAction;
     public final Timestamp nextActionDate;
-    public final String ocrData;
+    public final Map<String, String> ocrData;
```
lgonczar: OCR data can be either deserialised straight to a map, like here, or deserialised to string and converted later. The second solution would make the process a bit trickier, so I went with this one, but maybe you've got some thought?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
doncem: Was just a suggestion to use context 😉. I think ideally we all should just try use more features from libraries we already incorporate. Of course, if it makes sense to do so. Deserialisation seemed like a good start. It is construct - it is trying to construct a map for ocr data field. In case there can be more concrete problem identified - different method used.

Like I said, was just a suggestion

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/blob/InputScannableItem.java
doncem: Straight to `Map` is so much acceptable

PR #323, opened 2018-12-10
Move OCR data to envelope level in queue message

lgonczar:

### Change description ###

Move OCR data from document to envelope in queue message. Before this change, OCR data was kept in documents, but CCD schema defines it as a top-level field and the change aligns with it.

```
[ ] Yes
[x] No
```


lgonczar:
That's what's gonna happen. OCR data should only be in one scanned doc. There's a separate story for the validation of OCR data, where we should make sure this is covered.

PR #334, opened 2018-12-12
Update public keys 

nitinprabhu:
- Updated public keys for functional tests for AAT and Demo environments.

- We would need to update unit and integration test to use 2048 bit key rather than 1024 which will be done as a separate PR.

- Also we could extract some common code like setting values from `application.conf` to a base class which also can be done as separate PR.


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/BlobProcessorTest.java
```diff
@@ -64,7 +66,9 @@ public void should_process_zipfile_after_upload_and_set_status() throws Exceptio
         String metadataFile = "1111006_2.metadata.json";
         String destZipFilename = testHelper.getRandomFilename("24-06-2018-00-00-00.test.zip");
 
-        testHelper.uploadZipFile(testContainer, files, metadataFile, destZipFilename); // valid zip file
+        // valid zip file
+        testHelper.uploadZipFile(testContainer, files, metadataFile, destZipFilename,testPrivateKeyDer);
```
rkondratowicz: missing space

PR #338, opened 2018-12-12
RPE-876: Changing to the new version of the Open service broker for azure.

KevinJCross:
### Change description ###
The new version of the open service broker supports queue creation.
This now simplifies the deployment of the service and queue creation.


rkondratowicz:
closing PR to save aks resources

PR #355, opened 2018-12-17
Feature: error notification repository

doncem:

### Change description ###

Standalone repository. Next PR - adding as dependency to a service.

Note. Linking up to event table is up for discussion or just review with explanation/comment about it - not a problem to add

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
@@ -49,8 +49,11 @@ public class ErrorNotificationRepositoryTest {
     @Test
     public void should_successfully_create_a_record() {
         // given
-        ErrorNotification errorNotification = new ErrorNotification(ERR_AV_FAILED.name(), "antivirus hit");
-        errorNotification.setEvent(em.getReference(ProcessEvent.class, eventId));
+        ErrorNotification errorNotification = new ErrorNotification(
+            em.getReference(ProcessEvent.class, eventId),
+            ERR_AV_FAILED.name(),
+            "antivirus hit"
+        );
 
         // when
         long dbId = repository.save(errorNotification).getId();
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
@@ -1,0 +1,59 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.entity;
+
+import org.junit.After;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase;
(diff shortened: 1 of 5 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
```diff
@@ -0,0 +1,72 @@
...
+        errorNotification.setEvent(em.getReference(ProcessEvent.class, eventId));
+
+        // when
+        long dbId = repository.save(errorNotification).getId();
+
+        // then
+        ErrorNotification dbItem = repository.getOne(dbId);
+
+        assertThat(dbItem.getProcessEvent().getId()).isEqualTo(eventId);
+        assertThat(dbItem.getNotificationId()).isNullOrEmpty();
```
lgonczar: It would be good to test both null and non-null values

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertThat(dbItem.getErrorCode()).isEqualTo(ERR_AV_FAILED.name());
+        assertThat(dbItem.getErrorDescription()).isEqualTo("antivirus hit");
+
+        // and
+        String iso8601DateTime = ZonedDateTime
+            .ofInstant(errorNotification.getCreatedAt().toInstant(), ZoneId.of("UTC"))
+            .format(DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"));
+        assertThat(dbItem.getCreatedAt().toInstant()).isEqualTo(iso8601DateTime);
```
lgonczar: Why not check the instants directly (maybe adjusting the time zone)?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotificationRepositoryTest.java
doncem: this test is verifying the fact saving to db works. next pr verifies notification is set when client responds with 2xx and not set otherwise

src/main/resources/db/migration/V024__Error_notifications.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE error_notifications (
```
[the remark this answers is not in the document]
doncem: I'll remove description

src/main/resources/db/migration/V024__Error_notifications.sql
```diff
@@ -0,0 +1,9 @@
...
+  id BIGSERIAL PRIMARY KEY,
+  event_id BIGINT NOT NULL REFERENCES process_events(id),
+  notificationId VARCHAR(50) NULL,
+  documentControlNumber VARCHAR(100) NULL,
```
nitinprabhu: documentControlNumber and referenceId would be blank for most of the cases as of now. It will be only be applicable to rescan which has come from ccd.

lgonczar: Can notifications concern individual documents?

doncem: Like @nitinprabhu said, back from ccd when validating number with existing ones. Perhaps something more.

Looks like this table is as minimal as possible. Events table don't concern document control numbers so as per specs I have included it here

PR #359, opened 2018-12-17
Use `Optional` for error code mapping result

rkondratowicz:
Safer than returning `null`


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -250,32 +249,33 @@ private void handleInvalidFileError(
         Exception cause
     ) {
         Long eventId = handleEventRelatedError(fileValidationFailure, containerName, zipFilename, cause);
-        ErrorCode errorCode = ErrorMapping.getFor(cause.getClass());
+        ErrorMapping
+            .getFor(cause.getClass())
+            .ifPresent(errorCode -> {
```
lgonczar: Maybe a good moment to extract the content of ifPresent to a separate method?

PR #361, opened 2018-12-17
Deserialize ocr data to retain the fields insertion order

aliveni:
https://tools.hmcts.net/jira/browse/BPS-420

### Change description ###
Updated ocr data field type to LinkedHashSet 
to keep the ocr fields order same as in the input json file.

```
[ ] Yes
[x ] No
```


rkondratowicz:
why was this needed?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataSerializationJourneyTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.io.InputStream;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+
+public class OcrDataSerializationJourneyTest {
+
+    @Test
+    public void should_deserialize_all_ocr_fields_in_insertion_order() throws Exception {
```
lgonczar: I think this test should be an integration test, as it checks the whole journey. It should also include saving and reading from the DB, as that's a very uncertain bit here.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
```diff
@@ -41,7 +42,11 @@ public OcrDataDeserializer() {
         return ocrData.getFields().stream().collect(
             toMap(
                 field -> field.getName().textValue(),
-                field -> field.getValue().asText("")
+                field -> field.getValue().asText(""),
+                (u, v) -> {
+                    throw new IllegalStateException(String.format("Duplicate key for value '%s'", u));
```
rkondratowicz: How is `IllegalStateException` handled by the main job?
Will envelope end up in an appropriate state?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
```diff
@@ -41,7 +42,11 @@ public OcrDataDeserializer() {
         return ocrData.getFields().stream().collect(
             toMap(
                 field -> field.getName().textValue(),
-                field -> field.getValue().asText("")
+                field -> field.getValue().asText(""),
+                (u, v) -> {
```
rkondratowicz: what's `u` and `v`?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializerTest.java
```diff
@@ -33,6 +33,20 @@ public void should_convert_all_fields_to_their_string_representation() throws Ex
...
+        Throwable thrown = catchThrowable(() -> {
+                JsonParser jsonParser = getJsonParser("/ocr-data/invalid/invalid-ocr.json");
+                deserializer.deserialize(jsonParser, mock(DeserializationContext.class));
+            }
+        );
+
+        assertThat(thrown).isInstanceOf(OcrDataParseException.class)
+            .hasCauseExactlyInstanceOf(IllegalStateException.class);
+
+        assertThat(thrown.getCause().getMessage()).contains("Duplicate key for value 'some text'");
```
rkondratowicz: The duplicate key is `text_field`.
That would be much more useful information

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
nitinprabhu: may be key1,key2 ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/OcrDataSerializationJourneyTest.java
```diff
@@ -0,0 +1,73 @@
...
+@RunWith(SpringRunner.class)
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@DataJpaTest
+public class OcrDataSerializationJourneyTest {
+
+    @Autowired
+    private EnvelopeRepository repository;
+
+    @Test
+    public void should_deserialize_all_ocr_fields_in_insertion_order() throws Exception {
```
lgonczar: This test doesn't test deserialisation. It only test if the ocrData map in ScannableItem converts to the right JSON in the message. The rest of the journey is irrelevant.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataSerializationJourneyTest.java
aliveni: Moved the test to integration tests and added assertions for the envelope read from db

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
rkondratowicz: unfortunately, we lose the information about the duplicate key this way... :(
Perhaps we could identify whether there are any duplicates in the list before we convert it to a map?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/OcrDataSerializationJourneyTest.java
```diff
@@ -0,0 +1,80 @@
...
+        JsonNode jsonNode = mapper.readTree(bytes);
+
+        JsonNode ocrData = jsonNode.get("ocr_data");
+        Map<String, String> actualValue = mapper.convertValue(
+            ocrData,
+            new TypeReference<LinkedHashMap<String, String>>() {
+            }
+        );
+        AssertionsForInterfaceTypes.assertThat(actualValue.entrySet())
+            .containsExactlyElementsOf(expectedResult.entrySet());
```
lgonczar: 👍

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializerTest.java
lgonczar: If those are field values then we should just exclude them from the message. They might be sensitive information and we have to be sure it's not logged anywhere, etc..

src/test/resources/ocr-data/invalid/invalid-ocr.json
```diff
@@ -0,0 +1,20 @@
+{
```
lgonczar: There's a test file named `duplicate-field-name.json` in the same directory. It's almost identical, so I think you could just use that one.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/OcrDataSerializationJourneyTest.java
```diff
@@ -0,0 +1,80 @@
...
+        InputStream resourceAsStream = EnvelopeCreator.class.getResourceAsStream(
+            "/metafiles/valid/envelope-with-ocr-data.json"
+        );
+
+        Map<String, String> expectedResult = ImmutableMap.of(
+            "text_field", "some text",
+            "number_field", "123",
+            "boolean_field", "true",
+            "null_field", ""
+        );
```
doncem: hmm. expected result should have been retrieved from `inputEnvelope` and that used in assertion

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/features/OcrDataSerializationJourneyTest.java
doncem: it is a journey of ocr data. we don't really care bits in between what matters is input and output stays the same. which got somehow missed out 🤔

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializerTest.java
aliveni: I've updated the log message to hide the values. It just says, there are duplicate fields in the ocr data

src/test/resources/ocr-data/invalid/invalid-ocr.json
aliveni: done, deleted the new file and reusing the existing one

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
aliveni: I've created another jira ticket to do this as this pr is blocking sscs UAT testing

PR #362, opened 2018-12-17
Simplify queue error model.

rkondratowicz:
Always use `false` for `isTestOnly`.
A test endpoint will be used on testing environments, so the scheduled job can always do what it's supposed to do.


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/ErrorMsg.java
```diff
@@ -48,7 +45,7 @@ public String getMsgId() {
 
     @Override
     public boolean isTestOnly() {
-        return this.testOnly;
+        return false;
```
nitinprabhu: Do we need this anymore ?

rkondratowicz: Yeah, it's in the interface that we implement

PR #366, opened 2018-12-18
Save error notification in db

doncem:

### Change description ###

Saving attempt to send error notification. This is based from #355 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotification.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotification.java
@@ -33,10 +33,7 @@ public class ErrorNotification {
         // For use by hibernate.
     }
 
-    public ErrorNotification(
-        ProcessEvent event,
-        String errorCode
-    ) {
+    public ErrorNotification(ProcessEvent event, String errorCode) {
         this.event = event;
         this.errorCode = errorCode;
     }
@@ -51,6 +48,10 @@ public class ErrorNotification {
 
     public String getNotificationId() {
         return notificationId;
+    }
+
+    public void setNotificationId(String notificationId) {
+        this.notificationId = notificationId;
     }
 
     public String getErrorCode() {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
@@ -4,9 +4,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Service;
 import uk.gov.hmcts.reform.bulkscanprocessor.client.ErrorNotificationClient;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ErrorNotification;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ErrorNotificationRepository;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ProcessEvent;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.in.ErrorNotificationResponse;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.errors.ErrorNotificationRequest;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ErrorNotification.java
```diff
@@ -73,4 +77,12 @@ public String getErrorDescription() {
     public Timestamp getCreatedAt() {
         return createdAt;
     }
+
+    public void setDocumentControlNumber(String documentControlNumber) {
+        this.documentControlNumber = documentControlNumber;
+    }
+
+    public void setReferenceId(String referenceId) {
```
rkondratowicz: let's add those new params to a constructor so that we can create the object in single step.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
```diff
@@ -30,10 +43,24 @@ public void processServiceBusMessage(ErrorMsg message) {
...
+
+        try {
+            ErrorNotificationResponse response = client.notify(request);
+
+            entity.setNotificationId(response.getNotificationId());
 
-        ErrorNotificationResponse response = client.notify(request);
+            log.info("Error notification published. ID: {}", response.getNotificationId());
+        } catch (Exception exception) {
+            log.error("Failed to publish error notification. {}", exception.getMessage(), exception);
```
lgonczar: Adding exception message to the error message is unnecessary. It will be there in the log, anyway.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
```diff
@@ -30,10 +43,24 @@ public void processServiceBusMessage(ErrorMsg message) {
...
+            entityManager.getReference(ProcessEvent.class, message.eventId),
+            message.errorCode.name(),
+            message.documentControlNumber,
+            message.id
+        );
+
+        try {
+            ErrorNotificationResponse response = client.notify(request);
+
+            entity.setNotificationId(response.getNotificationId());
```
lgonczar: How about adding notification id to the constructor? Looks like we can just create a proper instance in one go.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
```diff
@@ -30,10 +43,24 @@ public void processServiceBusMessage(ErrorMsg message) {
...
-        ErrorNotificationResponse response = client.notify(request);
+            log.info("Error notification published. ID: {}", response.getNotificationId());
+        } catch (Exception exception) {
+            log.error("Failed to publish error notification. {}", exception.getMessage(), exception);
+        }
 
-        // store result in db. wip
-        log.info("Error notification published. ID: {}", response.getNotificationId());
+        // save the entity after everything is processed
+        repository.save(entity);
```
lgonczar: Should this be inside the `try` clause?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
doncem: oh will it? i vaguely remember missing those messages. but perhaps i'm confusing with something else

lgonczar: There used to be a problem, but it's solved now.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationService.java
doncem: possibly. we don't want to dead letter the notification just because we failed to save the fact

doncem: There is more to the story. The way message needs to be handled differs from the response received from client/service. Will introduce some extra error handling in another PR. It should cover all edge cases and conclude the story. Not sure whether I will finish it but I'll liaise with you anyway

PR #370, opened 2018-12-18
Refactor queues configuration

doncem:

### Change description ###

This makes additional feature (queue push) addition smooth implementation. Furthermore, next step for the ticket will be implementing the queue *READ* client. This also helps a lot

```
[ ] Yes
[x] No
```


rkondratowicz:
Not sure I see the benefits. ~120 new lines of code added, imho seems more complicated than it was and 'fix for fix' commits prove it's not very straightforward at all.

PR #373, opened 2018-12-20
Feature: consume notifications

doncem:

### Change description ###

Processing error notifications in the message handler way provided by azure service bus library. Use the benefit of java8 feature of `CompletableFuture` and separate pool as suggested - best to not use common one

```
[ ] Yes
[x] No
```


doncem:
let me split the configuration apart so only handler is here and configuration should come in with some more stable tests (working on that now)

PR #376, opened 2018-12-20
Feature: message handler to consume error notifications

doncem:

### Change description ###

Processing error notifications in the message handler way provided by azure service bus library. Use the benefit of java8 feature of `CompletableFuture` and separate execution pool as it is recommended - best to not use common one.

This is extract of #373 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/ErrorMsg.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/ErrorMsg.java
@@ -1,5 +1,8 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /**
  * Message describing an error processing an envelope.
  */
@@ -17,14 +20,14 @@ public class ErrorMsg implements Msg {
     // region constructors
     @SuppressWarnings("squid:S00107") // number of params
     public ErrorMsg(
-        String id,
-        Long eventId,
-        String zipFileName,
-        String jurisdiction,
-        String poBox,
-        String documentControlNumber,
-        ErrorCode errorCode,
-        String errorDescription
+        @JsonProperty(value = "id", required = true) String id,
+        @JsonProperty(value = "eventId", required = true) Long eventId,
+        @JsonProperty(value = "zipFileName", required = true) String zipFileName,
+        @JsonProperty("jurisdiction") String jurisdiction,
+        @JsonProperty("poBox") String poBox,
+        @JsonProperty("documentControlNumber") String documentControlNumber,
+        @JsonProperty(value = "errorCode", required = true) ErrorCode errorCode,
+        @JsonProperty(value = "errorDescription", required = true) String errorDescription
     ) {
         this.id = id;
         this.eventId = eventId;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class ErrorNotificationHandler implements IMessageHandler {
+
+    private static final Logger log = LoggerFactory.getLogger(ErrorNotificationHandler.class);
+
+    private final ErrorNotificationService service;
+
+    private final ObjectMapper mapper;
+
+    static final Executor SIMPLE_EXEC = Runnable::run;
```
nitinprabhu: are the scope of these not private ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+    static final Executor SERVICE_EXEC = Executors.newSingleThreadExecutor(r ->
```
nitinprabhu: same here ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/ErrorMsg.java
```diff
@@ -17,14 +20,14 @@
...
     public ErrorMsg(
-        String id,
-        Long eventId,
-        String zipFileName,
-        String jurisdiction,
-        String poBox,
-        String documentControlNumber,
-        ErrorCode errorCode,
-        String errorDescription
+        @JsonProperty(value = "id", required = true) String id,
```
nitinprabhu: Can we have a test which shows the exception we receive when required fields are not passed and service is able to handle it ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
doncem: It is private but was not sure how this gets registered, iirc those statics are purposefully reused across multiple instances within same server which is one of the main points on how to use futures in java. After stating the obvious - not use executors from common pull

But I must be overthinking too much. Do you think it's worth setting them private?

nitinprabhu: yes let's keep those private to make sure we follow encapsulation

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(ErrorNotificationHandler.class);
+
+    private final ErrorNotificationService service;
+
+    private final ObjectMapper mapper;
+
+    private static final Executor SIMPLE_EXEC = Runnable::run;
+
+    private static final Executor SERVICE_EXEC = Executors.newSingleThreadExecutor(r ->
```
nitinprabhu: hope this is all efficient and we may have not threading issues. Although I read this is more efficient 😉

PR #379, opened 2018-12-27
Remove a failure-related event from a non-failure case

lgonczar:
### Change description ###

Remove an failure-related event from a non-failure case. This event is replaced with logs in https://github.com/hmcts/bulk-scan-processor/pull/378

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
@@ -127,10 +127,7 @@ public abstract class Processor {
         try {
             // Lease needs to be broken before deleting the blob. 0 implies lease is broken immediately
             cloudBlockBlob.breakLease(0);
-            boolean deleted = cloudBlockBlob.deleteIfExists();
-            if (!deleted) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -127,10 +127,7 @@ private Boolean deleteBlob(
         try {
             // Lease needs to be broken before deleting the blob. 0 implies lease is broken immediately
             cloudBlockBlob.breakLease(0);
-            boolean deleted = cloudBlockBlob.deleteIfExists();
-            if (!deleted) {
-                handleBlobDeletionError(envelope, null);
```
lgonczar: This would actually fail if it was called, because of NullPointerException. #378 logs a warning instead.

doncem: "love" how `boolean` can be `null`

lgonczar: Primitive boolean can't, but `handleBlobDeletion` passes null where exception is expected.

PR #385, opened 2019-01-02
Add some logging related with sending notifications

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-424

### Change description ###

Add some logging related with sending notifications

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/OrchestratorNotificationTask.java
```diff
@@ -46,19 +48,42 @@ public OrchestratorNotificationTask(
...
                 } catch (Exception exc) {
                     createEvent(env, Event.DOC_PROCESSED_NOTIFICATION_FAILURE);
                     // log error and try with another envelope.
-                    LOGGER.error("Error sending envelope notification", exc);
+                    log.error("Error sending envelope notification", exc);
+                    return 0;
                 }
-            });
+            })
+            .count();
```
doncem: i think `.count` will return the size of the stream. instead we should use `.mapToInt` and `.sum` directly to actually reduce properly and get the correct count of success

lgonczar: Yes, of course. Or I could filter instead of mapping. Well spotted.

PR #386, opened 2019-01-02
Feature: exception handling for error notifications

doncem:

### Change description ###

First version of exception handling. Uses built in completable future feature to custom-ly handle outcome (whether failed or not) and proceed as one wishes. We only need to make sure that in certain cases message gets re-scheduled. For simplicity, client 5xx errors are re-scheduled only as any other at the time being is more of a failure which doesn't need to be re-addressed

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
@@ -38,8 +38,11 @@ public class ErrorNotificationExceptionHandler {
 
     private IMessage handleErrorNotificationException(IMessage message, ErrorNotificationException exception) {
         if (exception.getStatus().is5xxServerError()) {
-            // instant to be decided
-            errorNotificationPushClient.scheduleMessageAsync(message, Instant.now());
+            // 10th delivery time is 6h delay
+            // Eventually it'll fail in case client is down for quite some time
+            long secondsToAdd = (long) Math.floor(Math.exp((double) message.getDeliveryCount()));
+
+            errorNotificationPushClient.scheduleMessageAsync(message, Instant.now().plusSeconds(secondsToAdd));
 
             return message;
         }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
@@ -1,0 +1,23 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptionhandlers;
+
+import com.microsoft.azure.servicebus.IMessage;
+import com.microsoft.azure.servicebus.IQueueClient;
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
```diff
@@ -59,6 +70,10 @@ private void processMessage(ErrorMsg message) {
         service.processServiceBusMessage(message);
     }
 
+    private void logCompletion(String messageId) {
+        log.debug("Error notification consumed. ID {}", messageId);
```
lgonczar: I wonder if this should not be info instead. It's useful information and if it's debug, we'll never see it, except for local env maybe.

doncem: it would be duplicating log. service already announces the completion:
```
log.info("Error notification published. ID: {}", response.getNotificationId());
```

then it tries to save the db record. if in any case it fails - atm message will be dead lettered and relevant logs will be error'ed

nitinprabhu: I think notification failures must be only dead lettered if they are bad request ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    private IMessage handleErrorNotificationException(IMessage message, ErrorNotificationException exception) {
+        if (exception.getStatus().is5xxServerError()) {
+            // 10th delivery time is 6h delay
+            // Eventually it'll fail in case client is down for quite some time
+            long secondsToAdd = (long) Math.floor(Math.exp((double) message.getDeliveryCount()));
+
+            errorNotificationPushClient.scheduleMessageAsync(message, Instant.now().plusSeconds(secondsToAdd));
```
lgonczar: While IMessageHandler doesn't offer any alternative, there are some problems with this approach, compared to complete/do nothing/dead-letter/abandon - we put a new message on the queue. What if we fail to do that? Also, would there be problems with duplicate detection? Not sure about delivery count being incremented in this case, either.

doncem: now this is best intel - thanks. we have not configured queue to be [duplicate aware](https://github.com/hmcts/bulk-scan-shared-infrastructure/blob/b3836b8d89f233d018c7c60d0439047e8bb680a8/queue.tf#L20-L25) so this concern can be relaxed.

> what if we fail to do that?

well, message gets deadlettered disregarding

one thing is not clear - delivery count increase. aah! and i think i've understood the "gotcha" here. so i don't need to re-send "same" message to queue. i could just update scheduled time and throw the original exception. then `IMessageHandler` will understand it's a failure and return to the queue until delivery count has been reached. is this correct flow?

doncem:
the solution will be revised. current implementation is lacking the desired control needed. it is possible to register different handler to avoid `write` client presence in this scope

PR #389, opened 2019-01-03
BPS-401 Added proxy properties to enable proxy for azure client in demo

aliveni:
https://tools.hmcts.net/jira/browse/BPS-401

### Change description ###
Added terraform properties for enabling proxy in demo environment and 
for proxy hostname and port number.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/infrastructure/demo.tfvars
+++ b/infrastructure/demo.tfvars
@@ -2,7 +2,7 @@ token_validity = "5184000" #60 days in seconds
 vault_section = "preprod"
 reupload_enabled = true
 scan_enabled = true
-proxy_enabled = "true"
+storage_proxy_enabled = "true"
 storage_proxy_host = "proxyout.reform.hmcts.net"
 storage_proxy_port = "8080"
 
--- a/infrastructure/demo.tfvars
+++ b/infrastructure/demo.tfvars
@@ -2,6 +2,9 @@ token_validity = "5184000" #60 days in seconds
 vault_section = "preprod"
 reupload_enabled = true
 scan_enabled = true
+proxy_enabled = true
+storage_proxy_host = "proxyout.reform.hmcts.net"
+storage_proxy_port = 8080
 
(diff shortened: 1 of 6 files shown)


infrastructure/main.tf
```diff
@@ -93,6 +93,9 @@ module "bulk-scan" {
     STORAGE_ACCOUNT_NAME = "${local.storage_account_name}"
     STORAGE_KEY          = "${local.storage_account_primary_key}"
     STORAGE_URL          = "${local.storage_account_url}"
+    PROXY_ENABLED        = "${var.proxy_enabled}"
```
lgonczar: Should this start with `STORAGE_` to match the configuration?

lgonczar: Maybe this should have the same prefix as the other two? Up to you.

aliveni: I'll rename it

infrastructure/demo.tfvars
```diff
@@ -2,6 +2,9 @@ token_validity = "5184000" #60 days in seconds
 vault_section = "preprod"
 reupload_enabled = true
 scan_enabled = true
+proxy_enabled = true
```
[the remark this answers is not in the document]
lgonczar: It would be best if they could all follow the same convention.

src/integrationTest/resources/application.properties
```diff
@@ -14,6 +14,7 @@ storage.url=http://localhost
 storage.blob_copy_timeout_in_millis=30000
 storage.blob_copy_polling_delay_in_millis=300
 storage.blob_lease_timeout=60
+storage.proxy_enabled=false
```
nitinprabhu: This is false by default do we need this ?

aliveni: I tested in local after removing this property. the integration tests are running fine

PR #393, opened 2019-01-03
Fix invalid signature (format) handling

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-443

### Change description ###

Fix the handling of envelopes with signatures of invalid format. Currently, if, for example, the signature is too long for the public key, the processing of the envelope ends with a general failure, instead of the envelope being rejected. This PR fixes it.

Also, because `verifySignature` doesn't return boolean anymore, but throws an exception instead, `verifyFileNames` was changed the same way, for consistency.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
@@ -2,8 +2,6 @@ package uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor;
 
 import com.google.common.base.Strings;
 import com.google.common.io.Resources;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocSignatureFailureException;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.SignatureValidationException;
 
@@ -69,8 +67,6 @@ public class ZipVerifiers {
     public static final String SIGNATURE_SIG = "signature";
     public static final String INVALID_SIGNATURE_MESSAGE = "Zip signature failed verification";
 
-    private static final Logger log = LoggerFactory.getLogger(ZipVerifiers.class);
-
     private ZipVerifiers() {
     }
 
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiersTest.java
```diff
@@ -140,17 +176,34 @@ public void should_verify_signature_using_nonprod_public_key_for_file_signed_usi
...
 
     @Test
     public void should_not_verify_signature_using_wrong_pub_key_for_file_signed_using_nonprod_private_key()
         throws Exception {
         byte[] nonprodZip = toByteArray(getResource("signature/nonprod_envelope.zip")); // inner zip
         byte[] nonprodSignature = toByteArray(getResource("signature/nonprod_envelope_signature"));
-        String nonprodPublicKey =
-            Base64.getEncoder().encodeToString(toByteArray(getResource("nonprod_public_key.der")));
 
-        assertThat(ZipVerifiers.verifySignature(nonprodPublicKey, nonprodZip, nonprodSignature)).isTrue();
```
lgonczar: This test was actually wrong - it should expect a failure.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -125,32 +113,41 @@ static ZipInputStream sha256WithRsaVerification(ZipStreamWithSignature zipWithSi
...
             Signature signature = Signature.getInstance("SHA256withRSA");
             signature.initVerify(publicKey);
             signature.update(data);
-            return signature.verify(signed);
-        } catch (NoSuchAlgorithmException | InvalidKeyException | SignatureException e) {
+            if (!signature.verify(signed)) {
+                throw new DocSignatureFailureException(INVALID_SIGNATURE_MESSAGE);
+            }
+        } catch (SignatureException e) {
+            throw new DocSignatureFailureException(INVALID_SIGNATURE_MESSAGE, e);
```
lgonczar: This case is the reason why I had to make this method throw `DocSignatureFailureException` (with cause attached), instead of returning success/failure.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -91,23 +92,10 @@ static ZipInputStream noOpVerification(ZipStreamWithSignature zipWithSignature)
     static ZipInputStream sha256WithRsaVerification(ZipStreamWithSignature zipWithSignature)
         throws DocSignatureFailureException {
         Map<String, byte[]> zipEntries = extractZipEntries(zipWithSignature.zipInputStream);
-        if (!verifyFileNames(zipEntries)) {
-            log.warn("Signature Failure. Zip entries do not match expected file names. "
```
lgonczar: These warnings are gone, but they are unnecessary - the information is logged at the task level.

PR #396, opened 2019-01-05
Upgrade handler to message session

doncem:

### Change description ###

Upgrade `IMessageHandler` to `ISessionHandler` to have more/full control of the message lifecycle. Will be used for exception handling (abandon, deadletter or just complete)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
@@ -30,6 +30,10 @@ public class ErrorNotificationHandler implements ISessionHandler {
         new Thread(r, "error-notification-service")
     );
 
+    private static final Executor SERVICEBUS_EXEC = Executors.newSingleThreadExecutor(r ->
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandler.java
```diff
@@ -15,7 +16,7 @@
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 
-public class ErrorNotificationHandler implements IMessageHandler {
+public class ErrorNotificationHandler implements ISessionHandler {
```
lgonczar: This solution probably won't work, as `ISessionHandler` is for processing messages in sessions. We don't have sessions enabled in our queue and we don't need them.

I'd use IMessageHandler with `complete` and `deadLetter` functionalities passed to it, either as functions or in one interface. These methods/functions would point to the corresponding methods of the QueueClient in which the handler will be registered. Alternatively, IMessageReceiver could be used - it has the advantage of allowing us to pause reading messages, for example when some dependency is down.

doncem:
follow the comment from @lgonczar

PR #399, opened 2019-01-07
Added ocr data validation for new application

aliveni:
https://tools.hmcts.net/jira/browse/BPS-398

### Change description ###
Added validation to check ocr data exists for new applications.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -147,6 +150,28 @@ public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs)
...
+     *
+     * @param envelope to assert against
+     */
+    public static void assertEnvelopeContainsOcrDataForNewApplication(InputEnvelope envelope) {
+
+        if (Classification.NEW_APPLICATION.equals(envelope.classification)) {
+
+            boolean isValidScannableItems = envelope.scannableItems
+                .stream()
+                .anyMatch(item -> item.documentType.equals(InputDocumentType.SSCS1) && item.ocrData.isEmpty());
```
doncem: so `isValidScannableItems` means specific doc type and empty ocr data. isn't the requirement for ocr data to be present instead? if yes then giving such conditions as is now, the variable proposes to be `isNotValid...`

aliveni: No, ocr data shouldn't be empty. It is also missing jurisdiction condition. I'll push the changes soon

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -147,6 +151,29 @@ public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs)
         }
     }
 
+    /**
+     * Assert envelope contains at least one scannable item
+     * with valid document type and ocr data for new application.
```
rkondratowicz: No mention of the fact that this is only checked for SSCS jurisdiction.

This function seems very specific the way it is right now.
Perhaps it could take parameters?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
```diff
@@ -76,6 +77,31 @@ public void should_throw_exception_when_zip_file_has_mismatching_pdf() throws Ex
             .hasMessageContaining("Missing PDFs: 1111001.pdf");
     }
 
+    @Test
+    public void should_throw_exception_when_ocr_data_is_missing_for_sscs_new_application() throws Exception {
```
lgonczar: It would also be worth adding a test for BlobProcessorTask, checking if this method is used.

src/integrationTest/resources/zipcontents/missing_ocr_data/metadatajson.json
```diff
@@ -0,0 +1,21 @@
+{
+  "po_box": "SSCSPO",
+  "jurisdiction": "SSCS",
+  "delivery_date": "2018-06-23T00:00:00.000Z",
+  "opening_date": "2018-06-24T00:00:00.000Z",
+  "zip_file_createddate": "2018-02-24T00:00:00.000Z",
+  "zip_file_name": "8_24-06-2018-00-00-00.zip",
+  "envelope_classification": "new_application",
+  "scannable_items": [
```
lgonczar: Maybe it would also be good to test a case when there's no SSCS1 document among scannable items (which can contain OCR data)?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -147,6 +151,29 @@ public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs)
...
+     * Throws exception otherwise.
+     *
+     * @param envelope to assert against
+     */
+    public static void assertEnvelopeContainsOcrDataForNewApplication(InputEnvelope envelope) {
```
lgonczar: Currently, we've got a mix of static and non-static `assert...` methods. It would be best if all of them were the same. My preference would be to make them all non-static, as it's easier to test code with those. What do you think?

aliveni: I think non-static is better

src/integrationTest/resources/zipcontents/missing_ocr_data/metadatajson.json
aliveni: 👍

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -147,6 +151,29 @@ public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs)
...
+
+        if (envelope.jurisdiction.equalsIgnoreCase("SSCS")
+            && Classification.NEW_APPLICATION.equals(envelope.classification)) {
+
+            boolean isValidScannableItems = envelope.scannableItems
+                .stream()
+                .anyMatch(item -> item.documentType.equals(InputDocumentType.SSCS1) && !MapUtils.isEmpty(item.ocrData));
```
rkondratowicz: should probably check that **all** sscs1 documents have ocr data, not just at least one of them

lgonczar: There was a long discussion about this at some point and as far as I remember we couldn't get any more specific than that we definitely require at least one sscs1 document with ocr data.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -236,13 +237,15 @@ private ZipFileProcessingResult processZipFileContent(
 
             InputEnvelope envelope = envelopeProcessor.parseEnvelope(result.getMetadata(), zipFilename);
 
+            EnvelopeProcessor.assertEnvelopeContainsOcrDataForNewApplication(envelope);
```
rkondratowicz: perhaps this could be called only for SSCS

lgonczar: or `assertEnvelopeContainsOcrDataIfRequired` or sth like that

rkondratowicz:
Should error notification be sent? if so, an exception needs to be registered in `ErrorMapping` (and test added ;))

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
aliveni: We don't have a unit test for BlobProcessorTask. I think it's better to do it as part of another ticket

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
aliveni: I couldn't think of any other better name, so I renamed it as `assertEnvelopeContainsOcrDataIfRequired`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
aliveni: Updated javadoc to include the conditions for jurisdiction and document_type

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -117,7 +121,7 @@ public Envelope getEnvelopeByFileAndContainer(String container, String zipFileNa
      * @param envelope to assert against
      * @param pdfs     to assert against
      */
-    public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs) {
+    public void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs) {
```
rkondratowicz: unrelated, unnecessary change imho

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -147,6 +151,32 @@ public static void assertEnvelopeHasPdfs(InputEnvelope envelope, List<Pdf> pdfs)
...
+     */
+    public void assertEnvelopeContainsOcrDataIfRequired(InputEnvelope envelope) {
+
+        if (envelope.jurisdiction.equalsIgnoreCase("SSCS")
+            && Classification.NEW_APPLICATION.equals(envelope.classification)) {
+
+            boolean ocrDataExists = envelope.scannableItems
+                .stream()
+                .filter(item -> item.documentType.equals(InputDocumentType.SSCS1))
+                .noneMatch(item -> MapUtils.isEmpty(item.ocrData));
```
rkondratowicz: that doesn't mean ocr data exists, when there are no SSCS1 documents (ocr can be in a document of different type)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
rkondratowicz: more validation coming soon, let's leave it as is, I'll move validation to a separate class on next PR once this is merged.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: I meant 'named', though :)

PR #407, opened 2019-01-09
Feature: message handler custom completor

doncem:

### Change description ###

Introduce custom auto _completor_ for service bus messages. Wrapper model introduced in #397 becomes irrelevant so removed

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandlerTest.java
```diff
@@ -1,60 +1,83 @@
...
 public class ErrorNotificationExceptionHandlerTest {
 
-    private static final IMessage MESSAGE = new Message("content");
+    @Mock
+    private IMessageReceiver receiver;
 
-    private static final ErrorNotificationExceptionHandler HANDLER = new ErrorNotificationExceptionHandler();
+    private ErrorNotificationExceptionHandler handler;
+
+    private static final UUID LOCKE_TOKEN = UUID.randomUUID();
```
lgonczar: A typo

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageAutoCompletor.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    private final BiFunction<UUID, Map<String, Object>, CompletableFuture<Void>> abandonAsyncDelegate;
+
+    private final Function<UUID, CompletableFuture<Void>> completeAsyncDelegate;
+
+    private final BiFunction<UUID, DeadLetterReason, CompletableFuture<Void>> deadLetterAsyncDelegate;
+
+    public MessageAutoCompletor(IMessageReceiver receiver) {
```
lgonczar: I'm not sure we'll be able to get the right receiver. We could use the QueueClient instead.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageAutoCompletor.java
```diff
@@ -0,0 +1,65 @@
...
+        abandonAsyncDelegate = receiver::abandonAsync;
+        completeAsyncDelegate = receiver::completeAsync;
+        deadLetterAsyncDelegate = (uuid, deadLetterReason) -> receiver.deadLetterAsync(
+            uuid,
+            deadLetterReason.reason,
+            deadLetterReason.description
+        );
+    }
+
+    public CompletableFuture<Void> abandonAsync(UUID lockToken, Map<String, Object> propertiesToUpdate) {
```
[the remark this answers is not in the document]
doncem: We use only what's needed. I'll simplify because in my head it looked amazing but I understand how this "functional" look is not quite... functionally desirable

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -1,38 +1,50 @@
...
 
-    private ErrorNotificationMessageWrapper handleErrorNotificationException(
-        IMessage message,
+    private CompletableFuture<Void> handleErrorNotificationException(
+        UUID lockToken,
         ErrorNotificationException exception
     ) {
         if (exception.getStatus().is5xxServerError()) {
-            return ErrorNotificationMessageWrapper.forAbandoning(message);
+            return autoCompletor.abandonAsync(lockToken);
```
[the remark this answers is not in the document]
doncem: Yes we should definitely try to abandon. In another PR it'll have additional property. If abandonment turns out not quite as expected - can easily ignore this completion option

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageAutoCompletor.java
doncem: Oh I should've checked what QueueClient implements. So it's a `MessageAndSessionPump` interface. But perhaps for clarity can use `IQueueClient` one.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandlerTest.java
```diff
@@ -1,60 +1,83 @@
...
     @Test
     public void should_mark_for_acknowledgement_when_no_exception_sent_to_handle() {
-        ErrorNotificationMessageWrapper wrapper = HANDLER.handle(MESSAGE, null);
+        given(client.completeAsync(any(UUID.class))).willReturn(COMPLETED_FUTURE);
 
-        assertThat(wrapper.isAssignedForDeadLettering()).isFalse();
-        assertThat(wrapper.isCompletedAcknowledgement()).isTrue();
+        CompletableFuture<Void> handled = handler.handle(LOCK_TOKEN, null);
+
+        assertThat(handled.join()).isNull();
```
lgonczar: In the acknowledgement test and dead letter test it would be worth checking if we called complete/deadLetter. We probably won't be able to test it at a higher level.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -1,38 +1,43 @@
...
 
-import com.microsoft.azure.servicebus.IMessage;
 import uk.gov.hmcts.reform.bulkscanprocessor.client.ErrorNotificationException;
-import uk.gov.hmcts.reform.bulkscanprocessor.model.ErrorNotificationMessageWrapper;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.MessageAutoCompletor;
+
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
 
 public class ErrorNotificationExceptionHandler {
```
lgonczar: Something I didn't think about earlier - it would be good to add some logging, so that we can track the history of the message.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -1,38 +1,43 @@
...
         }
     }
 
-    private ErrorNotificationMessageWrapper handleErrorNotificationException(
-        IMessage message,
+    private CompletableFuture<Void> handleErrorNotificationException(
+        UUID lockToken,
         ErrorNotificationException exception
     ) {
         if (exception.getStatus().is5xxServerError()) {
```
lgonczar: This condition is not a part of this PR, so nothing to change as a part of it, but here's a thought: what if we get a 408 (timeout), for example?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandlerTest.java
doncem: we define in the given/state. otherwise tests will fail. verifying interaction would be putting more butter on top

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
doncem: Surely. Must've gone between multiple branches I have locally. ..or was not there at all. Gimme couple minutes

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
doncem: Yes, I think I mentioned before that conditions to decide when to void/deadletter can be grained more finely later on

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandlerTest.java
lgonczar: If we commented out the call to `MessageAutoCompletor.complete` in the app code, would this test fail?

doncem: I won't fail

P.S. There's `MessageAutoCompletorTest` which does what you are after

lgonczar: Yes, those are proper tests. Here I was thinking about verifying if we're calling those two AutoCompletor's methods (which are tested) where we're expected to.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -34,9 +40,12 @@ public ErrorNotificationExceptionHandler(MessageAutoCompletor autoCompletor) {
...
+            log.warn("Received server error from notification client", exception);
```
lgonczar: Worth including message Id in those logs, so that we can track what happened to particular messages.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
rkondratowicz: it may also be useful to log what we decide to do with the message itself (not just what the problem was).
Perhaps in the autocompletor class.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -1,38 +1,55 @@
...
-    private ErrorNotificationMessageWrapper handleErrorNotificationException(
-        IMessage message,
+    private CompletableFuture<Void> handleErrorNotificationException(
+        UUID lockToken,
+        String messageId,
         ErrorNotificationException exception
     ) {
         if (exception.getStatus().is5xxServerError()) {
-            return ErrorNotificationMessageWrapper.forAbandoning(message);
+            log.warn("Received server error from notification client. Message ID: {}", messageId, exception);
```
rkondratowicz: Is there a way to get the delivery count on from a queue message?
Could potentially be a useful information.

lgonczar: I think there is a way. May be useful, but not crucial - we can still check how many times we failed processing.

PR #411, opened 2019-01-10
Validate envelope against container

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-453

Requires updating tests (test container is called 'BULKSCAN' so 'SSCS' jurisdiction was invalid)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
```diff
@@ -180,6 +181,35 @@ public void should_not_throw_exception_when_ocr_data_is_not_missing() throws Exc
         assertThat(throwable).isNull();
     }
 
+    @Test
+    public void should_throw_an_exception_when_jurisdiction_and_container_dont_match() {
+        // given
+        InputEnvelope envelope = inputEnvelope("A");
```
lgonczar: I'd probably make the jurisdiction and container a bit longer, in order to make sure the assertions related with exception message don't pass by accident.

PR #412, opened 2019-01-10
Update OCR presence validation

rkondratowicz:
Checked with Ade and envelope of classification `NEW_APPLICATION` should always have ocr data in it.
(so at least one document of type `SSCS1` in case of `SSCS` jurisdiction)

Also updated tests to not use files from resources.


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/EnvelopeValidator.java
```diff
@@ -1,47 +1,67 @@
...
      * @param envelope to assert against
      */
     public static void assertEnvelopeContainsOcrDataIfRequired(InputEnvelope envelope) {
 
-        if (envelope.jurisdiction.equalsIgnoreCase("SSCS")
-            && Classification.NEW_APPLICATION.equals(envelope.classification)) {
+        if (Classification.NEW_APPLICATION.equals(envelope.classification)) {
 
-            boolean ocrDataExists = envelope.scannableItems
+            InputDocumentType typeThatShouldHaveOcrData = ocrDocumentTypePerJurisdiction.get(envelope.jurisdiction);
```
lgonczar: This will cause us to fail when NEW_APPLICATION envelopes start showing up for other services. It would be better to reject them.

rkondratowicz: apparently `NEW_APPLICATION` only makes sense if there is OCR present on a service specific document type.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/EnvelopeProcessorValidationTest.java
```diff
@@ -78,22 +85,75 @@ public void should_throw_exception_when_zip_file_has_mismatching_pdf() throws Ex
...
         assertThat(throwable).isInstanceOf(OcrDataNotFoundException.class)
-            .hasMessageContaining("No scannable items found with ocr data and document type SSCS1");
+            .hasMessageContaining("No documents");
     }
 
     @Test
-    public void should_not_throw_exception_when_ocr_exists_for_document_type_other() throws Exception {
-        ZipFileProcessingResult processingResult = processZip("zipcontents/ocr_data_not_required");
-        InputEnvelope envelope = EnvelopeCreator.getEnvelopeFromMetafile(processingResult.getMetadata());
+    public void should_throw_exception_when_document_type_for_new_applications_is_not_configured_for_given_jurisdiction() throws Exception {
```
lgonczar: At some point we should split those tests by methods that are tested

rkondratowicz: Or even have a separate class for every validation (for easier navigation)

lgonczar: Yes, that's exactly the kind of split I had in mind

PR #415, opened 2019-01-11
Enable notification consumption feature

doncem:

[Create scheduled job for processing notification](https://tools.hmcts.net/jira/browse/BPS-288)

### Change description ###

Provide configuration for error notification queue consumption:

- separate queue client listener
- wrapped in enable/disable flag
  - by default it's enabled
  - in prod it's disabled
  - locally it's disabled
- custom message handler registered with auto completion by servicebus disabled

Future work: functional test(s?). As quite a sum of code will have to be written, I have excluded initial part of it done last year and will continue working separately from setup. There is no harm for production as config is disabled ^. Can disable globally - just say the word

```
[ ] Yes
[x] No
```


infrastructure/main.tf
```diff
@@ -124,6 +126,7 @@ module "bulk-scan" {
 
     QUEUE_ENVELOPE_SEND = "${data.terraform_remote_state.shared_infra.queue_primary_send_connection_string}"
     QUEUE_NOTIFICATIONS_SEND = "${data.terraform_remote_state.shared_infra.notifications_queue_primary_send_connection_string}"
+    QUEUE_NOTIFICATIONS_READ = "${data.terraform_remote_state.shared_infra.notification_primary_listen_connection_string}"
```
rkondratowicz: should it not be `notifications_queue_primary_listen_connection_string ` instead?
https://github.com/hmcts/bulk-scan-shared-infrastructure/blob/master/queue.tf#L45

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
```diff
@@ -44,4 +59,32 @@ public ServiceBusHelper notificationsQueueHelper(
             objectMapper
         );
     }
+
+    @Bean(name = "notifications-read")
+    @ConditionalOnProperty(prefix = "queues.read-notifications", name = "enabled")
+    public IQueueClient notificationsQueueReader(
```
rkondratowicz: will this bean be used in next PR?

infrastructure/main.tf
doncem: ah it has changed since this part was coded

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
doncem: reader will be enabled in aks and functional test should assert notification has been sent. i think i need help in setting up configs though

infrastructure/main.tf
rkondratowicz: 11 Dec 2018 :)

infrastructure/prod.tfvars
```diff
@@ -4,3 +4,4 @@ api_gateway_test_valid_certificate_thumbprint = "3B0D23863EE7EF90290AD02B0ACFD7C
 allowed_client_certificate_thumbprints = ["919F8A770C2A2F06933BA11551001ED0581B8848"]
 
 blob_processing_delay_in_minutes = "30"
+error_notifications_enabled = "false"
```
lgonczar: I'm a fan of default disabling and enabling explicitly for each environment, so that nothing is enabled in prod by mistake. This is just a preference, so up to you.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
```diff
@@ -1,25 +1,55 @@
...
+    private ErrorNotificationService errorNotificationService;
+
+    @Value("${queues.read-notifications.enabled}")
+    private boolean readNotificationsEnabled;
+
+    @Value("${queues.read-notifications.connection-string}")
+    private String connectionString;
+
+    @Value("${queues.read-notifications.queue-name}")
+    private String queueName;
```
lgonczar: For `queueName` and `connectionString` it would be worth adding some prefix to make it clear these fields are related with notifications queue. This can be deducted from the annotations above declarations, but you'd have to navigate there to find out.

src/main/resources/application.yaml
```diff
@@ -60,6 +60,10 @@ queues:
   notifications:
     connection-string: ${QUEUE_NOTIFICATIONS_SEND:"NO_VALUE_SUPPLIED"}
     queue-name: notifications
+  read-notifications:
+    connection-string: ${QUEUE_NOTIFICATIONS_READ:"NO_VALUE_SUPPLIED"}
```
lgonczar: I can see this has been done by analogy to the other queue, so it's ok, but I wonder if we need a default value at all. Shouldn't the app fail to start if the env property is missing?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
```diff
@@ -44,4 +73,26 @@ public ServiceBusHelper notificationsQueueHelper(
             objectMapper
         );
     }
+
+    @PostConstruct
```
lgonczar: Is it possible for this to be a bean (returning QueueClient)?

lgonczar: My bad, never mind

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
doncem: fair point. can be amended in case other `@Value`s come into play. but it's not arm-breaking to include some prefix here

src/main/resources/application.yaml
doncem: in real life, if default is used - app will fail because of failure to parse the connection string. i believe this is provided so that application can be loaded in development. it also takes play in integration tests

infrastructure/prod.tfvars
doncem: depends on a feature, i suppose. right now error notifications were developed "enabled" by default. so it make no sense to disable consumption by default. except those environments which matter.

i have no preference just try to make approach sensible. all the message should be dead lettered by now so even if enabled now on some testing environment - it'll be fine to test out the functionality as discussed few times.

p.s. there should not be a mistake on production. it is disabled. in order to enable there has to be manual intervention

p.p.s. as mentioned in description - can disable globally if requested. stating as preference is not an argument giving the position described above ^^

src/main/resources/application.yaml
lgonczar: fair enough

PR #420, opened 2019-01-14
Add reports controller

rkondratowicz:
Jira: https://tools.hmcts.net/jira/browse/BPS-454

Just the API part here.
Db part to be added in next PR.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
@@ -1,0 +1,39 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.controllers;
+
+import io.swagger.annotations.ApiOperation;
+import org.springframework.format.annotation.DateTimeFormat;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.reports.EnvelopeCountSummaryReportResponse;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.EnvelopeCountSummary;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.ReportsService;
+
+import java.time.LocalDate;
+
+import static org.springframework.format.annotation.DateTimeFormat.ISO.DATE;
+
+@RestController
+@RequestMapping(path = "/reports", produces = MediaType.APPLICATION_JSON_VALUE)
+public class ReportsController {
+
+    private final ReportsService reportsService;
+
+    public ReportsController(ReportsService reportsService) {
+        this.reportsService = reportsService;
+    }
+
+    @GetMapping(path = "/count-summary")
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
```diff
@@ -0,0 +1,39 @@
...
+    private final ReportsService reportsService;
+
+    public ReportsController(ReportsService reportsService) {
+        this.reportsService = reportsService;
+    }
+
+    @GetMapping(path = "/count-summary")
+    @ApiOperation("Retrieves envelope count summary report")
+    public EnvelopeCountSummaryReportResponse getCountSummary(
+        @RequestParam(name = "date") @DateTimeFormat(iso = DATE) LocalDate date
```
lgonczar: How about path variable?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/reports/EnvelopeCountSummaryReportResponse.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.model.out.reports;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class EnvelopeCountSummaryReportResponse {
+
+    @JsonProperty("received")
```
lgonczar: Probably don't need the string names.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReportsControllerTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+        mockMvc
+            .perform(get("/reports/count-summary?date=2019-01-14"))
+            .andExpect(status().isOk())
+            .andExpect(jsonPath("$.received").value(received))
+            .andExpect(jsonPath("$.rejected").value(rejected));
+    }
+
+    @Test
+    public void should_return_400_if_date_is_invalid() throws Exception {
```
lgonczar: Alternatively, could just return 404

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
nitinprabhu: should we accept start date and end date rather than just date. This way we can get reports for a particular month/week if required. Not sure if it required now but something may be nice to have

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReportsControllerTest.java
rkondratowicz: invalid date sounds like an invalid parameter value, thus client error 400

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
lgonczar: Or we might need to have daily counts for a specific period. I don't know. Maybe we should stick to the simplest version and once something extended is required, we can adjust.

rkondratowicz: I went with this as it's both simple and allows extension with other filtering in the future (for example `endDate` as @nitinprabhu suggested, but also `jurisdiction` etc)

PR #423, opened 2019-01-15
Enable envelope count summary report

rkondratowicz:
tying the pieces built separately in previous PRs together.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
@@ -1,15 +1,32 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.services.reports;
 
-import org.apache.commons.lang3.NotImplementedException;
 import org.springframework.stereotype.Service;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.reports.EnvelopeCountSummaryRepository;
 
 import java.time.LocalDate;
 import java.util.List;
 
+import static java.util.stream.Collectors.toList;
+
 @Service
 public class ReportsService {
 
+    private final EnvelopeCountSummaryRepository repo;
+
+    public ReportsService(EnvelopeCountSummaryRepository repo) {
+        this.repo = repo;
+    }
+
     public List<EnvelopeCountSummary> getCountFor(LocalDate date) {
-        throw new NotImplementedException("Not yet implemented");
+        return repo
+            .getReportFor(date)
+            .stream()
+            .map(it -> new EnvelopeCountSummary(
+                it.getReceived(),
+                it.getRejected(),
+                it.getJurisdiction(),
+                it.getDate()
+            ))
+            .collect(toList());
     }
 }
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
@@ -13,6 +13,7 @@ import java.util.List;
 
 import static java.time.LocalDate.now;
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.BDDMockito.given;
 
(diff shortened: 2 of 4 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+public class ReportsServiceTest {
+
+    @Mock
+    EnvelopeCountSummaryRepository repo;
+
+    @Test
+    public void should_map_repo_result_properly() {
+        given(repo.getReportFor(now()))
+            .willReturn(asList(
+                new Item(now(), "A", 100, 1),
```
lgonczar: Best to use different dates

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+        ReportsService service = new ReportsService(repo);
+
+        // when
+        List<EnvelopeCountSummary> result = service.getCountFor(now());
+
+        // then
+        assertThat(result)
+            .usingFieldByFieldElementComparator()
+            .containsExactlyInAnyOrder(
```
lgonczar: Maybe we should require the same order? We might sort by jurisdiction at some point.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
```diff
@@ -1,15 +1,32 @@
...
+
+    public ReportsService(EnvelopeCountSummaryRepository repo) {
+        this.repo = repo;
+    }
+
     public List<EnvelopeCountSummary> getCountFor(LocalDate date) {
-        throw new NotImplementedException("Not yet implemented");
+        return repo
+            .getReportFor(date)
+            .stream()
```
doncem: i would favour for repository returning stream instead. especially when it has such possibility and reports make sense to be streamed instead of evaluated as early as possible

not part of the style though

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+        ReportsService service = new ReportsService(repo);
+
+        // when
+        List<EnvelopeCountSummary> result = service.getCountFor(now());
+
+        // then
+        assertThat(result)
+            .usingFieldByFieldElementComparator()
+            .containsExactly(
+                new EnvelopeCountSummary(100, 1, "A", now().plusDays(1)),
```
lgonczar: I'd probably reference dates via variables, as every now and then the millisecond will be different

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.mockito.BDDMockito.given;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ReportsServiceTest {
+
+    @Mock
+    EnvelopeCountSummaryRepository repo;
+
+    @Test
+    public void should_map_repo_result_properly() {
```
lgonczar: To consider: a case when the repo returns an empty list. Up to you.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
lgonczar: Aaa, it's LocalDate. All good.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
rkondratowicz: added

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
rkondratowicz: I'll do that on next PR (to not mess with repo layer and its' tests in this PR)

PR #424, opened 2019-01-16
Allow filtering out test jurisdiction from envelope count report

rkondratowicz:
By default `BULKSCAN` envelopes are now excluded from the report, but it is possible to add extra query parameter to include them (for testing purposes).

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReportsControllerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReportsControllerTest.java
@@ -64,6 +64,13 @@ public class ReportsControllerTest {
     }
 
     @Test
+    public void should_not_include_test_jurisdiction_if_exlicitly_not_requested_by_the_client() throws Exception {
+        mockMvc.perform(get("/reports/count-summary?date=2019-01-14&include-test=false"));
+
+        verify(reportsService).getCountFor(LocalDate.of(2019, 1, 14), false);
+    }
+
+    @Test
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsServiceTest.java
```diff
@@ -49,13 +50,30 @@ public void should_map_repo_result_properly() {
             );
     }
 
+    @Test
+    public void should_filter_out_test_jurisdiction_when_requested() {
```
lgonczar: To make it perfect, we could also check if `resultWithoutTestJurisdiction` filters out BULKSCAN and not the other jurisdiction.

rkondratowicz: added extra assertion

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReportsControllerTest.java
```diff
@@ -48,6 +49,20 @@ public void should_return_result_generated_by_the_service() throws Exception {
             .andExpect(jsonPath("$.data[0].date").value(countSummary.date.toString()));
     }
 
+    @Test
+    public void should_not_include_test_jurisdiction_by_default() throws Exception {
```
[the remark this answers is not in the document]
rkondratowicz: done

PR #426, opened 2019-01-16
Add helper to add 'zero' rows to report

rkondratowicz:
There's a new requirement, that there should be a row for every existing jurisdiction, even if there were no envelopes in DB for given day and jurisdiction.

I'll plug this helper to the service layer in next PR.

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ZeroRowFillerTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ZeroRowFillerTest.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.services.reports;
 
-
 import org.junit.Test;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ZeroRowFillerTest.java
```diff
@@ -0,0 +1,55 @@
...
+            .usingFieldByFieldElementComparator()
+            .containsExactly(
+                new EnvelopeCountSummary(100, 101, "A", now()),
+                new EnvelopeCountSummary(200, 201, "B", now()),
+                new EnvelopeCountSummary(0, 0, "C", now())
+            );
+    }
+
+    @Test
+    public void should_not_change_input_list_if_all_jurisdictions_are_present() {
```
lgonczar: Up to you: a case when there are more jurisdictions present than configured.

lgonczar:
@doncem The parameter is a bit misleading, but all data is for the same date. Theoretically, it could be retrieved from any item in the list, but there's no guarantee there will be any.

doncem:
not sure which parameter you are referring but i'm sure every _names_ can be changed to accommodate every needs :)

lgonczar:
I meant the previous comment, in which you provided code. I can see some grouping by date there, but it's just one date for all data.

PR #429, opened 2019-01-17
Added db migration script to update ocrData column type to JSON

aliveni:
OcrData column in the envelope is not being used at the moment.
Another PR will be created to use this column and drop the existing column (`ocrDataJson`)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
@@ -42,7 +42,7 @@ public class ScannableItem implements EnvelopeAssignable {
     private Timestamp nextActionDate;
 
     @Type(type = "jsonb")
-    @Column(name = "ocrDataJson", columnDefinition = "jsonb")
+    @Column(name = "ocrData", columnDefinition = "jsonb")
     private Map<String, String> ocrData;
 
     private String fileName;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
```diff
@@ -42,7 +42,7 @@
     private Timestamp nextActionDate;
 
     @Type(type = "jsonb")
-    @Column(name = "ocrDataJson", columnDefinition = "jsonb")
+    @Column(name = "ocrData", columnDefinition = "jsonb")
```
lgonczar: We should only switch to the new column once the application has started storing OCR data in the right format

aliveni: 👍

src/main/resources/db/migration/V025__Update_ocrData_column.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE scannable_items
+  DROP COLUMN ocrData;
+
+ALTER TABLE scannable_items
+  ADD COLUMN ocrData JSON null
```
rkondratowicz: It took my a while to realise `ocrData` column is currently not used.
I think this migration deserves a comment, at least on the PR.

aliveni: I'll add a comment. 
We want to do this in 2 PRs. The first PR to add the field and the second PR will have the migration script to use to drop the existing `ocrDataJson` and use the field `ocrData`.

lgonczar:
Why not merge this one?

src/main/resources/db/migration/V025__Update_ocrData_column.sql
lgonczar: Dropping ocrDataJson alone will need to be done in a separate deployment, as the last change.

aliveni: 👍

PR #430, opened 2019-01-17
Add message ID to error notification log

lgonczar:
### Change description ###

Add message ID to error notification log, so that it can be connected with file name and envelope ID in logs.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -290,20 +291,7 @@ private void handleInvalidFileError(
...
-                            containerName,
-                            null,
-                            null,
-                            errorCode,
-                            cause.getMessage()
-                        )
-                    );
-
-                    log.info("Sent error notification for file {} in container {}", zipFilename, containerName);
+                    sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
```
rkondratowicz: if you really want to extrac this to a function, perhaps include the catch block (so that it's self contained)?

lgonczar: This method will still stay as it is - it's long enough. I can wrap it in `trySendErrorMessageToQueue`, which would include exception handling.

PR #433, opened 2019-01-21
BPS-447 Keep ocr data in the same format as we receive in the message

aliveni:
https://tools.hmcts.net/jira/browse/BPS-447

### Change description ###
Updated the code to keep to keep ocr data in the same format 
as we receive in envelope message.

```
[ ] Yes
[x ] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
```diff
@@ -152,6 +155,25 @@ public static Envelope envelopeNotified() {
...
 
+    private static OcrData ocrData(Map<String, String> data) {
+        OcrData ocrData = new OcrData();
+        List<OcrDataField> ocrDataFields = data.entrySet()
+            .stream()
+            .map(
+                e -> {
+                    OcrDataField field = new OcrDataField();
+                    field.setMetadataFieldName(new TextNode("name1"));
+                    field.setMetadataFieldValue(new TextNode("value1"));
```
rkondratowicz: I presume `e` should be used

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
```diff
@@ -1,31 +1,32 @@
...
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.fasterxml.jackson.databind.node.ValueNode;
 
 public class OcrDataField {
 
-    private TextNode name;
-    private ValueNode value;
+    @JsonProperty(value = "metadata_field_name", required = true)
+    private TextNode metadataFieldName;
+    private ValueNode metadataFieldValue;
```
rkondratowicz: we can keep the field names unchanged. IMHO it will be clearer this way.

rkondratowicz: don we not need `@JsonProperty(value = "metadata_field_value")` on this one?

aliveni: No, we don't need `@JsonProperty` here. forgot to remove it, thanks!

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
aliveni: 👍

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
```diff
@@ -9,45 +9,38 @@
...
-                field -> field.getValue().asText(""),
-                (value1, value2) -> {
-                    throw new IllegalStateException("Ocr data contains duplicate fields");
-                },
-                LinkedHashMap::new
-            )
+        ocrData.getFields().forEach(
+            ocrDataField -> {
+                if (ocrDataField.getMetadataFieldName().isNull()) {
+                    throw new IllegalStateException("Ocr data field name must be provided.");
```
rkondratowicz: is this not already done by Jackson? `@JsonSetter(value = "metadata_field_name", nulls = Nulls.FAIL)`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
aliveni: > we can keep the field names unchanged. IMHO it will be clearer this way.

Do you mean the field names should be `name` and `value`? 
But we want to send the ocr data to services without making any changes. isn't it?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
```diff
@@ -16,11 +15,11 @@ public TextNode getName() {
     }
 
     @JsonSetter(value = "metadata_field_name", nulls = Nulls.FAIL)
-    public void setName(TextNode name) {
-        this.name = name;
+    public void setName(TextNode metadataFieldName) {
+        this.name = metadataFieldName;
```
rkondratowicz: can still be `name`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
rkondratowicz: If name is missing then this `ocrDataField.getName().isNull()` would throw a NPE, right?
What if we use jackson's `required=true`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
```diff
@@ -9,45 +9,38 @@
 
 import java.io.IOException;
 import java.util.Base64;
-import java.util.LinkedHashMap;
-import java.util.Map;
 
-import static java.util.stream.Collectors.toMap;
-
-public class OcrDataDeserializer extends StdDeserializer<Map<String, String>> {
+public class OcrDataDeserializer extends StdDeserializer<OcrData> {
```
rkondratowicz: is this custom deserializer still needed, when we use our own class with Jackson annotations??

aliveni: @rkondratowicz I tried not to use custom deserializer but it doesn't work without it, because the ocr_data is Base64 encoded. I tried to find if there is any annotation to specify that that value is in Base64 format, but I couldn't find anything. Any suggestions?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
rkondratowicz: that's why I suggested adding `@JsonProperty(value = "metadata_field_value")` so that it gets serialized to JSON in original format.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializer.java
rkondratowicz: oops, forgot it's base 64 encoded

rkondratowicz: missleading class name

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
aliveni: I tried it but it's not working

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/ocr/OcrDataField.java
```diff
@@ -1,31 +1,25 @@
...
-import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.node.TextNode;
 import com.fasterxml.jackson.databind.node.ValueNode;
 
 public class OcrDataField {
 
+    @JsonProperty(value = "metadata_field_name")
     private TextNode name;
```
lgonczar: How about making `name` and `value` strings?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/util/OcrDataDeserializerTest.java
```diff
@@ -21,30 +29,19 @@
...
+        List<OcrDataField> ocrDataFields = Arrays.asList(
+            createOcrDataField(new TextNode("text_field"), new TextNode("some text")),
+            createOcrDataField(new TextNode("number_field"), new IntNode(123)),
+            createOcrDataField(new TextNode("boolean_field"), BooleanNode.TRUE),
+            createOcrDataField(new TextNode("null_field"), NullNode.instance)
         );
 
-        assertThat(result).isEqualTo(expectedResult);
-    }
+        expectedOcrData.setFields(ocrDataFields);
```
lgonczar: Not strictly related with this PR, but I wonder if OcrData could be made immutable.

lgonczar:
Still worth changing field name and value to strings, but not crucial for this PR

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItem.java
```diff
@@ -43,7 +43,7 @@
 
     @Type(type = "jsonb")
     @Column(name = "ocrDataJson", columnDefinition = "jsonb")
```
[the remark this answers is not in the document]
aliveni: Updated to use the new column `ocrData`

aliveni:
> Is there a test for what json is being stored in database?
> I think you'll need to add annotation on fields in `OcrDataField` to keep the metadata field names.

There is an existing integration test where we are verifying all columns in envelope. Also, the annotations are added to `OcrDataField`

PR #435, opened 2019-01-22
Bug - move files to rejected container in case of an unrecognised error

rkondratowicz:
`handleEventRelatedError` just creates an event in db.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -267,7 +267,7 @@ public class BlobProcessorTask extends Processor {
             return null;
         } catch (Exception ex) {
             log.error("Failed to process file {} from container {}", zipFilename, containerName, ex);
-            handleEventRelatedError(Event.DOC_FAILURE, containerName, zipFilename, ex);
+            handleInvalidFileError(Event.DOC_FAILURE, containerName, zipFilename, leaseId, ex);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -263,11 +263,11 @@ private ZipFileProcessingResult processZipFileContent(
             return null;
         } catch (PreviouslyFailedToUploadException ex) {
             log.warn("Rejected file {} from container {} - failed previously", zipFilename, containerName, ex);
-            handleEventRelatedError(Event.DOC_UPLOAD_FAILURE, containerName, zipFilename, ex);
+            registerEvent(Event.DOC_UPLOAD_FAILURE, containerName, zipFilename, ex.getMessage());
```
rkondratowicz: to make it clear what's actually happening

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -284,7 +284,7 @@ private void handleInvalidFileError(
         String leaseId,
         Exception cause
     ) {
-        Long eventId = handleEventRelatedError(fileValidationFailure, containerName, zipFilename, cause);
+        Long eventId = registerEvent(fileValidationFailure, containerName, zipFilename, cause.getMessage());
```
rkondratowicz: and here.

PR #441, opened 2019-01-24
Adjust job  delays on prod

rkondratowicz:
main job: 10sec -> 1min
reupload job: 10sec -> 10min


lgonczar:
Do we even need the main job to run every minute? Why not 10, as well?

PR #445, opened 2019-01-24
Fix handling of exception occurring during future

doncem:

[Create scheduling job for processing notifications](https://tools.hmcts.net/jira/browse/BPS-288)

### Change description ###

Due to planned work on functional tests the essential integration test duplicating the work was left behind hence missing out one of the insidious tiny obvious thingy: exception which is handled in `CompletableFuture` is of instance `CompletionException` with the cause of original exception that we were after.

This PR introduces full blown integration test to verify this.

Further details, this is the culprit fixer:

```java
private Throwable processThrowable(Throwable throwable) {
    return throwable instanceof CompletionException ? throwable.getCause() : throwable;
}
```

Although it will always be `CompletionException`, the check is there for fullness

```
[ ] Yes
[x] No
```


src/integrationTest/resources/logback-test-base.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lgonczar: Why do we need to override logback configuration for integration tests?

doncem: ooh, embarrassing. i started typing output captures but my heart said it's not gonna work. so ended up over-engineering

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandlerTest.java
```diff
@@ -0,0 +1,249 @@
...
+                    "po_box",
+                    null,
+                    ERR_FILE_LIMIT_EXCEEDED,
+                    "description"
+                )
+            ),
+            MediaType.APPLICATION_JSON_UTF8_VALUE
+        ));
+
+        given(message.getLockToken()).willReturn(UUID.randomUUID());
```
nitinprabhu: worth putting these two given will return outside of this method

doncem: not sure how it's worth. these are dedicated to activity caused within servicebus for which we have no control of. there is no reason to verify them just make sure source code doesn't crash with null pointers

rkondratowicz:
can a `catch` block be added in the actual service instead?
```
try {
    ErrorNotificationResponse response = client.notify(request);
    entity.setNotificationId(response.getNotificationId());
    log.info("Error notification for {} published. ID: {}", message.zipFileName, response.getNotificationId());
} finally {
    repository.save(entity);
}
```

doncem:
it cannot. at least it cannot silence the exception throws. every single exception must be dealt in exception handler where the lifecycle of service bus message is decided

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptionhandlers/ErrorNotificationExceptionHandler.java
```diff
@@ -33,24 +34,35 @@ public ErrorNotificationExceptionHandler(MessageAutoCompletor autoCompletor) {
...
-        if (throwable instanceof ErrorNotificationException) {
+        Throwable processedThrowable = processThrowable(throwable);
+
+        if (processedThrowable instanceof ErrorNotificationException) {
             return handleErrorNotificationException(
                 lockToken,
                 messageId,
                 deliveryCount,
-                (ErrorNotificationException) throwable
+                (ErrorNotificationException) processedThrowable
```
rkondratowicz: may be worth logging a warning here as well.

doncem: yes it is, in that method 😉

PR #455, opened 2019-01-29
Check if blob exists before trying to process it

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-497

### Change description ###

Check if blob exists before trying to process it. Also, prevent failure of the whole job when the processing of a single file fails (i.e. keep processing files). 

Backstory: multiple jobs run at the same time, each doing the following:
- retrieve all blobs from given container
- process the blobs in random order

This can (and does) lead to situations when two jobs retrieve blob lists containing the same blob and by the time one job has got to the file, the other job has already processed and deleted it. There are two problems related with this:
1. Processing fails because file doesn't exist and we're polluting logs with this kind of errors
2. The whole job fails, i.e. no other files are processed

This PR addresses both issues.

Unit tests don't exist for `BlobProcessorTask`, only integration tests. There's a story to introduce them (https://tools.hmcts.net/jira/browse/BPS-489), but it's a bigger piece of work, as the code will need to be reformatted in order to allow for proper testing.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -140,7 +140,7 @@ public class BlobProcessorTask extends Processor {
         try {
             processZipFileIfEligible(container, zipFilename);
         } catch (Exception ex) {
-            log.error("Failed to process file {} from container {}", ex);
+            log.error("Failed to process file {} from container {}", zipFilename, container.getName(), ex);
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -108,7 +108,7 @@ public class BlobProcessorTask extends Processor {
     }
 
     @Scheduled(fixedDelayString = "${scheduling.task.scan.delay}")
-    public void processBlobs() throws IOException, StorageException, URISyntaxException {
+    public void processBlobs() {
         log.info("Started blob processing job");
 
         for (CloudBlobContainer container : blobManager.listInputContainers()) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -131,41 +130,50 @@ private void processZipFiles(CloudBlobContainer container)
...
 
         log.info("Finished processing blobs for container {}", container.getName());
     }
 
-    private void processZipFile(CloudBlobContainer container, String zipFilename)
+    private void tryProcessZipFile(CloudBlobContainer container, String zipFilename) {
+        try {
+            processZipFileIfEligible(container, zipFilename);
+        } catch (Exception ex) {
+            log.error("Failed to process file {} from container {}", ex);
```
rkondratowicz: missing file nae param

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -323,4 +331,28 @@ private void sendErrorMessageToQueue(
...
+
+    private void logAbortedProcessingNonExistingFile(String zipFilename, String containerName) {
+        log.info(
+            "Aborted processing of zip file {} from container {} - doesn't exist anymore.",
+            zipFilename,
+            containerName
+        );
+    }
+
+    private void logAbortedProcessingFilePresentInDb(String containerName, String zipFilename) {
```
rkondratowicz: all 3 methods log almost identical info.
Perhaps you could have just one method, something like
```
logProcessingAborted(String container, String fileName, String reason)
```
just an idea.

lgonczar: I was considering that myself, but decided against it, because there would be a hidden requirement for the reason to be a properly formatted string.

PR #457, opened 2019-01-29
Add service for returning zip file status

rkondratowicz:
### Create end-point to return envelope status and events by zip file name
https://tools.hmcts.net/jira/browse/BPS-451

1. **Add service layer**
2. Add controller layer

In this PR I just add the service that will be called by the new endpoint.

Note that there may exist multiple envelopes with the same zip file name (in different jurisdictions).

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
@@ -12,6 +12,7 @@ import java.util.List;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static uk.gov.hmcts.reform.bulkscanprocessor.helper.EnvelopeCreator.envelope;
 
 @SuppressWarnings("checkstyle:LineLength")
@@ -58,6 +59,24 @@ public class EnvelopeRepoTest {
 
         // then
         assertThat(result).hasSize(1);
+    }
+
+    @Test
+    public void findByZipFileName_should_find_envelops_in_db() {
+        // given
+        dbHas(
+            envelope("A.zip", "X", Status.PROCESSED),
+            envelope("A.zip", "Y", Status.UPLOAD_FAILURE),
+            envelope("B.zip", "Z", Status.UPLOAD_FAILURE)
+        );
+
+        // when
+        List<Envelope> resultForA = repo.findByZipFileName("A.zip");
+        List<Envelope> resultForX = repo.findByZipFileName("X.zip");
+
+        // then
+        assertThat(resultForA).hasSize(2);
+        assertThat(resultForX).hasSize(0);
     }
 
     @After
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
@@ -12,7 +12,6 @@ import java.util.List;
 
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static uk.gov.hmcts.reform.bulkscanprocessor.helper.EnvelopeCreator.envelope;
 
 @SuppressWarnings("checkstyle:LineLength")
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
@@ -25,6 +25,8 @@ public interface EnvelopeRepository extends JpaRepository<Envelope, UUID> {
 
     List<Envelope> findByStatus(Status status);
 
+    List<Envelope> findByZipFileName(String zipFileName);
+
     /**
      * Finds envelope with a blob not deleted for a given container and zip file name.
      *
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -25,6 +25,8 @@
 
     List<Envelope> findByStatus(Status status);
 
+    List<Envelope> findByZipFileName(String zipFileName);
```
nitinprabhu: I guess Exela generates different file name for two different zip files uploaded to two different containers at same time otherwise it should be by zip file name and container/jurisdiction ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusServiceTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.BDDMockito.given;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ZipFileStatusServiceTest {
```
nitinprabhu: Is it worth adding a test where only process event record exists and no record in envelope table ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
rkondratowicz: the story mentions retrieving status by file name only, not jurisdiction, hence the list.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusServiceTest.java
rkondratowicz: I think `should_return_empty_lists_when_no_data_for_zip_file_was_found` covers it I think

doncem: yes it covers but not proves. there should be an integration test case for each repository method introduced in the pr so you can verify it definitely returns empty list

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
lgonczar: Worth adding some tests

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusService.java
```diff
@@ -0,0 +1,29 @@
...
+    // region constructor
+
+    public ZipFileStatusService(ProcessEventRepository eventRepo, EnvelopeRepository envelopeRepo) {
+        this.eventRepo = eventRepo;
+        this.envelopeRepo = envelopeRepo;
+    }
+
+    // endregion
+
+    public ZipFileStatus getStatusFor(String zipFileName) {
```
lgonczar: Should we make it transactional to ensure data consistency between queries?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusServiceTest.java
rkondratowicz: Added integration tests for repos.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
rkondratowicz: Added tests for new repo methods

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
```diff
@@ -60,6 +61,24 @@ public void findEnvelopesToResend_should_filter_based_on_status_and_jurisdiction
...
+            envelope("A.zip", "Y", Status.UPLOAD_FAILURE),
+            envelope("B.zip", "Z", Status.UPLOAD_FAILURE)
+        );
+
+        // when
+        List<Envelope> resultForA = repo.findByZipFileName("A.zip");
+        List<Envelope> resultForX = repo.findByZipFileName("X.zip");
+
+        // then
+        assertThat(resultForA).hasSize(2);
```
lgonczar: Maybe we could do a similar thing to what's in ProcessEventRepositoryTest, i.e. `containsExactly...`?

lgonczar:
In this PR or not, but still within this story we should think about including jurisdiction in file status. Using file name only leads to us handling undesired cases, such as multiple envelopes and events possibly mixed for multiple files (for different jurisdictions).

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepoTest.java
rkondratowicz: @lgonczar I've followed other tests in this class. There is one test that checks mapping 
Note that this new repo method does not include a custom SQL query, jpa magically builds a query based on the method name. So just checking the filtering part here.

doncem: @lgonczar we can't - we are fully aware of that. we can write for conceptual matter, but what you are referring (`containsExactly...`) is addressed in unit test. no need to double check type safe ecosystem. integration test is more of a "does this definitely return a list at all times" given the unit test covers the rest

PR #458, opened 2019-01-29
Replace save() with saveAndFlush() when using JPA repositories

lgonczar:
### Change description ###

When saving envelopes to DB, call `saveAndFlush()` method on JPA repository, instead of `save()`. This way we make sure the envelope is saved in the DB immediately and we don't leave any pending changes.

```
[ ] Yes
[x] No
```


lgonczar:
I guess we don't need to call `saveAndFlush` where we're immediately calling `getId` after that, but changed the call there for consistency. Can revert in those places, let me know what you think.

PR #463, opened 2019-01-30
Zip file status: move mapping to service

rkondratowicz:
to simplify


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusService.java
```diff
@@ -22,8 +26,29 @@ public ZipFileStatusService(ProcessEventRepository eventRepo, EnvelopeRepository
 
     public ZipFileStatus getStatusFor(String zipFileName) {
         return new ZipFileStatus(
-            envelopeRepo.findByZipFileName(zipFileName),
-            eventRepo.findByZipFileName(zipFileName)
+            envelopeRepo
```
lgonczar: It would be best to extract those transformations into separate methods, so that this one would read easier.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusService.java
```diff
@@ -22,8 +26,29 @@ public ZipFileStatusService(ProcessEventRepository eventRepo, EnvelopeRepository
...
+                .stream()
+                .map(event ->
+                    new ZipFileEvent(
+                        event.getEvent().name(),
+                        event.getContainer(),
+                        event.getCreatedAt()
+                    )
+                )
+                .collect(toList())
+
```
doncem: instead of evaluating the into the list and re-streaming again we could make repository return streams instead. so code would be similar to following:

```java
try (Stream stream = repo.findByCondition(value)) {
    return stream.map(...).collect(toList());
}
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusService.java
rkondratowicz: updated, let me know what you think

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/zipfilestatus/ZipFileStatusServiceTest.java
```diff
@@ -80,4 +110,20 @@ public void should_return_empty_lists_when_no_data_for_zip_file_was_found() {
         assertThat(result.envelopes).isNotNull().isEmpty();
         assertThat(result.events).isNotNull().isEmpty();
     }
+
+    private Envelope envelope(UUID id, String container, Status status) {
+        Envelope envelope = mock(Envelope.class);
+        given(envelope.getId()).willReturn(id);
```
rkondratowicz: using mocks becuase there is no setter for ID, plus we only use a couple of fields from this big model

PR #464, opened 2019-01-31
Log response from error response received via Feign Client

doncem:
### Change description ###

Response interceptors are not supported with feign library so need to handle them separately.
Since decoder did all the work and passed info to `ErrorNotificationException` "wrapper", we just need to log it in relevant place this occurs which is exception handler. Logging as separate entry to not get chunked in case massive response comes back. Logging as info as following logs are the actual piece of information in which we suppose to pay more attention

```
[ ] Yes
[x] No
```


doncem:
@nitinprabhu perhaps. Test could be a candidate of an upgrade to include wiremocked client similarly as [this one](https://github.com/hmcts/bulk-scan-processor/blob/master/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/client/ErrorNotificationClientTest.java). In the end it'll still be the same but enriched with automatic decoding. Would have to manually build feign exceptions with bodies etc.

Maybe ok to do as a separate PR to go through this (or more) test case?

PR #466, opened 2019-01-31
BPS-416 Use Instant class to represent date fields

aliveni:
https://tools.hmcts.net/jira/browse/BPS-416

### Change description ###
Replaced `Timestamp` class with `Instant` to represent date fields.

```
[ ] Yes
[x] No
```


src/integrationTest/resources/envelope.json
```diff
@@ -4,18 +4,18 @@
       "case_number": "1111222233334446",
       "po_box": "SSCSPO",
       "jurisdiction": "BULKSCAN",
-      "delivery_date": "2018-06-23T00:00:00.000Z",
-      "opening_date": "2018-06-23T00:00:00.000Z",
+      "delivery_date": "2018-06-23T02:10:30.123Z",
+      "opening_date": "2018-06-23T02:10:30.123Z",
```
rkondratowicz: why did those change? is this related to using a different class?

lgonczar: it's a better value than when the time component is zero, though

src/integrationTest/resources/envelope.json
```diff
@@ -4,18 +4,18 @@
...
       "scannable_items": [
         {
           "document_control_number": "1111002",
-          "scanning_date": "2018-06-23T00:00:00.000Z",
+          "scanning_date": "2018-06-23T02:10:30.123Z",
           "ocr_accuracy": "test",
           "manual_intervention": "test",
           "next_action": "return",
-          "next_action_date": "2018-06-23T00:00:00.000Z",
+          "next_action_date": "2018-06-24T00:00:00Z",
```
lgonczar: Is this the agreed format?

src/integrationTest/resources/zipcontents/mismatching_pdfs/metadata.json
```diff
@@ -22,11 +22,11 @@
       },
       {
         "document_control_number": "1111001",
-        "scanning_date": "2018-06-23T00:00:00.000Z",
+        "scanning_date": "2018-06-23T02:10:30.123Z",
         "ocr_accuracy": "test",
         "manual_intervention": "test",
         "next_action": "return",
-        "next_action_date": "2018-06-23T00:00:00.000Z",
+        "next_action_date": "2018-06-24T00:00:00.000Z",
```
lgonczar: Why the date change?

src/integrationTest/resources/envelope.json
rkondratowicz: let keep the format as it was, this story was supposed to be about the class used internally.

aliveni: `2018-06-23T00:00:00.000Z` the time values are all zeros here. `Instant` class ignores the milliseconds for such values

nitinprabhu: This is not agreed format `2018-06-23T02:10:30.123Z` should be agreed one. I think json schema validation will fail ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/CustomTimestampDeserialiser.java
```diff
@@ -6,26 +6,32 @@
...
+    public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
 
         String timestampString = p.getText();
 
         try {
-            return DateFormatter.getTimestamp(timestampString);
+            return LocalDateTime.parse(timestampString, DateTimeFormatter.ofPattern(DATETIME_PATTERN)).toInstant(UTC);
         } catch (DateTimeParseException exception) {
-            throw new InvalidTimestampFormatException(DateFormatter.getPattern(), exception);
+            throw new InvalidTimestampFormatException(DATETIME_PATTERN, exception);
```
lgonczar: We could change the name of the exception now that it's not timestamp.

src/integrationTest/resources/zipcontents/mismatching_pdfs/metadata.json
aliveni: it doesn't need to change, It can be same value. 
I just want to show one date field value where hour, minutes, seconds and milliseconds are zeros.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/CustomTimestampDeserialiser.java
aliveni: `InvalidISODateFormatException`?

src/integrationTest/resources/envelope.json
aliveni: @nitinprabhu I see the same format in all test files and the validation is not failing

src/integrationTest/resources/zipcontents/mismatching_pdfs/metadata.json
rkondratowicz: it was zeros. day changed from 23 to 24

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/CustomTimestampDeserialiser.java
lgonczar: Or even just `InvalidDateFormatException`. Up to you. The name of the exception can be generic enough to say "this date is not in the format that we find valid".

src/integrationTest/resources/envelope.json
nitinprabhu: If you check the json schema it is ```"pattern": "^\\d{4}\\-\\d{2}\\-\\d{2}T\\d{2}\\:\\d{2}\\:\\d{2}\\.\\d{3}Z$"``` so 3 digits for seconds not sure why the validation doesn't fails ?

nitinprabhu: Looks like we will need to add more tests because I am sure when we did end to end tests that format won't work

aliveni: This is not metadata fail. this is the result of parsed envelope, which will be put on the service bus queue

nitinprabhu: it works even if the values are 0. So `2018-06-23T00:00:00.000Z` works

aliveni: yes it works. please see the metadata files used for the test.
src/integrationTest/resources/zipcontents/mismatching_pdfs/metadata.json
src/integrationTest/resources/zipcontents/ok/metadata.json

vakepati: I just uploaded a zip file with date format as  "2018-12-27T10:46:03.12Z",but it has thrown schema validation error like below.
"Failed validation for file 65755555551369_06-12-2018-07-22-36.zip against schema. Errors:
 error: ECMA 262 regex "^\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}Z$" does not match input string "2018-12-27T10:46:03.12Z" level: "error" schema: {"loadingURI":"#","pointer":"/properties/delivery_date"} instance: {"pointer":"/delivery_date"} domain: "validation" keyword: "pattern" regex: "^\\d{4}\\-\\d{2}\\-\\d{2}T\\d{2}\\:\\d{2}\\:\\d{2}\\.\\d{3}Z$" string: "2018-12-27T10:46:03.12Z" error: ECMA 262 regex "^\d{4}\-\d{2}\-\d{2}T\d{2}\:\d{2}\:\d{2}\.\d{3}Z$" does not match input string "2018-12-27T10:46:03.12Z" level: "error" schema: "

rkondratowicz: if it works with what it was, let's not change it

rkondratowicz: `2018-12-27T10:46:03.12Z` does not follow any of the formats discussed above, @vakepati

vakepati: > does `2018-06-23T00:00:00.000Z` fail?
Passed!

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/CustomTimestampDeserialiser.java
```diff
@@ -3,29 +3,29 @@
...
-import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidTimestampFormatException;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidDateFormatException;
 
 import java.io.IOException;
-import java.sql.Timestamp;
+import java.time.Instant;
 import java.time.format.DateTimeParseException;
 
-public class CustomTimestampDeserialiser extends StdDeserializer<Timestamp> {
+public class CustomTimestampDeserialiser extends StdDeserializer<Instant> {
```
lgonczar: Worth changing the name to sth like `CustomInstantDeserialiser`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/util/CustomTimestampSerialiser.java
```diff
@@ -5,29 +5,16 @@
...
 import java.io.IOException;
-import java.sql.Timestamp;
 import java.time.Instant;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
 
-import static java.time.ZoneOffset.UTC;
-
-public class CustomTimestampSerialiser extends StdSerializer<Timestamp> {
+public class CustomTimestampSerialiser extends StdSerializer<Instant> {
```
lgonczar: Like with the deserialiser, worth changing the name to sth like `CustomInstantSerialiser` or even `InstantSerialiser`, as the very fact of having it in our code implies that it's custom.

lgonczar:
@rkondratowicz I looks like the changes you requested have been made

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/EnvelopeResponse.java
```diff
@@ -68,12 +66,9 @@ public EnvelopeResponse(
         @JsonProperty("container") String container,
         @JsonProperty("po_box") String poBox,
         @JsonProperty("jurisdiction") String jurisdiction,
-        @JsonDeserialize(using = CustomTimestampDeserialiser.class)
-        @JsonProperty("delivery_date") Timestamp deliveryDate,
-        @JsonDeserialize(using = CustomTimestampDeserialiser.class)
-        @JsonProperty("opening_date") Timestamp openingDate,
-        @JsonDeserialize(using = CustomTimestampDeserialiser.class)
```
rkondratowicz: what happened to those?

aliveni: They are working without the deserializer. Anyway, I added them back

PR #467, opened 2019-01-31
Introduce request interceptor for error notification client

doncem:
### Change description ###

Just _lambda_ injecting the interceptor in feign client configuration. Feel free to suggest any other way because to my eyes having `ErrorNotificationConfiguration` logger was not really a spot on.

Lot's of refurbishment in client test to track request body in "nicer" way

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/client/ErrorNotificationConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/client/ErrorNotificationConfiguration.java
@@ -1,10 +1,13 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import feign.RequestInterceptor;
 import feign.auth.BasicAuthRequestInterceptor;
 import feign.codec.Decoder;
 import feign.codec.ErrorDecoder;
 import feign.jackson.JacksonDecoder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/client/ErrorNotificationConfiguration.java
```diff
@@ -25,6 +28,16 @@ public BasicAuthRequestInterceptor basicAuthRequestInterceptor(
         return new BasicAuthRequestInterceptor(username, password, StandardCharsets.UTF_8);
     }
 
+    @Bean
+    public RequestInterceptor logRequestBodyInterceptor() {
```
lgonczar: I don't think we can log request body, as there's a risk that it will contain sensitive information (OCR data).

We should log response body instead.

doncem: response is logged #464 

furthermore, if such comment exists and thumbed up by original requestor - there's something missing in between

not minding the above, can this PR be closed as not relevant?

nitinprabhu: We realised this when we having a discussion that ocr data also has pattern validation so there is a chance that if validation fails then ocr data will be logged. Hence the request

lgonczar: > response is logged #464
👍

doncem:
due to the nature of data being exposed this has to be either completely ignored or reviewed with censorship on

PR #469, opened 2019-02-01
Remove unused repository methods

lgonczar:

### Change description ###

Remove repository methods created in repositories in order to finalise processed messages without having to read envelopes. Finalising an envelope involves creating an event, which requires more information than comes from the queue. Therefore, these methods won't be used.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/ScannableItemRepository.java
```diff
@@ -1,14 +1,9 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.entity;
 
 import org.springframework.data.jpa.repository.JpaRepository;
-import org.springframework.data.jpa.repository.Modifying;
-import org.springframework.data.jpa.repository.Query;
-import org.springframework.data.repository.query.Param;
 
 import java.util.UUID;
 
 public interface ScannableItemRepository extends JpaRepository<ScannableItem, UUID> {
```
doncem: can we remove this repository entirely?

lgonczar: This repository is used in DocumentProcessor. I'm removing a method I've recently added, because it won't be used.

PR #476, opened 2019-02-05
Update base docker image

rkondratowicz:
The old one got removed O_o


rkondratowicz:
There are other issues later in the pipeline but this gets the build past the 'Container build' stage.

PR #478, opened 2019-02-06
Create a class for handling 'processed-envelopes' queue messages

lgonczar:

### Change description ###

Create a class for handling 'processed-envelopes' queue messages. This class is not plugged in, yet, so it will not process any messages.

Next:
- providing the necessary configuration
- registering the handler (includes functional tests)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
@@ -14,9 +14,16 @@ import uk.gov.hmcts.reform.bulkscanprocessor.model.in.msg.ProcessedEnvelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.EnvelopeFinaliserService;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.MessageAutoCompletor;
 
-import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
+/**
+ * Handler of messages form processed envelopes queue.
+ *
+ * <p>
+ *   Its purpose is to bring envelopes referenced by those messages to their final state.
+ *   This involves removing sensitive information, status change and creation of an appropriate event.
+ * </p>
+ */
 public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
 
     private static final Logger log = LoggerFactory.getLogger(ProcessedEnvelopeNotificationHandler.class);
@@ -126,13 +133,11 @@ public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
         }
     }
 
-    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws IOException {
+    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws Exception {
         try {
             return objectMapper.readValue(message.getBody(), ProcessedEnvelope.class);
         } catch (JsonParseException | JsonMappingException e) {
             throw new InvalidMessageException("Failed to parse 'processed envelope' message", e);
-        } catch (IOException e) {
-            throw e;
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
@@ -14,6 +14,7 @@ import uk.gov.hmcts.reform.bulkscanprocessor.model.in.msg.ProcessedEnvelope;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.EnvelopeFinaliserService;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.MessageAutoCompletor;
 
+import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -133,7 +134,7 @@ public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
         }
     }
 
-    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws Exception {
+    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws IOException {
         try {
             return objectMapper.readValue(message.getBody(), ProcessedEnvelope.class);
         } catch (JsonParseException | JsonMappingException e) {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
@@ -16,13 +16,14 @@ import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.MessageAutoComp
 
 import java.io.IOException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * Handler of messages form processed envelopes queue.
- *
  * <p>
- *   Its purpose is to bring envelopes referenced by those messages to their final state.
- *   This involves removing sensitive information, status change and creation of an appropriate event.
+ * Its purpose is to bring envelopes referenced by those messages to their final state.
+ * This involves removing sensitive information, status change and creation of an appropriate event.
  * </p>
  */
 public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
@@ -32,6 +33,7 @@ public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
     private final EnvelopeFinaliserService envelopeFinaliserService;
     private final ObjectMapper objectMapper;
     private final MessageAutoCompletor messageCompletor;
+    private ExecutorService executor = Executors.newFixedThreadPool(4);
 
     public ProcessedEnvelopeNotificationHandler(
         EnvelopeFinaliserService envelopeFinaliserService,
@@ -46,8 +48,8 @@ public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
     @Override
     public CompletableFuture<Void> onMessageAsync(IMessage message) {
         return CompletableFuture
-            .supplyAsync(() -> tryProcessMessage(message))
-            .thenAcceptAsync(processingResult -> tryFinaliseMessage(message, processingResult))
+            .supplyAsync(() -> tryProcessMessage(message), executor)
+            .thenComposeAsync(processingResult -> tryFinaliseMessageAync(message, processingResult), executor)
             .handleAsync((v, error) -> {
                 if (error != null) {
                     log.error(
@@ -69,39 +71,44 @@ public class ProcessedEnvelopeNotificationHandler implements IMessageHandler {
         );
     }
 
-    private void tryFinaliseMessage(IMessage message, MessageProcessingResult processingResult) {
-        try {
-            finaliseMessage(message, processingResult);
-        } catch (Exception e) {
-            log.error(
-                "An error occurred when trying to finalise 'processed envelope' message with ID {}",
-                message.getMessageId(),
-                e
-            );
-        }
+    private CompletableFuture<Void> tryFinaliseMessageAync(
+        IMessage message,
+        MessageProcessingResult processingResult
+    ) {
+        return finaliseMessageAsync(message, processingResult)
+            .exceptionally(error -> {
+                log.error(
+                    "An error occurred when trying to finalise 'processed envelope' message with ID {}",
+                    message.getMessageId(),
+                    error
+                );
+
+                return null;
+            });
     }
 
-    private void finaliseMessage(IMessage message, MessageProcessingResult processingResult) {
+    private CompletableFuture<Void> finaliseMessageAsync(IMessage message, MessageProcessingResult processingResult) {
         switch (processingResult.resultType) {
             case SUCCESS:
-                messageCompletor.completeAsync(message.getLockToken()).join();
-                log.info("Completed 'processed-envelope' message with ID {}", message.getMessageId());
-                break;
+                return messageCompletor.completeAsync(message.getLockToken()).thenRun(() ->
+                    log.info("Completed 'processed-envelope' message with ID {}", message.getMessageId())
+                );
             case UNRECOVERABLE_FAILURE:
-                messageCompletor.deadLetterAsync(
+                return messageCompletor.deadLetterAsync(
                     message.getLockToken(),
                     "Message processing error",
                     processingResult.exception.getMessage()
-                ).join();
-
-                log.info("Dead-lettered 'processed-envelope' message with ID {}", message.getMessageId());
-                break;
+                ).thenRun(() ->
+                    log.info("Dead-lettered 'processed-envelope' message with ID {}", message.getMessageId())
+                );
             default:
                 log.info(
                     "Letting 'processed envelope' message with ID {} return to the queue. Delivery attempt {}.",
                     message.getMessageId(),
                     message.getDeliveryCount() + 1
                 );
+
+                return CompletableFuture.completedFuture(null);
         }
     }
 
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
@@ -1,0 +1,158 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
+
+import com.fasterxml.jackson.core.JsonParseException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+                message.getMessageId(),
+                e
+            );
+        }
+    }
+
+    private void finaliseMessage(IMessage message, MessageProcessingResult processingResult) {
+        switch (processingResult.resultType) {
+            case SUCCESS:
+                messageCompletor.completeAsync(message.getLockToken()).join();
```
lgonczar: We're already in a future

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws IOException {
+        try {
+            return objectMapper.readValue(message.getBody(), ProcessedEnvelope.class);
+        } catch (JsonParseException | JsonMappingException e) {
+            throw new InvalidMessageException("Failed to parse 'processed envelope' message", e);
+        } catch (IOException e) {
+            throw e;
+        }
```
rkondratowicz: last `catch` not needed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+    ) {
+        this.envelopeFinaliserService = envelopeFinaliserService;
+        this.objectMapper = objectMapper;
+        this.messageCompletor = messageCompletor;
+    }
+
+    @Override
+    public CompletableFuture<Void> onMessageAsync(IMessage message) {
+        return CompletableFuture
+            .supplyAsync(() -> tryProcessMessage(message))
```
doncem: hmm. so this handler tries to exec everything in one go. this is not optimising the feature completable future provides. the way i see it, for instance, this can be extracted step by step making the implementation both easier to read and broken down to relevant stages

```java
// taken from notification one
CompletableFuture
    .supplyAsync(message::getBody, SIMPLE_EXEC)
    .thenApplyAsync(body -> readProcessedEnvelope(body, message.getMessageId(), message.getDeliveryCount()), SERVICE_EXEC)
    .thenApplyAsync(processedEnvelope -> processEnvelope(processedEnvelope, message.getMessageId(), SERVICE_EXEC)
    .thenAcceptAsync(result -> finaliseResult(result, message), SIMPLE_EXEC)
    .handleAsync((voided, throwable) -> { w/e (return future) }, SIMPLE_EXEC)
    .thenComposeAsync(Functions.identity(), ERROR_EXEC)
    .thenRunAsync(() -> log.debug("Error notification consumed. ID {}", message.getMessageId()), SIMPLE_EXEC);
```

each line would represent one piece of action. can pass whole message if need to

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            .thenAcceptAsync(processingResult -> tryFinaliseMessage(message, processingResult))
+            .handleAsync((v, error) -> {
+                if (error != null) {
+                    log.error(
+                        "An error occurred when trying to handle 'processed envelope' message with ID {}",
+                        message.getMessageId()
+                    );
+                }
+
+                return null;
```
doncem: i don't think `return null` does the trick. i could be confusing with something else

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            "An error occurred when handling processed envelope notification. Phase: {}",
+            exceptionPhase,
+            throwable
+        );
+    }
+
+    private void tryFinaliseMessage(IMessage message, MessageProcessingResult processingResult) {
+        try {
+            finaliseMessage(message, processingResult);
+        } catch (Exception e) {
```
doncem: if you are catching the exception here - there's no need to handle in the future composition in main method above.

this whole method is pointless really. future is there to handle such misbehaviour

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
doncem: we should return future and flatmap them

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            return new MessageProcessingResult(MessageProcessingResultType.POTENTIALLY_RECOVERABLE_FAILURE);
+        }
+    }
+
+    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws IOException {
+        try {
+            return objectMapper.readValue(message.getBody(), ProcessedEnvelope.class);
+        } catch (JsonParseException | JsonMappingException e) {
+            throw new InvalidMessageException("Failed to parse 'processed envelope' message", e);
+        } catch (IOException e) {
```
doncem: no need to catch this

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            );
+
+            return new MessageProcessingResult(MessageProcessingResultType.POTENTIALLY_RECOVERABLE_FAILURE);
+        }
+    }
+
+    private ProcessedEnvelope readProcessedEnvelope(IMessage message) throws IOException {
+        try {
+            return objectMapper.readValue(message.getBody(), ProcessedEnvelope.class);
+        } catch (JsonParseException | JsonMappingException e) {
```
doncem: catching `JsonMappingException` is quite wide iirc. again, i could be mixing up

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -32,6 +33,7 @@
     private final EnvelopeFinaliserService envelopeFinaliserService;
     private final ObjectMapper objectMapper;
     private final MessageAutoCompletor messageCompletor;
+    private ExecutorService executor = Executors.newFixedThreadPool(4);
```
doncem: perhaps `static final` ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: There are actually two steps:
- process the envelope, i.e. do all the db updates
- finalise the message in the queue (complete/dead-letter)

These are the only ones that take any meaningful amount of time and there's no benefit of splitting them further.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: There's no instantiation of Void

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -46,8 +48,8 @@ public ProcessedEnvelopeNotificationHandler(
     @Override
     public CompletableFuture<Void> onMessageAsync(IMessage message) {
         return CompletableFuture
-            .supplyAsync(() -> tryProcessMessage(message))
-            .thenAcceptAsync(processingResult -> tryFinaliseMessage(message, processingResult))
+            .supplyAsync(() -> tryProcessMessage(message), executor)
```
doncem: still want to go with "full blown" `tryProcess`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            .handleAsync((v, error) -> {
+                if (error != null) {
+                    log.error(
+                        "An error occurred when trying to handle 'processed envelope' message with ID {}",
+                        message.getMessageId()
+                    );
+                }
+
+                return null;
+            });
```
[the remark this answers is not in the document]
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
doncem: p.s. people where concerned about having more than 1 thread and iirc there is a ticket to investigate best/better approach for that. not the issue here but worth mentioning so we are all aware

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: The one in the main method catches anything that wasn't handled in any of the steps. There shouldn't be any such thing, but if it happens, we'd like to see it in logs. This catch, on the other hand (I've just changed it slightly, though) allows for providing an error message specific to this stage.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -69,39 +71,44 @@ public void notifyException(Throwable throwable, ExceptionPhase exceptionPhase)
...
-    private void finaliseMessage(IMessage message, MessageProcessingResult processingResult) {
+    private CompletableFuture<Void> finaliseMessageAsync(IMessage message, MessageProcessingResult processingResult) {
         switch (processingResult.resultType) {
             case SUCCESS:
-                messageCompletor.completeAsync(message.getLockToken()).join();
-                log.info("Completed 'processed-envelope' message with ID {}", message.getMessageId());
-                break;
+                return messageCompletor.completeAsync(message.getLockToken()).thenRun(() ->
+                    log.info("Completed 'processed-envelope' message with ID {}", message.getMessageId())
+                );
```
doncem: just a design suggestion:
```
return messageCompletor
    .completeAsync(...)
    .thenRun(...)
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: removed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
```diff
@@ -69,39 +71,44 @@ public void notifyException(Throwable throwable, ExceptionPhase exceptionPhase)
...
+        return finaliseMessageAsync(message, processingResult)
+            .exceptionally(error -> {
+                log.error(
+                    "An error occurred when trying to finalise 'processed envelope' message with ID {}",
+                    message.getMessageId(),
+                    error
+                );
+
+                return null;
+            });
```
doncem: not sure this falls into `handleAsync` in main method. do you have integration test for that?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
doncem: one can argue but nevermind. as of wider scope you are right 👍

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: I'll start with one thread then

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: This won't fall into `handleAsync` in main method. The error won't be propagated further, which is what we want.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: There's a lot of error handling logic that in my opinion would be best not to expose at this level.

lgonczar: I've added a comment in handleAsync to make it clear what purpose it serves there

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
doncem: then why the need of a `handelAync`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
doncem: which is why single point of entry to deal with exceptions is the exception handler. where you can catch everything singularly despite how origin works. so, in short, processing can be left alone with just doing what it suppose to do. any exception? - handle in exception handler

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: I'm being extra cautious. In the past we've had situations where exceptions were supposed to be handled at a lower level, but it turned out that some error cases weren't covered, which resulted with failures with no logs. All you could see was that it was going fine, but never completed.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessedEnvelopeNotificationHandler.java
lgonczar: I completely disagree. It matters which exception is thrown at which point. We have to catch exceptions in the first step, in order to provide the second step with the right result (so that it knows what to do with the message). And when we fail to do some specific thing, it's much better to see an error saying we've failed to do just that (with details that are available in this context) than that the whole task failed.

When getting rid of context by centralising exception handling we're likely to overcomplicate our logic. Single exception handlers aren't always the right thing to do, unless they handle some general failures, just like the one in the main method.

PR #479, opened 2019-02-07
Refactor Service Bus-specific Spring configuration

lgonczar:

### Change description ###

Refactor Service Bus-specific Spring configuration, so that more message handlers can be added.

In particular, `ProcessedEnvelopeNotificationHandler` needs to have its dependencies injected in the constructor (one of them has `@Transactional` annotation on its method) and a queue client is one of those dependencies.

```
[ ] Yes
[x] No
```


src/integrationTest/resources/application.properties
```diff
@@ -18,3 +18,6 @@ storage.blob_lease_timeout=60
 servicebus.queue_envelope_send=Hostname=bulkscanning.servicebus.windows.net;EntityPath=envelopes
 
 user.timezone=GMT
+
+# no communication with Service Bus
+spring.profiles.active=nosb
```
lgonczar: `nosb` is a profile name used in the orchestrator, so I used it for consistency

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/QueueClientConfig.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Bean("notifications")
+    public IQueueClient notificationsQueueClient(
+        @Value("${queues.notifications.connection-string}") String connectionString,
+        @Value("${queues.notifications.queue-name}") String queueName
+    ) throws InterruptedException, ServiceBusException {
+        return createQueueClient(connectionString, queueName);
+    }
+
+    @Bean("read-notifications")
```
rkondratowicz: So we have `notifications` and `read-notifications`.
Is the first on for both reading and writing. Or just writing?
Worth clarifying imho.

lgonczar: I was following the existing convention - notifications vs read-notifications. Probably best to leave it as it is and change everything in a separate PR to let this one address one thing.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
```diff
@@ -1,98 +1,47 @@
...
-    @Value("${queues.read-notifications.enabled}")
-    private boolean readNotificationsEnabled;
 
-    @Value("${queues.read-notifications.connection-string}")
-    private String readNotificationConnectionString;
-
-    @Value("${queues.read-notifications.queue-name}")
-    private String readNotificationQueueName;
-
-    @Bean(name = "envelopes")
```
doncem: missing `envelopes-helper` bean name - it is used in `OrchestratorNotificationTask`

lgonczar: It is right in the line below.

Helper aside, maybe I should change the bean name for queue client to `envelopes-client` or sth?

doncem: ah, must've crossed eyes and missed it

since you sifting around servicebus configs, worth naming/renaming those clients too ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
```diff
@@ -1,98 +1,47 @@
...
 import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.MessageAutoCompletor;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus.ServiceBusHelper;
-import uk.gov.hmcts.reform.bulkscanprocessor.tasks.ErrorNotificationHandler;
-
-import java.time.Duration;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import javax.annotation.PostConstruct;
 
 @Lazy
```
doncem: I think `@Lazy` was so clients won't get initialised until called. Since connection string (evaluations) has been moved to `QueueClientConfig`, perhaps this annotation has to be moved there too? Or left here but added to `QueueClientConfig` as well

src/main/resources/application.yaml
```diff
@@ -12,7 +12,7 @@ spring:
   datasource:
     driver-class-name: org.postgresql.Driver
     url: jdbc:postgresql://${BULK_SCANNING_DB_HOST:localhost}:${BULK_SCANNING_DB_PORT:5435}/${BULK_SCANNING_DB_NAME:bulkscans}${BULK_SCANNING_DB_CONN_OPTIONS:}
-    username: ${BULK_SCANNING_DB_USER_NAME:bulkscans}
+    username: ${BULK_SCANNING_DB_USER_NAME:bulkscan}
```
doncem: 😳

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
lgonczar: Or maybe we should just remove it from here? The creation of QueueClientConfig's beans depends on the active Spring profile.

src/main/resources/application.yaml
lgonczar: Gone. Accidents...

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/ServiceBusConfiguration.java
doncem: I suppose this one doesn't really matter since `QueueClientConfig` will be evaluated on boot so yes - can be removed ¯\\\_(ツ)_/¯

lgonczar:
> Just final not on `@Lazy`
> 
> P.S. how about introducing custom annotations to replace the following:
> 
> ```java
> @Configuration
> @Profile("!nosb") // no servicebus queue handler registration
> ```
> 
> e.g. `@ServiveBusConfiguration`

Not a bad idea. Will do.

lgonczar:
applied the comments

lgonczar:
Codecov is complaining about decreased coverage, but the only changes made in this PR are configuration files, annotations and config classes. We should probably add exclusions for the last ones, just like we do in sonar. Other than that, everything is good.

PR #486, opened 2019-02-12
Handle empty zip file names

rkondratowicz:
AAT gets flooded with errors caused by attempt of processing zip files when passed file name is null or empty.
>The argument must not be null or an empty string. Argument name: blobName.
>com.microsoft.azure.storage.blob.CloudBlobContainer.getBlockBlobReference

In this PR such files are skipped and their URIs are logged.

Investigation to follow based on those logs...

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -4,6 +4,7 @@ import com.microsoft.azure.storage.StorageException;
 import com.microsoft.azure.storage.blob.BlobInputStream;
 import com.microsoft.azure.storage.blob.CloudBlobContainer;
 import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import joptsimple.internal.Strings;
 import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -125,9 +126,16 @@ private void processZipFiles(CloudBlobContainer container) {
...
         // implements RandomAccess (e.g. ArrayList)
         List<String> zipFilenames = new ArrayList<>();
-        container.listBlobs().forEach(
-            b -> zipFilenames.add(FilenameUtils.getName(b.getUri().toString()))
-        );
+        container
+            .listBlobs()
+            .forEach(b -> {
+                String fileName = FilenameUtils.getName(b.getUri().toString());
+                if (Strings.isNullOrEmpty(fileName)) {
```
lgonczar: I wonder if we'll even get here. Won't `b.getUri()` be null?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -125,9 +126,16 @@ private void processZipFiles(CloudBlobContainer container) {
...
+                    log.warn("Cannot extract filename from list blob item. URI: " + b.getUri().toString());
```
lgonczar: Maybe it's best to fail here?

PR #488, opened 2019-02-12
Extract base functional test class

lgonczar:
### Change description ###

Create a base class for functional tests, in order to remove duplication. There's a lot of setup-related code duplication among functional tests and after adding another test (for handling processed-envelopes queue) there would be even more. This PR is here to extract some common logic into a base class.

This is a proposal. Alternative ideas are welcome.

```
[ ] Yes
[x] No
```


lgonczar:
Composition-based approach was the first one I thought about, but it needs a bit more work, including modifying the already existing TestHelper class to take some of the responsibilities (so that it wouldn't need test classes to do the whole setup for it) or even splitting it into blob storage-specific tool and an API client. This PR is the first step, which removes code duplication (and lets us avoid even more duplication in tests to come).

PR #492, opened 2019-02-14
Handle invalid zip files

rkondratowicz:
Move file to rejected container.
Previously file was being left where it was causing the job to process it on next run (and creating error event. over and over again).

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
@@ -3,6 +3,7 @@ package uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor;
 import com.google.common.base.Strings;
 import com.google.common.io.Resources;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.DocSignatureFailureException;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidZipArchiveException;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -107,7 +108,7 @@ static ZipInputStream sha256WithRsaVerification(ZipStreamWithSignature zipWithSi
 
             return zipEntries;
         } catch (IOException ioe) {
-            throw new SignatureValidationException(ioe);
+            throw new InvalidZipArchiveException("Error extracting zip entries", ioe);
```
rkondratowicz: nothing to do with signatures, hence new exception.
It extends `InvalidEnvelopeException` which is handled and results in moving file to rejected container.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/ZipVerifiers.java
```diff
@@ -107,7 +108,7 @@ static ZipInputStream sha256WithRsaVerification(ZipStreamWithSignature zipWithSi
 
             return zipEntries;
         } catch (IOException ioe) {
-            throw new SignatureValidationException(ioe);
```
doncem: Has `SignatureValidationException` or its construct became retired?

rkondratowicz: still used in 5 other places in code.

rkondratowicz:
closing, will create a new PR to start with a new AKS pipeline

PR #515, opened 2019-02-20
Handle invalid zip files

rkondratowicz:
Recreating https://github.com/hmcts/bulk-scan-processor/pull/492 due to pipeline issues

>Move file to rejected container.
>Previously file was being left where it was causing the job to process it on next run (and creating error event. over and over again).


rkondratowicz:
Added integration test

PR #526, opened 2019-02-25
Exclude config classes from line coverage in Sonar

lgonczar:
### Change description ###

Exclude config classes from line coverage in Sonar

```
[ ] Yes
[x] No
```


lgonczar:
PR builds currently don't work. Given master is broken and this change only adds an exclusion for sonar, which should fix master, I'm merging this one.

PR #529, opened 2019-02-25
BPS-514 Service layer changes

aliveni:
https://tools.hmcts.net/jira/browse/BPS-514

### Change description ###
Renamed response model.
Updated ReportsService to convert the data to ZipFileSummaryResponse.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
@@ -1,15 +1,23 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.services.reports;
 
-import org.apache.commons.lang3.NotImplementedException;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReportsService.java
```diff
@@ -37,8 +49,24 @@ public ReportsService(
             .collect(toList());
     }
 
-    public List<ZipFileSummary> getZipFilesSummary(LocalDate date, String jurisdiction) {
-        throw new NotImplementedException("Not yet implemented");
+    public List<ZipFileSummaryResponse> getZipFilesSummary(LocalDate date, String jurisdiction) {
```
rkondratowicz: It would be good to either add javadoc explaining that null or empty `jurisdiction` does not filter or add another method:
```
List<ZipFileSummaryResponse> getZipFilesSummary(LocalDate date)
```

aliveni: @rkondratowicz Added Javadoc

PR #530, opened 2019-02-25
Fix PMD issues

rkondratowicz:
some of them...

### DIFF
--- a/config/pmd/ruleset.xml
+++ b/config/pmd/ruleset.xml
@@ -17,6 +17,7 @@
     <exclude name="MethodArgumentCouldBeFinal" />
     <exclude name="OnlyOneReturn" />
     <exclude name="TooManyStaticImports" />
+    <exclude name="ConfusingTernary" />
   </rule>
   <rule ref="category/java/codestyle.xml/ClassNamingConventions">
     <properties>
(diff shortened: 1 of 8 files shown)


config/pmd/ruleset.xml
```diff
@@ -17,6 +17,7 @@
     <exclude name="MethodArgumentCouldBeFinal" />
     <exclude name="OnlyOneReturn" />
     <exclude name="TooManyStaticImports" />
+    <exclude name="ConfusingTernary" />
```
rkondratowicz: to allow `if(x != y) { ... }`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/AuthServiceConfig.java
```diff
@@ -39,14 +39,17 @@ public AuthTokenValidator tokenValidator(ServiceAuthorisationApi s2sApi) {
...
             }
 
+            @Override
             public void validate(String token, List<String> roles) {
                 throw new NotImplementedException();
             }
 
+            @Override
             public String getServiceName(String token) {
                 return "some_service_name";
```
aliveni: Why are we overriding the above methods?

rkondratowicz: methods were always overridden, just adding annotation to make it clear (as suggested by PMD)

aliveni: oh, got it. didn't notice the interface earlier

PR #532, opened 2019-02-26
Create 'ZIPFILE_PROCESSING_STARTED' event for old files.

lgonczar:
### Change description ###

Create a migration script adding `ZIPFILE_PROCESSING_STARTED` event for old files, i.e. files processed before the service started creating this kind of event.

The presence of this event for all files will help create reports for dates when this event wasn't created by the service.

```
[ ] Yes
[x] No
```


lgonczar:
> I was wondering should we also add date in the where clause before which(15th) we want this to be applied

That would be tricky, as during the transition some instances could create the event while others weren't updated, yet.

This change creates the event only if it doesn't exist for a given file in a given container.

PR #536, opened 2019-02-27
Clean up csv report generation

rkondratowicz:
- use single url (use Accept header to return data in appropriate format)
- return csv in response body instead of downloading a file


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ReportsController.java
```diff
@@ -79,21 +77,19 @@ public ZipFilesSummaryReportListResponse getZipFilesSummary(
         );
     }
 
-    @GetMapping(path = "/zip-files-summary-csv", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
+    @GetMapping(path = "/zip-files-summary", produces = "text/csv")
```
aliveni: "text/csv" will just return the comma separated values. I think we want to download the file in csv format

PR #539, opened 2019-03-01
Add indexes to process_events

aliveni:
https://tools.hmcts.net/jira/browse/BPS-514

### Change description ###
Added indexes for the columns in `process_events` table,  
which are used for the zip files summary query

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/main/resources/db/migration/V038__Add_indexes_to_process_events.sql
+++ b/src/main/resources/db/migration/V038__Add_indexes_to_process_events.sql
@@ -1,0 +1,2 @@
+CREATE INDEX process_events_container_zip_event_idx ON process_events (container, zipfilename, event);
+CREATE INDEX process_events_event_createdat_idx ON process_events (event, createdat);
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V038__Add_indexes_to_process_events.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX process_events_container_zip_event_idx ON process_events (container, zipfilename, event);
```
lgonczar: There already exists an index on container + zipfilename. If we need this one, the other one should be dropped.

aliveni: @lgonczar I think we still need the existing index for container and zipfilename. 
part of the reports query fetches the data on `container and zipfilename`

lgonczar: This new index can still be used by PostgreSQL to search by container and zipfilename, as they're the first two columns in the index.

aliveni: ok, I'll update the script to drop the existing column

PR #554, opened 2019-03-08
Add service for deleting old rejected files

rkondratowicz:
This PR only adds the service and tests for it.
It is not used by the app and no scheduling is configured (to come in next PR(s))

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
@@ -60,13 +60,13 @@ public class CleanUpRejectedFilesTaskTest {
     @Test
     public void should_delete_old_files() throws Exception {
         // given
-        upload("foo.zip");
-        upload("bar.zip");
+        // there are two files in rejected container
+        rejectedContainer.getBlockBlobReference("foo.zip").uploadText("some content");
+        rejectedContainer.getBlockBlobReference("bar.zip").uploadText("some content");
 
-        // sanity check
-        assertThat(rejectedContainer.listBlobs()).hasSize(2);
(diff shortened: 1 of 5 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
```diff
@@ -0,0 +1,88 @@
...
+        this.rejectedContainer = cloudBlobClient.getContainerReference("test-rejected");
+        this.rejectedContainer.createIfNotExists();
+    }
+
+    @Test
+    public void should_delete_old_files() throws Exception {
+        // given
+        Duration deleteDelay = Duration.ZERO; // delete immediately
+
+        upload("foo.zip");
```
nitinprabhu: this is upload to rejected container. may be rename it as uploadToRejectedContainer ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -0,0 +1,73 @@
...
+                container
+                    .listBlobs(null, true, EnumSet.of(SNAPSHOTS), null, null)
+                    .forEach(listItem -> {
+                        try {
+                            CloudBlockBlob blob = container.getBlockBlobReference(
+                                FilenameUtils.getName(listItem.getUri().toString())
+                            );
+
+                            if (canBeDeleted(blob)) {
+                                blob.delete(INCLUDE_SNAPSHOTS, null, null, null);
```
nitinprabhu: what would happen if two processes pick up the file for deletion ? Are we going to have some scheduler lock so that only one process will run at a time ?

rkondratowicz: Yes, shedlock will be used for that (as for other scheduled jobs).
Frequency still TBD.

PR #571, opened 2019-03-19
Make the processor crash on failed message handler registration

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-445

### Change description ###

Make the processor crash on failed message handler registration. Before this change, it would catch the exception and log the failure on preview, causing functional tests within the same build to fail.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/MessageHandlerConfig.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/MessageHandlerConfig.java
@@ -7,7 +7,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/MessageHandlerConfig.java
```diff
@@ -53,36 +49,18 @@
...
-            // The application has to keep working on Preview - otherwise the pipeline
-            // wouldn't create queues which it relies on.
-            // The problem will be addresses in BPS-445
-            log.error("An error occurred when trying to register message handlers", cause);
-        }
+        processedEnvelopesQueueClient.registerMessageHandler(
+            processedEnvelopeNotificationHandler,
+            messageHandlerOptions,
+            processedEnvelopesReadExecutor
+        );
```
doncem: should this block be treated same as the one above? (if condition wrap)

lgonczar: The condition is about reading from `notifications` queue. I thought it had some business meaning, e.g. we may want to turn it off or sth. If that's for the purpose of making the behaviour environment-specific, I'd say we should remove it and always register the handler.

PR #568, opened 2019-03-14
Update gradle

rkondratowicz:
4.10.2 -> 5.2.1 😱

PR #574, opened 2019-03-19
Update spring dependencies and alike

doncem:
### Change description ###

Project itself is incorrectly configured giving the breaking change of bean override rule enforced from spring boot 2.1. For this PR we just allow it to happen via simple configuration setting.

There are few bonuses dragging along: fix flyway plugin, gradle upgrade leftovers from #568, owasp.. Follow commits and diff - it should not be too complicated.

Main issue was a jump from v3 to v4 of java logging library. v4 to v5 jump is not really ground breaking. Because of the former, tests which are using `OutputCapture` rule are not trustworthy - removed the rule

```
[ ] Yes
[x] No
```


doncem:
¯\\\_(ツ)_/¯

PR #582, opened 2019-03-22
Fix rejected files cleanup job

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-595

### Change description ###

Fix the parsing of rejected files' TTL and make the cleanup job class a Spring service (so that the schedule could work).

```
[ ] Yes
[x] No
```


lgonczar:
> Also could we please enable the job on AAT(if not already) and disable for prod. After testing can be enabled for prod. Alternatively if we can write functional tests that should be fine.

I've disabled the job for prod. At the same time, I changed the schedule in AAT to every 10 minutes. This way, having many old files, won't have to wait 3+ days to verify the job.

PR #597, opened 2019-03-26
Update recipients for report sender

doncem:

### Change description ###

List of recipients must come through config which must be set from key vault. Since list is dynamic for painless config setup suggesting to store as comma separated value - usual approach for this kind of task.

Bonus: introducing extra config for auto configuration of `JavaMailSender` bean. ~Added `test-connection` flag. It tests connection before registering bean. Should fail the application failing that.~ Disabled at the moment. Giving the configuration is present, also _enabling_ possibility to automatically register report sender once `JavaMailSender` is enabled

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
@@ -3,8 +3,10 @@ package uk.gov.hmcts.reform.bulkscanprocessor.services.email;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.MimeMessageHelper;
+import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.ReportsService;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.ZipFileSummaryResponse;
 import uk.gov.hmcts.reform.bulkscanprocessor.util.CsvWriter;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
```diff
@@ -31,16 +31,11 @@
     public ReportSender(
         JavaMailSender mailSender,
         ReportsService reportsService,
-        String[] recipients
+        @Value("${reports.recipients:}") String reportRecipients
```
nitinprabhu: It would be nice to configure defaults in yaml file itself rather than here

nitinprabhu: reportRecipients -> reportRecipientsCSV ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderTest.java
```diff
@@ -39,23 +35,27 @@
...
         reportSender.send();
 
         // then
         MimeMessageParser msg = new MimeMessageParser(greenMail.getReceivedMessages()[0]).parse();
 
-        assertThat(msg.getTo()).extracting(Address::toString).containsExactlyElementsOf(asList(recipients));
+        assertThat(msg.getTo())
+            .extracting(Address::toString)
+            .hasSize(2)
+            .containsExactlyElementsOf(
```
nitinprabhu: why not check contains exactly only `Foo <foo@hmcts.net>` and `bar@hmcts.net`

src/main/resources/application.yaml
```diff
@@ -34,6 +34,18 @@ spring:
...
+    host: false #smtp.office365.com
+    password: password
+    port: 587
+    properties:
+      mail:
+        smtp:
+          auth: true
+          starttls:
+            enable: true
+    test-connection: true
```
nitinprabhu: Not feeling that strong but these are internally reports do we really want to fail the application is the test connection fails ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderTest.java
doncem: Was just _reusing_ what's there. With possible email configs

src/main/resources/application.yaml
doncem: Provided possible config early. This in particular was to get feedback ;)

PR #598, opened 2019-03-27
Feature: enable reports

doncem:

### Change description ###

**NOTE**: based on #597

Secret created in available key vaults:

```bash
INFO: Found secret reports-recipients in environment saat
INFO: Found secret reports-recipients in environment sprod
INFO: Found secret reports-recipients in environment demo
INFO: Found secret reports-recipients in environment aat
INFO: Found secret reports-recipients in environment prod
```

`prod` contains all listed in the ticket, others only bulk scan one with minor modification. There has to be some value. Reports will be and shall be enabled only for `prod`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/demo.tfvars
+++ b/infrastructure/demo.tfvars
@@ -9,3 +9,5 @@ orchestrator_notifications_task_delay = "3000"
 api_gateway_test_valid_certificate_thumbprint = "C4784AD48B4B99F427D6B56FB38184D0E6457744"
 blob_signature_verification_key_file = "nonprod_public_key.der"
 allowed_client_certificate_thumbprints = ["4AF638E82FBD9959A5B8286C673360AC2C2E7053"]
+
+smtp_host = "smtp.office365.com"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -133,6 +133,12 @@ module "bulk-scan" {
     QUEUE_NOTIFICATIONS_READ = "${data.terraform_remote_state.shared_infra.notifications_queue_primary_listen_connection_string}"
     QUEUE_PROCESSED_ENVELOPES_READ = "${data.terraform_remote_state.shared_infra.processed_envelopes_queue_primary_listen_connection_string}"
 
+    SMTP_HOST          = "${var.smtp_host}"
+    SMTP_USERNAME      = "${data.azurerm_key_vault_secret.smtp_username.value}"
+    SMTP_PASSWORD      = "${data.azurerm_key_vault_secret.smtp_password.value}"
+    REPORTS_CRON       = "${var.reports_cron}"
+    REPORTS_RECIPIENTS = "${data.azurerm_key_vault_secret.reports_recipients.value}"
+
     // silence the "bad implementation" logs
     LOGBACK_REQUIRE_ALERT_LEVEL = "false"
     LOGBACK_REQUIRE_ERROR_CODE  = "false"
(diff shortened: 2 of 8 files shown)


infrastructure/main.tf
```diff
@@ -179,6 +185,21 @@ data "azurerm_key_vault_secret" "s2s_secret" {
   vault_uri = "${local.s2s_vault_url}"
 }
 
+data "azurerm_key_vault_secret" "reports_recipients" {
+  name      = "reports-recipients"
+  vault_uri = "${data.azurerm_key_vault.key_vault.vault_uri}"
+}
+
+data "azurerm_key_vault_secret" "smtp_username" {
+  name      = "reports-email-username"
```
nitinprabhu: make sure this secret is added to demo vault too (same as aat vault)

infrastructure/main.tf
```diff
@@ -179,6 +185,21 @@ data "azurerm_key_vault_secret" "s2s_secret" {
...
+  vault_uri = "${data.azurerm_key_vault.key_vault.vault_uri}"
+}
+
+data "azurerm_key_vault_secret" "smtp_password" {
+  name      = "reports-email-password"
```
nitinprabhu: same here as above

infrastructure/variables.tf
```diff
@@ -117,6 +117,20 @@ variable "delete_rejected_files_cron" {
...
+
+# region reports
+variable "smtp_host" {
+  type        = "string"
+  default     = "false"
+  description = "SMTP host for sending out reports via JavaMailSender"
+}
+
+variable "reports_cron" {
+  default = "0 0 18 * * *"
```
nitinprabhu: comment would be nice what frequency it's running with :)

src/integrationTest/resources/application.properties
```diff
@@ -28,3 +28,5 @@ user.timezone=GMT
 
 # no communication with Service Bus
 spring.profiles.active=nosb
+
+reports.recipients=integration@test
```
nitinprabhu: may be use a proper email format ? something like` integration@fake.com`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
```diff
@@ -47,6 +49,8 @@ public ReportSender(
     }
     // endregion
 
+    @Scheduled(cron = "${reports.cron:-}")
```
nitinprabhu: can we please remove the default value from here and configure the prop in app yaml. Makes it consistent with what we have in other places.

infrastructure/demo.tfvars
```diff
@@ -9,3 +9,5 @@ orchestrator_notifications_task_delay = "3000"
 api_gateway_test_valid_certificate_thumbprint = "C4784AD48B4B99F427D6B56FB38184D0E6457744"
 blob_signature_verification_key_file = "nonprod_public_key.der"
 allowed_client_certificate_thumbprints = ["4AF638E82FBD9959A5B8286C673360AC2C2E7053"]
+
+smtp_host = "smtp.office365.com"
```
nitinprabhu: Does this determine if the reports are enabled or not. Is there any other way we can do this else let's add a comment here

doncem:
closing as people are confusing which pr is first in line

src/integrationTest/resources/application.properties
doncem: don't want to use domain fully. it is valid to assign partial domain and java mail sender allows it. it's like having you local machine setup with `username@localhost`

infrastructure/demo.tfvars
doncem: Same as feign clients are conditioned

PR #601, opened 2019-04-02
Update recipients for report sender

doncem:
Reference: original PR #597. Description below is copied

### Change description ###

List of recipients must come through config which must be set from key vault. Since list is dynamic for painless config setup suggesting to store as comma separated value - usual approach for this kind of task.

Bonus: introducing extra config for auto configuration of `JavaMailSender` bean. ~Added `test-connection` flag. It tests connection before registering bean. Should fail the application failing that.~ Disabled at the moment. Giving the configuration is present, also _enabling_ possibility to automatically register report sender once `JavaMailSender` is enabled

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderDisabledTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderDisabledTest.java
@@ -1,0 +1,27 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.services.email;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.context.ApplicationContext;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest(
+    properties = {
+        "spring.mail.host=false"
+    }
+)
+@RunWith(SpringRunner.class)
+public class ReportSenderDisabledTest {
+
+    @Autowired
+    private ApplicationContext context;
+
+    @Test
+    public void should_not_have_report_sender_in_context() {
+        assertThat(context.getBeanNamesForType(ReportSender.class)).isEmpty();
+    }
+}
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
@@ -3,8 +3,10 @@ package uk.gov.hmcts.reform.bulkscanprocessor.services.email;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.mail.javamail.JavaMailSender;
 import org.springframework.mail.javamail.MimeMessageHelper;
+import org.springframework.stereotype.Component;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.ReportsService;
 import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.ZipFileSummaryResponse;
 import uk.gov.hmcts.reform.bulkscanprocessor.util.CsvWriter;
@@ -15,6 +17,8 @@ import java.time.LocalDate;
 import java.util.List;
 import javax.mail.internet.MimeMessage;
 
+@Component
+@ConditionalOnBean(JavaMailSender.class)
 public class ReportSender {
 
     private static final Logger log = LoggerFactory.getLogger(ReportSender.class);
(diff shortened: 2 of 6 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderDisabledTest.java
```diff
@@ -0,0 +1,27 @@
...
+)
+@RunWith(SpringRunner.class)
+public class ReportSenderDisabledTest {
+
+    @Autowired
+    private ApplicationContext context;
+
+    @Test
+    public void should_not_have_report_sender_in_context() {
+        assertThat(context.getBeanNamesForType(ReportSender.class)).isEmpty();
```
nitinprabhu: This test is as such not required

doncem: proves the point `ReportSender` is not present when `JavaMailSender` is disabled. could even add more (if possible, not sure i tried) - check schedule is not registered and it would be pretty much functionally tested

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
```diff
@@ -31,16 +35,11 @@
...
     ) {
         this.mailSender = mailSender;
         this.reportsService = reportsService;
-
-        if (recipients == null) {
-            this.recipients = new String[0];
-        } else {
-            this.recipients = Arrays.copyOf(recipients, recipients.length);
-        }
+        this.recipients = reportRecipientsCsv.split(",");
```
rkondratowicz: does spring not support lists in config?
I'm pretty sure we've used it somewhere already. List of jurisdictions I think.

doncem: it's the way you store list of recipients. since they have to be populated in vault and members of list vary this can be configured as comma separated values. especially when content is as we have now

rkondratowicz: list of jurisdictions was stored the same way.
see here https://github.com/hmcts/bulk-scan-processor/pull/482

src/main/resources/application.yaml
```diff
@@ -34,6 +34,20 @@ spring:
...
+    host: ${SMTP_HOST:false}
+    password: password
+    port: 587
+    properties:
+      mail:
+        smtp:
+          auth: true
+          starttls:
+            enable: true
+    username: username
```
rkondratowicz: perhaps this could sit next to password.

src/main/resources/application.yaml
```diff
@@ -34,6 +34,20 @@ spring:
           lob:
             # silence the 'wall-of-text' - unnecessary exception throw about blob types
             non_contextual_creation: true
+  mail:
+    host: ${SMTP_HOST:false}
+    password: password
```
rkondratowicz: do we want this default value?

rkondratowicz: any reason to not use explicit env var here?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
doncem: i see what you mean now. automatically parse into array of strings as spring will do the split for you

src/main/resources/application.yaml
doncem: all env vars come in next pr when feature is actually enabled. all this is default/test setup

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSender.java
```diff
@@ -31,16 +35,11 @@
...
     ) {
         this.mailSender = mailSender;
         this.reportsService = reportsService;
-
-        if (recipients == null) {
-            this.recipients = new String[0];
-        } else {
-            this.recipients = Arrays.copyOf(recipients, recipients.length);
-        }
+        this.recipients = reportRecipients;
```
rkondratowicz: why this change??

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderTest.java
```diff
@@ -65,19 +67,15 @@ public void should_send_email_to_all_recipients() throws Exception {
...
         // given
         ReportsService reportsService = mock(ReportsService.class);
         JavaMailSender mailSender = mock(JavaMailSender.class);
 
         given(mailSender.createMimeMessage())
             .willReturn(new JavaMailSenderImpl().createMimeMessage());
 
-        willThrow(MailSendException.class)
-            .given(mailSender)
-            .send(any(MimeMessage.class));
```
rkondratowicz: where's the exception that was supposed to be handled now?

src/main/resources/application.yaml
lgonczar: I think it's best for the app to crash if it doesn't receive the host. Otherwise we might have problems figuring out what's wrong if, for example, this env var is not set.

PR #604, opened 2019-04-05
Feature: enable reports

doncem:

### Change description ###

Enabling reports for demo environment. All secrets are present in vault. Follow original PR at #598 if necessary. Difference is removal of default values

```
[ ] Yes
[x] No
```


charts/bulk-scan-processor/values.aat.template.yaml
```diff
@@ -5,7 +5,9 @@ java:
     DOCUMENT_MANAGEMENT_URL: "http://dm-store-aat.service.core-compute-aat.internal"
     SCAN_ENABLED: "true"
     SCAN_DELAY: "4000"
+    REPORTS_CRON: "0 0 0 1 1 ?"
```
rkondratowicz: 1st Jan at midnight?

doncem: apparently this is a value instead of having a default assigned in java code, requested by reviewer in previous PR :|

nitinprabhu: Is there a reason we are enabling it for PR ?

PR #607, opened 2019-04-08
Make doc subtype a free string

rkondratowicz:
A non-breaking change in preparation for https://tools.hmcts.net/jira/browse/BPS-585 (changes to doc types, arbitrary subtype sent as part of envelope metadata)

ℹ️ Note: both subtype and type are strings in orchestrator atm

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/common/DocumentSubtype.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/common/DocumentSubtype.java
@@ -1,9 +1,12 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.model.common;
 
-public class DocumentSubtype {
+public final class DocumentSubtype {
 
     public static final String WILL = "will";
     public static final String SSCS1 = "sscs1";
     public static final String COVERSHEET = "coversheet";
 
+    private DocumentSubtype() {
+        // util class
+    }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/common/DocumentSubtype.java
```diff
@@ -1,21 +1,9 @@
...
-import com.fasterxml.jackson.annotation.JsonValue;
+public class DocumentSubtype {
 
-public enum DocumentSubtype {
-    WILL("will"),
-    SSCS1("sscs1"),
-    COVERSHEET("coversheet");
+    public static final String WILL = "will";
+    public static final String SSCS1 = "sscs1";
+    public static final String COVERSHEET = "coversheet";
```
nitinprabhu: These values would be `Will` `SSCS1` and `Coversheet`

rkondratowicz: no change here

nitinprabhu: I meant those will also be getting updated. Of course since supplier would be sending these values there won't be any change on our side but at some point this needs to be updated :)

rkondratowicz: shouldn't the old mapping remain as is?

nitinprabhu: They are remain till services have confirmed the dependency on these subtypes have been removed. https://tools.hmcts.net/confluence/display/RBS/Bulk+Scanning+Feature+Changes

PR #613, opened 2019-04-10
Upgrade spring to latest (2.1.4)

doncem:
### Change description ###

Migration hard dependency on 2 configurations:

```yml
management:
  endpoint:
    health:
      show-details: always

spring:
  main:
    allow-bean-definition-overriding: true
```

and (possibly) soft dependencies:

```groovy
compile('uk.gov.hmcts.reform:health-spring-boot-starter:0.0.4')
compile('uk.gov.hmcts.reform:java-logging-appinsights:5.0.1')
compile('uk.gov.hmcts.reform:service-auth-provider-client:3.0.0')
```

The plan is to get rid of `allow-bean-definition-overriding` flag by fixing configuration files. `show-details` is mandatory for `health-spring-boot-starter` to work for PRs.

`health-spring-boot-starter:0.0.3` was incompatible and worked only with bean override flag on. `v0.0.4` is now compatible, see hmcts/health-spring-boot-starter#6

`service-auth-provider-client:3.0.0` is for spring boot 2.1 series

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -45,6 +48,8 @@ spring:
           auth: true
           starttls:
             enable: true
+  main:
+    allow-bean-definition-overriding: true
```
timja: what's left that requires this?

doncem: for application itself just auth config. integration tests need further investigation. i'll move that to integration config and apply fix to auth config ;)

PR #616, opened 2019-04-11
Remove health indicator for mail

doncem:
### Change description ###

Replace health indicator with application boot test check

```
[ ] Yes
[x] No
```


rkondratowicz:
Can you provide more context? / describe **why** this change was done?

doncem:
As part of health check smtp server will be hit constantly checking whether configured account is OK. This is overkill and not needed feature. Instead just do per application boot

PR #620, opened 2019-04-15
Run report schedule only on workday

doncem:

### Change description ###

Use crontab expression for "workday" schedule

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/variables.tf
+++ b/infrastructure/variables.tf
@@ -128,8 +128,8 @@ variable "smtp_host" {
 }
 
 variable "reports_cron" {
-  default     = "0 0 18 * * *"
-  description = "Cron signature for job to send out reports to be executed. Default value is 6PM everyday"
+  default     = "0 0 18 ? * MON-FRI"
+  description = "Cron signature for job to send out reports to be executed. Default value is 6PM (server time) every workday"
 }
 
 # endregion
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -51,7 +51,7 @@ spring:
 
 # setting env vars + defaults as env vars will be overridden otherwise
 reports:
-  cron: ${REPORTS_CRON:0 0 18 * * *}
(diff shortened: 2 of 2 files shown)


src/main/resources/application.yaml
```diff
@@ -51,7 +51,7 @@ spring:
 
 # setting env vars + defaults as env vars will be overridden otherwise
 reports:
-  cron: ${REPORTS_CRON:0 0 18 * * *}
+  cron: ${REPORTS_CRON:0 0 18 ? * W}
```
rkondratowicz: >W (weekday) – it is used to specify the weekday (Monday to Friday) nearest to a given day of the month. For example, if we specify “10W” in the <day-of-month> field, then it means the “weekday near to 10th of that month”. So if “10th” is a Saturday, then the job will be triggered on “9th”, and if “10th” is a Sunday, then it will trigger on “11th”. If you specify “1W” in the <day-of-month> and if “1st” is Saturday, then the job will be triggered on “3rd” which is Monday, and it will not jump back to the previous month

`MON-FRI` can be used 

https://www.baeldung.com/cron-expressions

doncem: I read it correctly, but then after doing all text still assumed as workday. derp

infrastructure/variables.tf
```diff
@@ -128,8 +128,8 @@ variable "smtp_host" {
 }
 
 variable "reports_cron" {
-  default     = "0 0 18 * * *"
-  description = "Cron signature for job to send out reports to be executed. Default value is 6PM everyday"
+  default     = "0 0 18 ? * MON-FRI"
```
nitinprabhu: Rather than having a default value ideal way would be default is null and prod.tfvars has that value. Job would be disabled on all other envrs anyways sp should not need the cron for non prod environments.

doncem: I had it as no value initially :/

PR #628, opened 2019-04-18
New task for incomplete envelopes monitoring

doncem:

[Daily alert for unprocessed envelopes](https://tools.hmcts.net/jira/browse/BPS-610)

### Change description ###

This PR adds a component class with empty scheduled content which will incorporate query in next PR. Query itself is introduced here. Using defaults in application config - necessary infrastructure changes and task completion will be in the following PR. Integration test assures there is no such bean registered in the application context

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
@@ -9,8 +9,6 @@ import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
 import org.springframework.test.context.junit4.SpringRunner;
 
 import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
 import javax.persistence.EntityManager;
 
 import static java.time.LocalDateTime.now;
@@ -125,14 +123,5 @@ public class EnvelopeRepositoryTest {
 
     private void dbHas(Envelope... envelopes) {
         repo.saveAll(asList(envelopes));
-    }
-
-    private void assertEnvelopeHasStatus(UUID envelopeId, Status status) {
-        // make sure there's no outdated cache
-        entityManager.clear();
-
-        Optional<Envelope> envelope = repo.findById(envelopeId);
-        assertThat(envelope).isPresent();
-        assertThat(envelope.get().getStatus()).isEqualTo(status);
     }
 }
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
@@ -95,14 +95,17 @@ public class EnvelopeRepositoryTest {
         // given
         dbHas(
             envelope("A.zip", "X", Status.PROCESSED),
-            envelope("B.zip", "Y", Status.COMPLETED),
-            envelope("C.zip", "Z", Status.UPLOAD_FAILURE),
-            envelope("D.zip", "Z", Status.COMPLETED)
+            envelope("B.zip", "Y", Status.NOTIFICATION_SENT),
+            envelope("C.zip", "Y", Status.COMPLETED),
+            envelope("D.zip", "Z", Status.UPLOAD_FAILURE),
+            envelope("E.zip", "Z", Status.COMPLETED)
         );
 
         // and update createAt to yesterday
         entityManager.createNativeQuery(
-            "UPDATE envelopes SET createdat = '" + now().minusDays(1) + "' WHERE zipfilename IN ('A.zip', 'B.zip')"
+            "UPDATE envelopes SET createdat = '"
+                + now().minusDays(1)
+                + "' WHERE zipfilename IN ('A.zip', 'B.zip', 'C.zip')"
         ).executeUpdate();
 
         // then
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
@@ -13,6 +13,7 @@ import java.util.Optional;
 import java.util.UUID;
 import javax.persistence.EntityManager;
 
+import static java.time.LocalDateTime.now;
 import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static uk.gov.hmcts.reform.bulkscanprocessor.helper.EnvelopeCreator.envelope;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -84,4 +85,12 @@
         @Param("maxFailureCount") int maxFailureCount,
         Pageable pageable
     );
+
+    @Query(
+        nativeQuery = true,
+        value = "SELECT COUNT(1) AS incomplete\n"
+            + "FROM envelopes\n"
+            + "WHERE DATE(createdat) < :date and status != 'COMPLETED'"
```
rkondratowicz: `AND`

nitinprabhu: Should it be not just doc notification sent ? else doc upload failure metadata validation failure will be included which we don't want ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesDisabledTaskTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+import org.springframework.test.context.junit4.SpringRunner;
+import uk.gov.hmcts.reform.bulkscanprocessor.config.IntegrationTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@IntegrationTest
+@RunWith(SpringRunner.class)
+public class IncompleteEnvelopesDisabledTaskTest {
```
rkondratowicz: do we need this?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
```diff
@@ -89,6 +90,25 @@ public void findByZipFileName_should_find_envelops_in_db() {
...
+        // given
+        dbHas(
+            envelope("A.zip", "X", Status.PROCESSED),
+            envelope("B.zip", "Y", Status.COMPLETED),
+            envelope("C.zip", "Z", Status.UPLOAD_FAILURE),
+            envelope("D.zip", "Z", Status.COMPLETED)
+        );
+
+        // and update createAt to yesterday
+        entityManager.createNativeQuery(
```
nitinprabhu: you could  just overload a method in envelope creator which accepts date field

doncem: created at is final - not changeable

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesDisabledTaskTest.java
doncem: same as discussed previously in hmcts/bulk-scan-processor#601 ([specific comment](https://github.com/hmcts/bulk-scan-processor/pull/601#discussion-diff-272147812)). each flag we have we could "assure" it works with such kind of test. perhaps even a bundle in one

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
doncem: 😇

doncem: let me review the states. i believe i missed this step in my tiny (as of yet) stream of commits

doncem:
I did consider. Additional changes to repo were just too small. Benefit now:

- you get to see the bigger picture easier
- less time wasted discussing/(dis)agreeing what's what

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
```diff
@@ -84,4 +85,15 @@
...
     );
+
+    // `notification sent` should be the final state for `metadata failure` and `signature failure`.
+    // including all as notification itself can fail too
+    @Query(
+        nativeQuery = true,
+        value = "SELECT COUNT(1) AS incomplete\n"
+            + "FROM envelopes\n"
+            + "WHERE DATE(createdat) < :date\n"
+            + "  AND status NOT IN ('METADATA_FAILURE', 'SIGNATURE_FAILURE', 'NOTIFICATION_SENT', 'COMPLETED')"
```
lgonczar: I think we should just compare to `COMPLETED` and everything that isn't in this state, should contribute to the count. `NOTIFICATION_SENT` should be taken into account, as it marks the point at which the envelope was sent to the orchestrator via queue (a lot can go wrong from here) and the failure statuses probably can't even exist for an envelope, because they represent situations in which an envelope wouldn't be created.

nitinprabhu: what about `DOC_UPLOAD_FAILURE`?

nitinprabhu: Failures such as metafile validation or signature error should not be considered as we would have notified Exela for those failures. Only files which are not processed because it failed due to doc upload or message was sent to queue but not completed should be reported in my opinion.

doncem: i don't think there's going back from those states. there has to be a new envelope

lgonczar: > Failures such as metafile validation or signature error should not be considered as we would have notified Exela for those failures. Only files which are not processed because it failed due to doc upload or message was sent to queue but not completed should be reported in my opinion.

Can envelopes even exist with those statuses? I don't think we'll create them in the DB in case of validation/signature failure. We don't need to list them here. The only status that should remain is `COMPLETED` - everything that's in a different one for too long is definitely a problem.

nitinprabhu: Good point @lgonczar yeah we create it only in events table. I forgot that we have events and envelope table. Yes query can be just !COMPLETED.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesTask.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Scheduled(cron = "${monitoring.incomplete-envelopes.cron}")
+    @SchedulerLock(name = "incomplete-envelopes-monitoring")
+    public void run() {
+        log.info("Checking for incomplete envelopes");
+
+        // get data and log here
+
+        log.debug("Finished checking for incomplete envelopes");
```
rkondratowicz: is this ever logged?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepository.java
doncem: lol. back to the beginning

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesTask.java
doncem: at the moment this method stands for wrapper what's coming. naturally we should have debugs but apparently it's not in team behaviour as inspected over the time. this log is not essential and just marks the fact job has finished. would argue with first one too, but, like i said, team is in favour of infos

PR #631, opened 2019-04-23
Include repository in task construct

doncem:

[daily alert for unprocessed envelopes](https://tools.hmcts.net/jira/browse/BPS-610)

### Change description ###

Implement incomplete envelopes task by including repository from #628 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesTaskTest.java
@@ -1,0 +1,55 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.tasks.monitoring;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Primary;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+import uk.gov.hmcts.reform.bulkscanprocessor.config.IntegrationTest;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.EnvelopeRepository;
+
+import java.time.LocalDate;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.AdditionalAnswers.delegatesTo;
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/monitoring/IncompleteEnvelopesTaskTest.java
```diff
@@ -0,0 +1,55 @@
...
+    private IncompleteEnvelopesTask task;
+
+    // only verifying integrity as validity is covered with unit test
+    // no dependencies involved - just log messages
+    // checking output is not recommended especially after moving to next gen of java-logging library
+    @Test
+    public void should_call_repository_method_and_not_fail_the_task() {
+        assertThatCode(task::run).doesNotThrowAnyException();
+
+        verify(repository).getIncompleteEnvelopesCountBefore(LocalDate.now());
```
lgonczar: Given we're only testing whether the repository was called, I think it could be a unit test.

doncem: Unit tests introduced as well. This integration test actually loads up repository as real class and therefore assures task is working properly

lgonczar: Integration tests for the repository cover the interaction with the DB, so in my opinion doing verifying calls on a mock (of this integration-tested class) would be ideal. This is just a suggestion.

doncem: This is not entirely a mock. It is a way to spy on repository proxy which is loaded by JPA for us.

I don't want to repeat myself again and again on a question similar to "how do you make sure it works?" This is how. Everything is loaded in application context and verifying necessary steps with spies. Why `mock` function seen here, explained in first sentence. This is not the first time it has been done.

P.S. this on it's own serves as semi-functional test too. Schedules work - we know that. Multiple functional tests have been created just to "verify" schedules are working

lgonczar: I'm aware of it being a spy. What I meant was this could be a mock instead, in a unit test. We've got DB interaction covered already at the right granularity level for integration tests. Given that, no other usage of a real repository will give us any more confidence regarding whether it works, other than functional and smoke tests, which rely on environment-specific configuration.

doncem:
The concerns taken into account. Further discussion can be addressed in person. As it stands - it is good to merge

PR #632, opened 2019-04-23
Enable incomplete envelopes task for production

doncem:

[daily alert for unprocessed envelopes](https://tools.hmcts.net/jira/browse/BPS-610)

### Change description ###

Introduce environment variables for easier customisation. Enabling for prod only

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -140,8 +140,8 @@ module "bulk-scan" {
     REPORTS_CRON       = "${var.reports_cron}"
     REPORTS_RECIPIENTS = "${data.azurerm_key_vault_secret.reports_recipients.value}"
 
-    INCOMPLETE_ENVELOPES_CRON    = "${var.incomplete_envelopes_cron}"
-    INCOMPLETE_ENVELOPES_ENABLED = "${var.incomplete_envelopes_enabled}"
+    INCOMPLETE_ENVELOPES_TASK_CRON    = "${var.incomplete_envelopes_cron}"
+    INCOMPLETE_ENVELOPES_TASK_ENABLED = "${var.incomplete_envelopes_enabled}"
 
     // silence the "bad implementation" logs
     LOGBACK_REQUIRE_ALERT_LEVEL = "false"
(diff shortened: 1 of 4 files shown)


infrastructure/main.tf
```diff
@@ -140,6 +140,9 @@ module "bulk-scan" {
     REPORTS_CRON       = "${var.reports_cron}"
     REPORTS_RECIPIENTS = "${data.azurerm_key_vault_secret.reports_recipients.value}"
 
+    INCOMPLETE_ENVELOPES_CRON    = "${var.incomplete_envelopes_cron}"
+    INCOMPLETE_ENVELOPES_ENABLED = "${var.incomplete_envelopes_enabled}"
```
nitinprabhu: may be INCOMPLETE_ENVELOPES_TASK_ENABLED ?

infrastructure/variables.tf
```diff
@@ -134,6 +134,20 @@ variable "reports_cron" {
 
 # endregion
 
+# region incomplete envelopes monitoring
+
+variable "incomplete_envelopes_cron" {
+  default     = "0 0 7 ? * MON-FRI"
```
nitinprabhu: may be we should run it after 9 ?

src/main/resources/application.yaml
```diff
@@ -51,8 +51,8 @@ spring:
 
 monitoring:
   incomplete-envelopes:
-    cron: 0 0 7 ? * MON-FRI # can be discussed later
-    enabled: false
+    cron: ${INCOMPLETE_ENVELOPES_CRON:0 0 7 ? * MON-FRI} # can be discussed later
```
nitinprabhu: Do we need the defaults here ?

doncem: i would prefer, same as reporting config couple lines below

infrastructure/variables.tf
doncem: as commented in config - up for discussion. but 9am sounds good to me. will change

src/main/resources/application.yaml
```diff
@@ -11,7 +11,7 @@ management:
 
 spring:
   application:
-    name: Bulk Scanning
+    name: Bulk Scan Processor
```
nitinprabhu: nice this will also make app insight map clear for tracking :)

PR #635, opened 2019-04-25
Force infra changes for app settings which are not applied in prod

doncem:
### Change description ###

Settings are not applied. Forcing the change

```
[ ] Yes
[x] No
```


doncem:
merging as infra is not executed for prs

PR #637, opened 2019-04-25
Update incomplete envelopes task config

doncem:
### Change description ###

There's problem applying config changes via our pipeline. As discussed - no harm to enable by default for the time being

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/aat.tfvars
+++ b/infrastructure/aat.tfvars
@@ -14,3 +14,5 @@ delete_rejected_files_cron = "0 0/10 * * * *"
 api_gateway_test_valid_certificate_thumbprint = "C4784AD48B4B99F427D6B56FB38184D0E6457744"
 allowed_client_certificate_thumbprints = ["4AF638E82FBD9959A5B8286C673360AC2C2E7053"]
 blob_signature_verification_key_file = "nonprod_public_key.der"
+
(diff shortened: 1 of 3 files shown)


infrastructure/aat.tfvars
```diff
@@ -14,3 +14,5 @@ delete_rejected_files_cron = "0 0/10 * * * *"
 api_gateway_test_valid_certificate_thumbprint = "C4784AD48B4B99F427D6B56FB38184D0E6457744"
 allowed_client_certificate_thumbprints = ["4AF638E82FBD9959A5B8286C673360AC2C2E7053"]
 blob_signature_verification_key_file = "nonprod_public_key.der"
+
+incomplete_envelopes_cron = "0 0 * * * *"
```
nitinprabhu: Is there a reason you are overriding it for AAT ? There is already a default set. So just the application yaml change would do ?

rkondratowicz: default is 9am weekdays, which is used on prod

doncem: we discussed this earlier today and figured it would be good to just let aat run the job so we can at least verify logs are there. not just waiting for a day. ideally default value should be "disabled" and this configured properly in azure portal. oh well

PR #643, opened 2019-04-30
Migrate ServiceBus `(get|set)Body` functions to new methods

doncem:
### Change description ###

Removing layer of deprecation and using same functionality as inside the ServiceBus library. No breaking change.

Next step - backport same library version jump in orchestrator. Until that happens - no version jump to 2.0.0 can happen

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageBodyRetriever.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageBodyRetriever.java
@@ -1,0 +1,19 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus;
+
+import com.microsoft.azure.servicebus.MessageBody;
+import org.springframework.util.CollectionUtils;
+
+import java.util.List;
+
+public final class MessageBodyRetriever {
+
+    public static byte[] getBinaryData(MessageBody messageBody) {
+        List<byte[]> binaryData = messageBody.getBinaryData();
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/MessageBodyRetriever.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.util.CollectionUtils;
+
+import java.util.List;
+
+public final class MessageBodyRetriever {
+
+    public static byte[] getBinaryData(MessageBody messageBody) {
+        List<byte[]> binaryData = messageBody.getBinaryData();
+
+        return CollectionUtils.isEmpty(binaryData) ? null : binaryData.get(0);
```
nitinprabhu: I am not sure why we are retrieving the first byte array ?
https://docs.microsoft.com/bs-latn-ba/java/api/com.microsoft.azure.servicebus.messagebody.getbinarydata?view=azure-java-stable#returns

doncem: From servicebus library itself, extracted get/set usage:

```java
@Override
@Deprecated
public byte[] getBody()
{
    return Utils.getDataFromMessageBody(this.messageBody);
}
```

```java
// Util.class
// inspect everything is package private
static byte[] getDataFromMessageBody(MessageBody messageBody) {
    List<byte[]> binaryData = messageBody.getBinaryData();
    if (binaryData == null || binaryData.size() == 0) {
        return null;
    } else {
        return binaryData.get(0);
    }
}
```

PR #644, opened 2019-05-01
Clean up FailedDocUploadProcessor

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/FailedDocUploadProcessor.java
```diff
@@ -55,19 +54,18 @@ public FailedDocUploadProcessor(
         EnvelopeProcessor envelopeProcessor,
         EnvelopeRepository envelopeRepository,
         ProcessEventRepository eventRepository,
-        ContainerMappings containerMappings,
+        ContainerMappings containerMappings, // NOSONAR
```
lgonczar: Why are we keeping this parameter here?

rkondratowicz: see the existing comment in code above:
>     // NOTE: this is needed for testing as children of this class are instantiated

In my opinion the usage of inheritance here is damaging, I don't plan to change this as part of this PR.

PR #646, opened 2019-05-01
Clean up yaml config for services accessing envelopes

rkondratowicz:
For some reason this config was also used for specifying jurisdiction for the retry job and empty entries were added.
This is no longer the case. Entries can be removed as those services do not read or update their envelopes.

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -178,12 +178,6 @@ envelope-access:
      - jurisdiction: BULKSCAN
        readService: bulk_scan_processor_tests
        updateService: bulk_scan_processor_tests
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -178,12 +178,6 @@ envelope-access:
      - jurisdiction: BULKSCAN
```
nitinprabhu: Do we need bulk scan and SSCS ? Can we not completely get rid of this config ?

rkondratowicz: bulkscan is a test service. Whether SSCS uses the read endpoint I do not know.
Just removing the services that were there to enable the retry job.

PR #651, opened 2019-05-10
Fix the check for processed envelope

lgonczar:
### Change description ###

Fix the check for processed envelope.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
@@ -38,7 +38,7 @@ public enum Status {
     }
 
     public boolean isProcessed() {
-        return this == PROCESSED || this == NOTIFICATION_SENT;
+        return this == PROCESSED || this == NOTIFICATION_SENT || this == COMPLETED;
     }
 
 }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
@@ -37,6 +37,10 @@ public enum Status {
         }
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/StatusTest.java
```diff
@@ -10,7 +10,7 @@
     @Test
     public void notification_sent_or_processed_only_is_processed() {
```
nitinprabhu: method name needs update ?

lgonczar: true, done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
```diff
@@ -38,7 +38,7 @@
     }
 
     public boolean isProcessed() {
-        return this == PROCESSED || this == NOTIFICATION_SENT;
+        return this == PROCESSED || this == NOTIFICATION_SENT || this == COMPLETED;
```
rkondratowicz: can you explain?
what does 'processed' actually mean and when is it used?

rkondratowicz: adding javadoc to this method would be great, I don't think it's clear with those 3 different statuses, out of which one is called 'processed'

rkondratowicz: also, probably good idea to rename this method

lgonczar: I'll add a comment. The name of me method may be right - all those statuses represent a processed envelope, from processor's perspective. And `PROCESSED` itself is probably going away, as it duplicates another status - `UPLOADED`.

PR #652, opened 2019-05-10
Clean up checking whether file can be deleted

rkondratowicz:
The `isProcessed()` method on Status was only used in one place for very particular reason.
Moving the logic to where it's used.

IMHO clearer.

Also added logging for when we decide the files should not be deleted yet.


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -244,6 +246,8 @@ private void deleteIfProcessed(CloudBlockBlob cloudBlockBlob, Envelope envelope,
                     envelopeProcessor.saveEnvelope(envelope);
                     log.info("Marked envelope from file {} (container {}) as deleted", blobName, containerName);
                 }
+            } else {
+                log.info("File not ready to be deleted.");
```
lgonczar: Good idea. Needs file name.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -303,6 +307,14 @@ private boolean isReadyToBeProcessed(CloudBlockBlob blob) {
         return blob.getProperties().getLastModified().before(cutoff);
     }
 
+    private boolean isReadyToBeDeleted(Envelope envelope) {
```
lgonczar: I agree with this change, it should have never been the responsibility of the enum to decide. This, however, exposes the lack of testing based on the status. Still, there's a story to address testability of the job.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
rkondratowicz: oops. fixed

PR #661, opened 2019-05-15
BPS-641 Migration script to add documentUuid column

aliveni:
…s and set the value from documentUrl column

### Change description ###
DB Migration script to add documentUuid column to scannable items table.
~Update script to set documentUuid value from the documentUrl for the existing data.~

```
[ ] Yes
[ x] No
```
**Update:** 
Setting documentUuid value for the existing records will be added in a new PR.

### DIFF
--- a/src/main/resources/db/migration/V040__Add_document_uuid_to_scannable_items.sql
+++ b/src/main/resources/db/migration/V040__Add_document_uuid_to_scannable_items.sql
@@ -1,9 +1,3 @@
 ALTER TABLE scannable_items
   ADD COLUMN documentUuid VARCHAR(100) NULL;
 
--- Extract the text after the last slash (/) in the documentUrl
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V040__Add_document_uuid_to_scannable_items.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE scannable_items
+  ADD COLUMN documentUuid VARCHAR(100) NULL;
+
+UPDATE scannable_items
+   SET documentUuid = (
+     SELECT documentId[1] FROM
+        REGEXP_MATCHES(
+        documenturl,
```
rkondratowicz: case doesn't matter is sql I guess, but could be `documentUrl` for readability

PR #667, opened 2019-05-16
Bps 641 populate document UUID values

aliveni:

### Change description ###
DB migration script to populate documentUuid values from documentUrl

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/resources/db/migration/V041__Update_document_uuid_values.sql
+++ b/src/main/resources/db/migration/V041__Update_document_uuid_values.sql
@@ -1,3 +1,4 @@
 -- Extracts the text after the last slash (/) in the documentUrl
 UPDATE scannable_items
  SET documentUuid = REVERSE(SPLIT_PART(REVERSE(documentUrl), '/', 1))
+ WHERE documentUuid IS NULL;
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V041__Update_document_uuid_values.sql
```diff
@@ -0,0 +1,3 @@
+-- Extracts the text after the last slash (/) in the documentUrl
+UPDATE scannable_items
+ SET documentUuid = REVERSE(SPLIT_PART(REVERSE(documentUrl), '/', 1))
```
rkondratowicz: are both slots writing to that new column now?

aliveni: the new column is in both slots. The previous pr (which writes the value to the new column) is merged but deployed once. the second build  is still in progress

nitinprabhu: should we include a where clause here where documentuuid is null ?

aliveni: added where clause

aliveni:
Merged duplicate PR https://github.com/hmcts/bulk-scan-processor/pull/671

PR #675, opened 2019-05-22
DB migration script to make documentUuid field not null

aliveni:

### Change description ###
Added migration script to make the documentUuid field not null.
Note: `documentUrl` field is Nullable so there is no script to drop the `not null` constraint (As mentioned in the ticket)

```
[ ] Yes
[x ] No
```


aliveni:
Closing this PR as we do not need to add `not null` constraint on `documentUuid` field

PR #678, opened 2019-05-24
Re-process stuck envelopes

doncem:

[[SPIKE]: Handle envelopes stuck in `CREATED` status](https://tools.hmcts.net/jira/browse/BPS-628)

### Change description ###

A non-destructive integration of a spike which solves rare situation of envelope being stuck in `CREATED` status. All steps are present so just needed to re-use existing envelope.

Breaking apart task is not involved - another ticket. And would be very harmful

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -34,7 +34,9 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.verify;
+import static uk.gov.hmcts.reform.bulkscanprocessor.entity.Status.CREATED;
 import static uk.gov.hmcts.reform.bulkscanprocessor.entity.Status.PROCESSED;
 import static uk.gov.hmcts.reform.bulkscanprocessor.entity.Status.UPLOAD_FAILURE;
 import static uk.gov.hmcts.reform.bulkscanprocessor.helper.DirectoryZipper.zipAndSignDir;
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -61,6 +63,34 @@ public void should_read_blob_and_save_metadata_in_database_when_zip_contains_met
...
+
+        // pretend it is stuck
+        envelopeRepository.saveAll(
+            envelopeRepository
+                .findAll()
+                .stream()
+                .peek(envelope -> {
+                    envelope.setStatus(CREATED);
+                    envelope.setZipDeleted(false);
+                })
```
rkondratowicz: please use `.map`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -174,14 +174,24 @@ private void processZipFileIfEligible(CloudBlobContainer container, String zipFi
         Envelope existingEnvelope =
             envelopeProcessor.getEnvelopeByFileAndContainer(container.getName(), zipFilename);
 
-        if (existingEnvelope != null) {
+        if (existingEnvelope != null && !existingEnvelope.getStatus().equals(Status.CREATED)) {
             log.warn(
-                "Envelope for zip file {} (container {}) already exists. Aborting its processing. Envelope ID: {}",
+                "Envelope for zip file {} (container {}) with {} status already exists. "
+                    + "Aborting its processing. Envelope ID: {}",
```
rkondratowicz: Perhaps it would be more suitable to add envelope status in the last part? (after ID)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -56,6 +56,9 @@ protected void processParsedEnvelopeDocuments(
         List<Pdf> pdfs,
         CloudBlockBlob cloudBlockBlob
     ) {
+        if (envelope == null) {
+            return;
```
rkondratowicz: why is envelope null? Looks like something that deserves a log

doncem:
@lgonczar as part of spike you are allowed to produce some code. They may turn out as a real thing or binned. If one achieves something during spike why not share and then whole team can elaborate whether it's good or bad or.. good enough. The way I see it it's good enough. Race condition is valid point. Highly unlikely. Blob lease is in place and, unless it is released or expired - you won't even reach the place amended here. The other way I see it is just a drastic painful split of processor which is another story tbh. Separate task is fairly non-acceptable. But an option.

Once again - just a spike. This fraction of code I did while I was sleeping - baby task. No need to drop dirt just because "it is a spike only".

Spikes **do** produce code whether it's throw away or real thing.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
doncem: It is logged. There's a separation in zip processor in this PR. There's a chance envelope can be null

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
doncem: java doesn't live in immutable world, but yep, whatever ;)

rkondratowicz:
Closing to free up preview resources

PR #694, opened 2019-06-21
Include envelope status in zip file summary (json only)

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-688

### Change description ###

This is a proposal, so if you don't think we should make this change, please let me know.

Include envelope status in zip file summary (JSON version only), so that if there's a discrepancy between the status of the last event and the status of the envelope (which seems to be the case somewhere), we should be able to discover it easily.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
@@ -13,30 +13,33 @@ public interface ZipFilesSummaryRepository extends JpaRepository<Envelope, UUID>
 
     @Query(
         nativeQuery = true,
-        value = "SELECT createdEvent.container, createdEvent.zipfilename, createdEvent.createdDate, \n"
-            + "  lastEventName.event AS status,\n"
-            + "  (CASE WHEN lastEventName.event = 'COMPLETED'\n"
-            + "        THEN lastEvent.eventDate ELSE null END\n"
-            + "  ) AS completedDate\n"
-            + "FROM (\n"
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
```diff
@@ -13,30 +13,33 @@
...
+            + "  ON createdEvent.container = lastEvent.container "
+            + "  AND createdEvent.zipfilename = lastEvent.zipfilename "
+            + "JOIN ( "
+            + "  SELECT container, zipfilename, createdat, event "
+            + "  FROM process_events "
+            + ") lastEventName " /* latest event of the zipfile */
+            + "  ON lastEvent.container = lastEventName.container "
+            + "  AND lastEvent.zipfilename = lastEventName.zipfilename "
+            + "  AND lastEvent.eventDate = lastEventName.createdat "
+            + "LEFT JOIN envelopes AS envelope "
```
lgonczar: This left join (and the related column in the select part) is the only change here. In other lines, I removed unnecessary `\n`s from the query, hoping the diff would be smarter about displaying those changes.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/reports/ZipFilesSummaryReportItem.java
```diff
@@ -28,8 +28,11 @@
     @JsonProperty("container")
     public final String jurisdiction;
 
-    @JsonProperty("status")
-    public final String status;
+    @JsonProperty("last_event_status")
+    public final String lastEventStatus;
```
lgonczar: Introducing envelope status created the need for this field to be more descriptive.

PR #699, opened 2019-07-02
Throw exception when metafile contains unknown properties

aliveni:
https://tools.hmcts.net/jira/browse/BPS-670

### Change description ###
Updated `MetafileJsonValidator` to throw an exception when the metafile contains unknown properties. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidator.java
@@ -2,7 +2,6 @@ package uk.gov.hmcts.reform.bulkscanprocessor.validation;
 
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
 import com.github.fge.jsonschema.core.exceptions.ProcessingException;
 import com.github.fge.jsonschema.core.report.ProcessingReport;
 import com.github.fge.jsonschema.main.JsonSchema;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidator.java
```diff
@@ -45,6 +46,10 @@ public void validate(byte[] metafile, String zipFileName) throws IOException, Pr
     }
 
     public InputEnvelope parseMetafile(byte[] metafile) throws IOException {
-        return MAPPER.readValue(metafile, InputEnvelope.class);
+        try {
+            return MAPPER.readValue(metafile, InputEnvelope.class);
+        } catch (UnrecognizedPropertyException e) {
+            throw new InvalidEnvelopeSchemaException("Unrecognized property in the metadata file", e);
+        }
```
rkondratowicz: looks like we validate the json file agains the schema before we even attempt to map it to a java object:
```
        try {
            schemaValidator.validate(metadataStream, zipFileName);

            return schemaValidator.parseMetafile(metadataStream);
        } catch (JsonParseException exception) {
            // invalid json files should also be reported to provider
            throw new InvalidEnvelopeSchemaException("Error occurred while parsing metafile", exception);
        }
```
Perhaps we could adjust the schema validation to not allow unknown fields somehow.

rkondratowicz: or, alternatively, just add `UnrecognizedPropertyException` to catch above.

rkondratowicz: or even `IOException`

aliveni: I've just found that we can include a property (`additionalProperties`) in the schema file to validate unknown properties in the metafile.
I added `additionalProperties: false` to `metafile-schema.json` which forces the validator to throw an exception when the metafile contains unknown properties.

PR #706, opened 2019-07-10
Clarify building ocr for queue message

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -143,25 +141,19 @@ public String toString() {
         return envelope
             .getScannableItems()
             .stream()
-            .filter(si -> si.getOcrData() != null)
-            .map(item -> convertFromInputOcrData(item.getOcrData()))
-            .findFirst()
+            .filter(item -> item.getOcrData() != null)
+            .findFirst() // there's always only one scannable item with OCR data
+            .map(item -> item.getOcrData())
+            .map(ocrData -> ocrData
```
lgonczar: Why not have this mapping in a separate method? Could be one method instead of two, but in my opinion it reads easier if these details are hidden behind a name.

rkondratowicz:
closing because implementing BPS-711 will require changes anyway

PR #708, opened 2019-07-11
Add client for ocr validation endpoint

rkondratowicz:

Not used (yet), but with tests.

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/OcrValidationClientTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/OcrValidationClientTest.java
@@ -1,13 +1,17 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.ocrvalidation;
 
+import com.github.tomakehurst.wiremock.client.ResponseDefinitionBuilder;
 import com.github.tomakehurst.wiremock.core.Options;
+import org.apache.commons.lang3.tuple.Pair;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.contract.wiremock.AutoConfigureWireMock;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringRunner;
-import org.springframework.web.client.HttpServerErrorException.InternalServerError;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
+import org.springframework.web.client.HttpStatusCodeException;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.IntegrationContextInitializer;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.IntegrationTest;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.Profiles;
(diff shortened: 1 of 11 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/OcrValidationClientTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+@ActiveProfiles( {
+    IntegrationContextInitializer.PROFILE_WIREMOCK,
+    Profiles.SERVICE_BUS_STUB,
+    Profiles.STORAGE_STUB
+})
+@AutoConfigureWireMock
+@IntegrationTest
+@RunWith(SpringRunner.class)
+public class OcrValidationClientTest {
```
nitinprabhu: Could we please add tests when there is success and when there are only warnings ? and also for invalid s2s token and 403 scenarios ?

rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/OcrValidationClient.java
```diff
@@ -0,0 +1,30 @@
...
+@Component
+public class OcrValidationClient {
+
+    private final RestTemplate restTemplate;
+
+    public OcrValidationClient(RestTemplate restTemplate) {
+        this.restTemplate = restTemplate;
+    }
+
+    public ValidationResponse validate(String url, FormData formData, String s2sToken) {
```
lgonczar: Could it get some s2s token provider maybe?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/model/req/OcrDataField.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.ocrvalidation.client.model.req;
+
+public class OcrDataField {
+
+    public final String name;
+    public final String value;
```
lgonczar: Don't these two need `@JsonProperty` annotation?

lgonczar:
Worth adding some wiremock tests or whatever kind of tests that would check if the client is actually hitting some enpoint.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/OcrValidationClient.java
rkondratowicz: my plan is for the service to take a provider in constructor

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/model/req/OcrDataField.java
rkondratowicz: only if you want to set the json property name explicitly

lgonczar:
> Worth adding some wiremock tests or whatever kind of tests that would check if the client is actually hitting some endpoint.

Which is exactly what we've already got. Sorry, my bad.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/OcrValidationClient.java
rkondratowicz: to be more specific, I plan to create `OcrValidationService` that will take envelope input models as parameters, map them to REST api models and call the client from this PR.

PR #710, opened 2019-07-11
Feature: make DCN unique across all envelopes

doncem:

[Reject envelopes containing docs with DCN received in the past](https://tools.hmcts.net/jira/browse/BPS-712)

### Change description ###

Migration script to add constraint on dcn field. Handling specific exception introduced due to that. Constraint violation is reported to provider

```
[x] Yes // need to clean dbs in test envs. production is good
[ ] No
```


src/functionalTest/resources/1111006.metadata.json
```diff
@@ -9,14 +9,14 @@
   "envelope_classification": "exception",
   "scannable_items": [
       {
-        "document_control_number": "1111006",
+        "document_control_number": "$$1111006$$",
```
rkondratowicz: is the middle part needed? maybe we could just provided the entire DCN?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -140,24 +148,43 @@ public String getRandomFilename(String suffix) {
...
+
+                zos.write(sb.toString().getBytes());
                 zos.closeEntry();
             }
         }
         return outputStream.toByteArray();
     }
 
+    private String generateDcnNumber() {
+        return DateTimeFormatter.ofPattern("yyyyMMDDHHmmss").format(Instant.now()) + RANDOM.nextInt();
```
rkondratowicz: `documentControlNumber VARCHAR(100)` so maybe just UUID?

src/functionalTest/resources/1111006.metadata.json
doncem: kept same formatting as before. dollar sign is not great for regex really. did not run integration suite in the office to verify everything as they take too long. will do that from home - much faster

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
doncem: no real preference.. can do uuid, can do time based. uuid is more of a "real life identifier" while this one is clearly test purpose

doncem:
Closing as following exception (does not!) make sense 🤕:

```bash
Caused by: uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidEnvelopeSchemaException: Failed validation for file 1722015180_12-02-2019-00-00-00.test.zip against schema. Errors:
	error: ECMA 262 regex "^[0-9]+$" does not match input string "1563204150244-1245830519"
    level: "error"
    schema: {"loadingURI":"#","pointer":"/properties/scannable_items/items/properties/document_control_number"}
    instance: {"pointer":"/scannable_items/1/document_control_number"}
    domain: "validation"
    keyword: "pattern"
    regex: "^[0-9]+$"
    string: "1563204150244-1245830519"
```

the dash in that disk should not be there.

will create separate pr to start fresh

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
rkondratowicz: only digits allowed in json schema, nevermind

PR #711, opened 2019-07-12
Add ocr validator

rkondratowicz:

Not used yet. Just adding service with tests.

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
@@ -27,7 +27,7 @@ public class OcrValidator {
         this.client = client;
     }
 
-    public void validate(InputEnvelope envelope) {
+    public void assertIsValid(InputEnvelope envelope) {
         List<InputScannableItem> docsWithOcr =
             envelope
                 .scannableItems
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
@@ -90,7 +90,7 @@ public class OcrValidator {
             .filter(mapping -> Objects.equals(mapping.getPoBox(), envelope.poBox))
             .findFirst()
             .map(mapping -> mapping.getOcrValidationUrl())
-            .filter(Strings::isNullOrEmpty);
+            .filter(target -> !Strings.isNullOrEmpty(target));
     }
 
     private FormData toFormData(InputScannableItem doc) {
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
@@ -5,6 +5,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Component;
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
 import uk.gov.hmcts.reform.bulkscanprocessor.config.ContainerMappings;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.OcrValidationException;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.blob.InputEnvelope;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -0,0 +1,105 @@
...
+        AuthTokenGenerator authTokenGenerator
+    ) {
+        this.client = client;
+        this.containerMappings = containerMappings;
+        this.authTokenGenerator = authTokenGenerator;
+    }
+    //endregion
+
+    public void assertIsValid(InputEnvelope envelope) {
+        Optionals.ifAllPresent(
```
doncem: so in case following functions return empty optional values, this function will just return null. is this expected behaviour?

doncem: how about just flattening the functionality here? so instead using some library, use core java:

```java
option1.flatmap(v1 -> option2.map(v2 - try block))
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -0,0 +1,105 @@
...
+                                log.info("Validation ended with warnings. File name: {}", envelope.zipFileName);
+                                break;
+                            default:
+                                break;
+                        }
+                    })
+                    .onFailure(exc -> {
+                        log.error("Error calling validation endpoint. Url: {}", validationUrl, exc);
+                        // log error and proceed
+                    })
```
doncem: would be nice to have a biconsumer reference used here giving the current signature of the method. would be even nicer if, somehow, is possible to reduce usage of `option` and `try` in one place as separate core elements here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -0,0 +1,105 @@
...
+            .stream()
+            .filter(mapping -> Objects.equals(mapping.getPoBox(), envelope.poBox))
+            .findFirst()
+            .map(mapping -> mapping.getOcrValidationUrl())
+            .filter(url -> !Strings.isNullOrEmpty(url));
+    }
+
+    private Optional<InputScannableItem> findDocWithOcr(InputEnvelope envelope) {
+        List<InputScannableItem> docsWithOcr =
+            envelope
```
doncem: just wondering as seen in many places: why is this in new line?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -0,0 +1,105 @@
...
+                        }
+                    })
+                    .onFailure(exc -> {
+                        log.error("Error calling validation endpoint. Url: {}", validationUrl, exc);
+                        // log error and proceed
+                    })
+        );
+    }
+
+    private Optional<String> findValidationUrl(InputEnvelope envelope) {
```
doncem: only `poBox` is used in this method. shall we pass that instead?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -0,0 +1,105 @@
...
+    private final AuthTokenGenerator authTokenGenerator;
+
+    //region constructor
+    public OcrValidator(
+        OcrValidationClient client,
+        ContainerMappings containerMappings,
+        AuthTokenGenerator authTokenGenerator
+    ) {
+        this.client = client;
+        this.containerMappings = containerMappings;
```
doncem: Only `getMappings` used in one function. shall we just build object with that value instead? it'll be private anyway

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
rkondratowicz: `Optionals.ifAllPresent` does not return a value, nor does the `assertIsValid` method

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
rkondratowicz: why not? ;)
To have `.scannableItems` and all the further operations close to envelope

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
rkondratowicz: Not sure.
We'd still need a map (from poBox to Url).

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
rkondratowicz: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
doncem: Em this now definitely looks big enough - candidate for a split. Whichever way, biconsumer, onsuccess/onfailure registrar, both 🙃 🤷‍♂

PR #714, opened 2019-07-15
Feature/make dcn unique across all envelopes

doncem:
### Change description ###

Reference #710 - original PR. Just recreating to start fresh resources

```
[x] Yes // please use reference
[ ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -295,6 +297,24 @@ private ZipFileProcessingResult processZipFileContent(
...
+            if (ex.getCause() instanceof ConstraintViolationException) {
+                ConstraintViolationException exception = (ConstraintViolationException) ex.getCause();
+                log.warn(
+                    "Rejected file {} from container {} - DB constraint violation {}",
+                    zipFilename,
+                    containerName,
+                    exception.getConstraintName(),
+                    exception
+                );
+                handleInvalidFileError(Event.FILE_VALIDATION_FAILURE, containerName, zipFilename, leaseId, exception);
```
lgonczar: I think we should be very strict here and rely on the DNC-specific constraint to conclude the file is invalid. Otherwise, it's a file processing error (to be investigated by us) which shouldn't result in rejection.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
```diff
@@ -128,6 +128,14 @@ public void should_not_accept_invalid_statuses_when_reading_envelopes_by_status(
     private List<EnvelopeResponse> envelopesInDb() throws Exception {
         Envelope envelope = EnvelopeCreator.envelope("BULKSCAN", Status.PROCESSED);
         envelope.setZipFileName("7_24-06-2018-00-00-00.zip"); // matches expected response file
+        // todo: make this better :(
+        envelope.getScannableItems().forEach(item -> {
```
lgonczar: Why do we set DCNs like this?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: Let's organise a chat as this conversation might get stuck or go on forwards and backwards 😕

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
doncem: Please review full PR and then make questions/assumptions.

This is because we now make DCN unique and `EnvelopeCreator.envelope` is used across many tests. Since it is dynamic and verification part is still static, for integration test it is fine to flip back to previous value which makes single test case pass. That single test case verifies full response object returned by controller giving the data (which is presumably returned from db) is given.

P.S. The change in `EnvelopeCreator.envelope` can be inspected further below in this PR

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -140,24 +146,43 @@ public String getRandomFilename(String suffix) {
...
+
+                zos.write(sb.toString().getBytes());
                 zos.closeEntry();
             }
         }
         return outputStream.toByteArray();
     }
 
+    private String generateDcnNumber() {
+        return Long.toString(System.currentTimeMillis()) + Math.abs(RANDOM.nextInt());
```
lgonczar: Why the milliseconds in front?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
lgonczar: By looking at the code it's very hard to figure out where these values come from, what they mean and why we're setting them. How about feeding `expectedEnvelopes()` with generated DCNs instead, so that the expected JSON would reflect the actual unchanged envelopes? Would require some dynamic replacements in the JSON file.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: Or can introduce extra condition like:
```java
if (exception.getConstraintName().equals("scannable_item_dcn")) {
    handleInvalidFileError(Event.FILE_VALIDATION_FAILURE, containerName, zipFilename, leaseId, exception);
} else {
    handleEventRelatedError(Event.DOC_FAILURE, containerName, zipFilename, exception);
}
```
would that be sufficient?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
doncem: exactly, extra replacements for json files. i'd rather keep it as separate pr hence the `//todo` comment

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: They don't hurt, though

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: Fine by me

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -316,6 +333,27 @@ private boolean isReadyToBeDeleted(Envelope envelope) {
...
+        log.warn(
+            "Rejected file {} from container {} - DB constraint violation {}",
+            zipFilename,
+            containerName,
+            exception.getConstraintName(),
+            exception
+        );
+
+        if (exception.getConstraintName().equals("scannable_item_dcn")) {
+            handleInvalidFileError(Event.FILE_VALIDATION_FAILURE, containerName, zipFilename, leaseId, exception);
```
lgonczar: Just wondering - will Exela know why the file was invalid?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -140,24 +146,43 @@ public String getRandomFilename(String suffix) {
...
+                Matcher matcher = pattern.matcher(metadataTemplate);
+                StringBuffer sb = new StringBuffer();
+
+                while (matcher.find()) {
+                    matcher.appendReplacement(sb, tokens.get(matcher.group(1)));
+                }
+
+                matcher.appendTail(sb);
+
+                zos.write(sb.toString().getBytes());
```
lgonczar: The part where things are replaced in metadata file is very hard to follow. Maybe it could be simplified or extracted to a separate method?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: I wonder - will the third party that uploaded the envelope will know why the file was invalid?

doncem: the message will be in a form of

```java
new ErrorNotificationRequest(
            message.zipFileName,
            message.poBox,
            message.documentControlNumber,
            message.errorCode.name(),
            message.errorDescription,
            message.id
        );
```

where `errorDescription = exception.getMessage()` which from Constraint Exception transforms to `could not execute statement` :(. ok, perhaps need a better one and can re-use existing custom exception about duplicate dcns

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
```diff
@@ -128,6 +128,14 @@ public void should_not_accept_invalid_statuses_when_reading_envelopes_by_status(
...
+            if (item.getOcrData() == null) {
+                item.setDocumentControlNumber("1111001");
+            } else {
+                item.setDocumentControlNumber("1111002");
```
rkondratowicz: do those string have a meaning? If they do, maybe a comment would help

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -316,6 +334,35 @@ private boolean isReadyToBeDeleted(Envelope envelope) {
...
+
+        if (exception.getConstraintName().equals("scannable_item_dcn")) {
+            handleInvalidFileError(
+                Event.FILE_VALIDATION_FAILURE,
+                containerName,
+                zipFilename,
+                leaseId,
+                new DuplicateDocumentControlNumbersInEnvelopeException(
+                    "Received envelope with 'document_control_number' already present in the system"
+                )
```
rkondratowicz: in all other places an exception that has been thrown is sent to this method, while here it is artificially created, why?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -140,24 +146,43 @@ public String getRandomFilename(String suffix) {
     ) throws IOException {
         ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
         try (ZipOutputStream zos = new ZipOutputStream(outputStream)) {
+            Map<String, String> tokens = new ConcurrentHashMap<>();
```
rkondratowicz: do you expect multiple threads accessing this map?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -140,24 +146,43 @@ public String getRandomFilename(String suffix) {
...
+
             for (String file : files) {
-                zos.putNextEntry(new ZipEntry(file));
+                String dcn = generateDcnNumber();
+                tokens.put("##" + file.replace(".pdf", "") + "##", dcn);
+                zos.putNextEntry(new ZipEntry(dcn + ".pdf"));
```
rkondratowicz: why did the file name change? `files` is no longer files if we no longer use the elements to name files.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
rkondratowicz: Could be as simply iterate over `token` and do `replaceAll`.
No performance to gain imho if that's what you were after here originally.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: Oh, I missed that. Yes, that's weird. Maybe we could have an override to handle this kind of situation. Plus the exception's name suggests the envelope contains duplicate control numbers within itself.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: I'd separate file names from DCNs. They used to be the same by accident or convention, but there's no need for that.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: That's what I was thinking, too. It will only be negligibly slower, i.e. no difference to be noticed. And we're talking about tests. The benefit will be simplicity.

doncem: `replaceAll` replaces any regex occurrence with a single (same) string. there are metafiles with 2 different scannable items

lgonczar: I was thinking about something more or less like this:
```
        tokens
            .entrySet()
            .stream()
            .reduce(
                metadataTemplate,
               (data, token) -> data.replaceAll(token.getKey(), token.getValue())
            );
```

Whatever the solution, it would be really good to have all the field value replacement in a separate method.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
doncem: i've already made a todo comment to look it over. of course some meaning exists. like "hardcoded response against actual values are compared should match" and dcn is now dynamically generated in `EnvelopeCreator.envelope`. better to look it over separately, iirc it has unused methods too

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: I was thinking about something more or less like this:

        tokens
            .keySet()
            .stream()
            .reduce(
                metadataTemplate,
               (data, token) -> data.replaceAll(token, tokens.get(token))
            );
Whatever the solution, it would be really good to have all the field value replacement in a separate method.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -149,15 +152,23 @@ public String getRandomFilename(String suffix) {
             if (metadataFile != null) {
                 String metadataTemplate =
                     Resources.toString(Resources.getResource(metadataFile), StandardCharsets.UTF_8);
-                String metadata = metadataTemplate.replace("$$zip_file_name$$", zipFilename);
+                String metadata = metadataTemplate
+                    .replace("$$zip_file_name$$", zipFilename)
+                    .replace("$$dcn1$$", generateDcnNumber())
+                    .replace("$$dcn2$$", generateDcnNumber());
```
lgonczar: This method accepts a dynamic list of file names, but then assumes there are exactly two of them. If that's always the case, maybe we could align the parameters in the method's signature with these replacements?

src/functionalTest/resources/1111006.metadata.json
```diff
@@ -9,7 +9,7 @@
   "envelope_classification": "exception",
   "scannable_items": [
       {
-        "document_control_number": "1111006",
+        "document_control_number": "$$dcn1$$",
```
lgonczar: 👍

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
lgonczar: Since this is going to be done anyway, how about solving `EnvelopeCreator.envelope` in another PR and rebasing this PR on that change?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
lgonczar: Here's some idea (not well thought through). What if we catch `ConstraintViolationException` at a lower level, closer to saving the envelope and throwing `DuplicateDocumentControlNumberException` from there? Since it implements `InvalidEnvelopeException`, no special handling would be required.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -316,6 +334,37 @@ private boolean isReadyToBeDeleted(Envelope envelope) {
         ).contains(envelope.getStatus());
     }
 
+    private void handleConstraintViolation(
+        String containerName,
+        String zipFilename,
+        String leaseId,
+        ConstraintViolationException exception
+    ) {
+        log.warn(
```
lgonczar: If we keep this method, this warning would have to go to the `if` where we reject the envelope.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
doncem: previous suggestion to use `reduce` should solve any doubts (or so i hope). wasn't thinking in such way for a while tbh hence few raw implementation attempts as you inspect

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: Oh is it? It can be again

Since extra layer before saving to db is used, can move as you suggest. Sounds good. Reduces main task tampering as well 👍

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
doncem: This change is fairly serious and was requested to be done asap. I'll mute further comments on that matter. Should not be too specific as all PRs we have but like I said - I'll mute

Making test framework PR changes as priority is a bit too eager at this point. But understandable

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: Oh wait. I missed the fact that this is a completely new exception. In this case it has to implement that interface (which is nothing but a marker) and all should be fine.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
lgonczar: Fair enough. For the time being, it would be worth adding a short comment explaining the intention behind the values.

doncem: isn't `// todo` comment enough? i'd rather review the whole change implicated here instead of keep wondering "why this or that has some values i have no idea even though i myself implemented them" ;)

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
lgonczar: Why I put the above comment here is a mistery even to me

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailedStatus.java
```diff
@@ -261,6 +267,53 @@ public void should_reject_file_that_is_not_a_valid_zip_archive() throws Exceptio
...
+            .willReturn(ImmutableMap.of(
+                "1111002.pdf", DOCUMENT_URL2
+            ));
+
+        // when
+        processor.processBlobs();
+
+        // and given
+        String filenameForDuplicate = "DUP" + SAMPLE_ZIP_FILE_NAME;
+        byte[] duplicateZipBytes = zipDir("zipcontents/duplicate_cdn_with_ok");
```
rkondratowicz: `duplicate_cdn_with_ok` is invalid because it has 2 identical dcns in the same envelope (which is going to be checked before we even try to save to DB.
We can just send the `ok` envelope under a different name.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/controller/ReadEnvelopesControllerTest.java
nitinprabhu: yeah can be done as a separate PR

PR #718, opened 2019-07-17
Fix/Update config for Jenkins pipeline(includes sonarqube config deprecations)

doncem:
### Change description ###

As from hmcts/cnp-jenkins-library#509 the constructor of `TeamNames` is different. Fixing that

Bonus: resolve sonarqube config deprecations

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,13 +124,12 @@ jacocoTestReport {
   }
 }
 
-project.tasks['sonarqube'].dependsOn test, integration
+project.tasks['sonarqube'].dependsOn jacocoTestReport
 
 sonarqube {
   properties {
     property "sonar.projectName", "Reform :: Bulk Scan Processor"
-    property "sonar.jacoco.reportPath", "${project.buildDir}/jacoco/test.exec"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -124,13 +124,12 @@ jacocoTestReport {
   }
 }
 
-project.tasks['sonarqube'].dependsOn test, integration
+project.tasks['sonarqube'].dependsOn jacocoTestReport
```
nitinprabhu: would be nice if this is a separate PR ? we like bonuses but easier for tracking later if required

doncem: i'm not seeing any benefit in separating this. very tiny change and already proven in different pr. very simple boy scout activity here. separate PRs waste time and resources

doncem: + commit log is very clear. pr is very tiny (once again)

nitinprabhu: Point is not how big or small it is those are two different changes. Although you can figure out from log but that doesn't mean two unrelated changes should go into one :) Approving for now but in future it would be helpful

PR #719, opened 2019-07-17
Use OCR validator

rkondratowicz:
created in my previous PR


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/FailedDocUploadProcessor.java
```diff
@@ -55,6 +56,7 @@ public FailedDocUploadProcessor(
         EnvelopeRepository envelopeRepository,
         ProcessEventRepository eventRepository,
         ContainerMappings containerMappings, // NOSONAR
+        OcrValidator ocrValidator, // NOSONAR
```
rkondratowicz: inheritance 😢

lgonczar: what does it say?

rkondratowicz: Note that this job (faied doc upload) does not need this validator at all, but it's injected for the compatibility reasons.
Same as `ContainerMappings` and `ServiceBusHelper`

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeControllerTest.java
```diff
@@ -101,6 +102,9 @@
     @Mock
     private DocumentManagementService documentManagementService;
 
+    @Mock
+    private OcrValidator ocrValidator;
```
lgonczar: Given OCR is a part of the metadata file, I wonder if OCR validator should be a dependency of envelope validator.

rkondratowicz: all methods in EnvelopeValidator are currently static :/

lgonczar: One reason to not make static tools, to remember.

PR #720, opened 2019-07-18
Log a warning when multiple docs contain OCR data

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-711


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -119,7 +120,12 @@ private void handleRestClientException(
                 .filter(it -> it.ocrData != null)
                 .collect(toList());
         if (docsWithOcr.size() > 1) {
-            log.warn("Multiple documents with OCR in envelope. File name: {}", envelope.zipFileName);
+            log.warn(
```
nitinprabhu: do we get alerts for warning ?

rkondratowicz: should it be an error?

nitinprabhu: yes I think as it not something which can be ignored and scanning supplier would need to fix it.

rkondratowicz: if that's the case, maybe it would be better to just reject the envelope?

nitinprabhu: that was the discussion we had and recommended :) we can continue this discussion on slack channel ?

PR #722, opened 2019-07-19
Remove setter from DB model introduced in #714

doncem:
### Change description ###

PR eventually should stand for a real replacement if approved. Can have a discussion about how to do it here or privately. I'm sure there are ways. First suggestion without any complication is provided here

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/EnvelopeCreator.java
```diff
@@ -148,11 +149,11 @@ public static Envelope envelopeNotified() {
         return envelope("SSCS", Status.NOTIFICATION_SENT);
     }
 
-    private static List<ScannableItem> scannableItems() {
+    public static List<ScannableItem> scannableItems(Iterator<String> dcnFeed) {
```
nitinprabhu: How about below rather than passing iterator reference ?

```
 private static ScannableItem scannableItem(
        String dcn,
        OcrData ocr,
        DocumentType documentType,
        String documentSubtype
    ) {
        Instant timestamp = getInstant();

        return new ScannableItem(
            dcn,
            timestamp,
            "test",
            "test",
            "return",
            timestamp,
            ocr,
            dcn + ".pdf",
            "test",
            documentType,
            documentSubtype
        );
    }

```

doncem: Let's see how it looks ;)

PR #727, opened 2019-07-24
Update url for OCR validation

rkondratowicz:

Request body to be changed in next PR


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/ocrvalidation/client/OcrValidationClient.java
```diff
@@ -4,23 +4,32 @@
...
+
     private final RestTemplate restTemplate;
 
     public OcrValidationClient(RestTemplate restTemplate) {
         this.restTemplate = restTemplate;
     }
 
-    public ValidationResponse validate(String url, FormData formData, String s2sToken) {
-
+    public ValidationResponse validate(String baseUrl, FormData formData, String s2sToken) {
```
lgonczar: Doesn't it require a change in the call, too?

rkondratowicz: currently no URLs are configured

PR #729, opened 2019-07-24
Log info when OCR validation url is not configured for service

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -104,13 +104,19 @@ private void handleRestClientException(
...
+        Optional<String> validationUrl = containerMappings
             .getMappings()
             .stream()
             .filter(mapping -> Objects.equals(mapping.getPoBox(), poBox))
             .findFirst()
             .map(mapping -> mapping.getOcrValidationUrl())
             .filter(url -> !Strings.isNullOrEmpty(url));
+
+        if (!validationUrl.isPresent()) {
+            log.info("OCR validation for po box {} not configured", poBox);
```
lgonczar: Or, alternatively, we could log it a level higher and add `... - skipping validation` to the end of the message. Up to you.

PR #730, opened 2019-07-24
Enable ocr validation for sample app

rkondratowicz:
So that functional tests call the sample app to validate OCR in envelope.

### DIFF
--- a/infrastructure/prod.tfvars
+++ b/infrastructure/prod.tfvars
@@ -19,3 +19,5 @@ blob_processing_delay_in_minutes = "30"
 smtp_host = "smtp.office365.com"
 
 incomplete_envelopes_enabled = "true"
+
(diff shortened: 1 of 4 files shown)


infrastructure/prod.tfvars
```diff
@@ -19,3 +19,5 @@ blob_processing_delay_in_minutes = "30"
 smtp_host = "smtp.office365.com"
 
 incomplete_envelopes_enabled = "true"
+
+ocr_validation_url_bulkscan_sample_app = "https://bulk-scan-sample-app-prod.service.core-compute-prod.internal"
```
aliveni: we don't have the sample app in `prod`. only in `aat`

rkondratowicz: fixed

PR #750, opened 2019-08-09
Include OCR validation warnings in queue messages

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-759

### Change description ###

Include OCR validation warnings in queue messages. This involves storing those warnings together with the envelope.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
@@ -36,7 +36,11 @@ public class EnvelopeMapper {
         // utility class
     }
 
-    public static Envelope toDbEnvelope(InputEnvelope envelope, String containerName) {
+    public static Envelope toDbEnvelope(
+        InputEnvelope envelope,
+        String containerName,
+        List<String> ocrValidationWarnings
+    ) {
         return new Envelope(
             envelope.poBox,
             envelope.jurisdiction,
@@ -47,25 +51,35 @@ public class EnvelopeMapper {
             envelope.caseNumber,
             envelope.previousServiceCaseReference,
             envelope.classification,
-            toDbScannableItems(envelope.scannableItems),
+            toDbScannableItems(envelope.scannableItems, ocrValidationWarnings),
             toDbPayments(envelope.payments),
             toDbNonScannableItems(envelope.nonScannableItems),
             containerName
         );
     }
 
-    private static List<ScannableItem> toDbScannableItems(List<InputScannableItem> scannableItems) {
+    private static List<ScannableItem> toDbScannableItems(
+        List<InputScannableItem> scannableItems,
+        List<String> ocrValidationWarnings
+    ) {
         if (scannableItems != null) {
             return scannableItems
                 .stream()
-                .map(EnvelopeMapper::toDbScannableItem)
+                .map(item ->
+                    toDbScannableItem(
+                        item,
+                        item.ocrData != null ? ocrValidationWarnings.toArray(new String[0]) : null
+                    ))
                 .collect(toList());
         } else {
             return emptyList();
         }
     }
 
-    public static ScannableItem toDbScannableItem(InputScannableItem scannableItem) {
+    public static ScannableItem toDbScannableItem(
+        InputScannableItem scannableItem,
+        String[] ocrValidationWarnings
+    ) {
         return new ScannableItem(
             scannableItem.documentControlNumber,
             scannableItem.scanningDate,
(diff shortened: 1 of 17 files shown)


src/main/resources/db/migration/V044__Add_validation_warnings_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scannable_items
+ADD COLUMN validationWarnings VARCHAR(1000)[] NULL;
```
rkondratowicz: Field in code is called `ocrValidationWarnings`.
It would be good to keep it consistent I think. The 'ocr' adds clarification, so I'd go with that.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
```diff
@@ -139,6 +146,17 @@ public String toString() {
             + "}";
     }
 
+    private List<String> retrieveOcrValidationWarnings(Envelope envelope) {
```
rkondratowicz: maybe this and `retrieveOcrData` below could use the same function to the the relevant scannable item

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
```diff
@@ -47,25 +51,35 @@ public static Envelope toDbEnvelope(InputEnvelope envelope, String containerName
...
+        List<String> ocrValidationWarnings
+    ) {
         if (scannableItems != null) {
             return scannableItems
                 .stream()
-                .map(EnvelopeMapper::toDbScannableItem)
+                .map(item ->
+                    toDbScannableItem(
+                        item,
+                        item.ocrData != null ? ocrValidationWarnings.toArray(new String[0]) : null
```
rkondratowicz: this will copy the same warnings to multiple scannable items there are more than 1 containing OCR data

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -47,8 +49,14 @@ public OcrValidator(
...
 
-    public void assertIsValid(InputEnvelope envelope) {
-        Optionals.ifAllPresent(
+    /**
+     * If required, validates the OCR data of the given envelope.
+     *
+     * @return List of warnings for valid OCR data, to be displayed to the caseworker
+     * @throws OcrValidationException if the OCR data is invalid
+     */
+    public List<String> assertIsValid(InputEnvelope envelope) {
```
doncem: How about firstly making changes in db model (or servicebus message) and on second PR - enabling this method to store **and** send warnings?

src/main/resources/db/migration/V044__Add_validation_warnings_column.sql
lgonczar: I agree it would be better like that. Story description says `validation_warnings` (we don't use underscored by convention, unfortunately). @nitinprabhu as you proposed current name, would you be ok with adding `ocr` to the beginning?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/msg/EnvelopeMsg.java
lgonczar: done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapper.java
lgonczar: solved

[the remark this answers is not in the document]
lgonczar: it's always empty now

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -61,6 +71,20 @@ public void assertIsValid(InputEnvelope envelope) {
...
+                            singletonList("OCR validation was not performed due to errors")
+                        )
+                    )
+        );
+    }
+
+    private OcrValidationWarnings ocrValidationWarnings(InputScannableItem scannableItem, List<String> warnings) {
+        return new OcrValidationWarnings(
+            scannableItem.documentControlNumber,
+            warnings != null ? warnings : emptyList()
```
doncem: so is there a test simulating `null` and `warning = empty list` situation? perhaps just missing assertions in existing tests

PR #756, opened 2019-08-14
Update Jenkins files for secret loading

doncem:
### Change description ###

Fixing deprecation around loading secrets from vault. Reference for simpler transition: hmcts/bulk-scan-ccd-event-handler-sample-app#61 (processor repo might look like a lot changed even though it's a single movement)

```
[ ] Yes
[x] No
```


Jenkinsfile_CNP
```diff
@@ -13,29 +13,34 @@ def component = "processor"
...
-  secret('test-not-yet-valid-key-store-password', 'TEST_NOT_YET_VALID_KEY_STORE_PASSWORD'),
+    // Base64-encoded PKCS12 key store containing private key and certificate known to the API, but not yet valid
+    secret('test-not-yet-valid-key-store', 'TEST_NOT_YET_VALID_KEY_STORE'),
+    secret('test-not-yet-valid-key-store-password', 'TEST_NOT_YET_VALID_KEY_STORE_PASSWORD'),
 
-  secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
+    secret('test-subscription-key', 'TEST_SUBSCRIPTION_KEY'),
 
-  secret('test-storage-account-key', 'TEST_STORAGE_ACCOUNT_KEY')
+    secret('storage-account-primary-key', 'TEST_STORAGE_ACCOUNT_KEY')
```
rkondratowicz: did the name of a secret in vault change?

Jenkinsfile_CNP
```diff
@@ -13,29 +13,34 @@ def component = "processor"
...
+  ]
 ]
 
-
-List<LinkedHashMap<String, Object>> commonSecrets = [
-  secret('test-s2s-secret', 'TEST_S2S_SECRET'),
-  secret('s2s-secret', 'S2S_SECRET'),
```
rkondratowicz: are those gone now?

Jenkinsfile_CNP
doncem: No it did not. Basically with this change/deprecation the presence of `test-config.tf` became obsolete - there is no need to add extra resources to vault with some `test-` prefix and just use them directly from vault specified in `secrets` definition in jenkins files ;)

In short: `test-storage-account-key` was just masking over the `storage-account-primary-key` from/in `bulk-scan` vault

Jenkinsfile_CNP
doncem: re-read full scope of those secret definitions please :). s2s secrets are retrieved directly from s2s vault and not from masked values inside bulk scan vault

doncem:
not yet ready. check send letter pipeline. migrating secrets with multiple vaults in use requires `vaultName` to be removed. that breaks the build on db migration step (maybe even further) because jenkins library is requesting such output presence

doncem:
Pipeline is fixed - ready to merge

PR #758, opened 2019-08-15
Reject envelope when JsonMappingException is thrown when parsing metadata file

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-769

### Change description ###

Reject envelope when JsonMappingException is thrown when parsing metadata file. This happens when custom deserialisers fail.

The affected class is completely untested - created a separate story for it: https://tools.hmcts.net/jira/browse/BPS-770

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
@@ -1,6 +1,7 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor;
 
 import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
 import com.github.fge.jsonschema.core.exceptions.ProcessingException;
 import org.hibernate.exception.ConstraintViolationException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/EnvelopeProcessor.java
```diff
@@ -64,7 +65,7 @@ public InputEnvelope parseEnvelope(
             schemaValidator.validate(metadataStream, zipFileName);
 
             return schemaValidator.parseMetafile(metadataStream);
-        } catch (JsonParseException exception) {
+        } catch (JsonParseException | JsonMappingException exception) {
```
doncem: This makes `"additionalProperties": false` feature obsolete or vice versa. Shall we remove extra flag from all occurrences in schema definition?

lgonczar: I've made the check stricter, i.e. we're only catching JsonMappingException if it represents OCR data parsing issue. Are you ok with the change?

doncem: of course ;)

PR #769, opened 2019-08-21
Validate OCR presence

rkondratowicz:
TODO: pass doc with OCR to `OcrValidator` (currently it finds the document to validate in envelope)

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrPresenceException.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrPresenceException.java
@@ -1,6 +1,6 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
 
-public class OcrPresenceException extends InvalidEnvelopeException{
+public class OcrPresenceException extends InvalidEnvelopeException {
     public OcrPresenceException(String message) {
         super(message);
     }
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrPresenceException.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrPresenceException.java
@@ -1,0 +1,7 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+public class OcrPresenceException extends InvalidEnvelopeException{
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/FailedDocUploadProcessor.java
```diff
@@ -56,6 +57,7 @@ public FailedDocUploadProcessor(
         EnvelopeRepository envelopeRepository,
         ProcessEventRepository eventRepository,
         ContainerMappings containerMappings, // NOSONAR
+        OcrPresenceValidator ocrPresenceValidator, // NOSONAR
```
rkondratowicz: can't wait to get rid of that inheritance...

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrPresenceValidator.java
```diff
@@ -0,0 +1,41 @@
...
+    public static final String MISPLACED_OCR_MSG = "OCR on document of invalid type";
+
+    /**
+     * Checks whether OCR data is on the correct document.
+     *
+     * @return Document with OCR.
+     */
+    public Optional<InputScannableItem> assertHasProperlySetOcr(List<InputScannableItem> docs) {
+
+        throwIf(docs.stream().filter(it -> it.ocrData != null).count() > 1, MULTIPLE_OCR_MSG);
```
nitinprabhu: we will need more information in the message like document control number which had the OCR and also if we can document type/subtype that would be bonus.

rkondratowicz: how about I catch, log where this is being thrown, and rethrow?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrPresenceException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+public class OcrPresenceException extends InvalidEnvelopeException {
```
lgonczar: I'd squeeze `Validation` work into this name

PR #774, opened 2019-08-23
Combine 2 sets of secrets into one

doncem:

[Update Jenkinsfiles to solve deprecated vault sets](https://tools.hmcts.net/jira/browse/BPS-775)

### Change description ###

For brainless fix: just use one set of secrets

```
[ ] Yes
[x] No
```


doncem:
Need more delicate approach on this

PR #776, opened 2019-08-25
Update Jenkins config files

doncem:

[Update Jenkinsfiles to solve deprecated vault sets](https://tools.hmcts.net/jira/browse/BPS-775)

### Change description ###

Cobination of #756 and #774 latter being slightly modified by using map merger instead

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -1,5 +1,11 @@
 #!groovy
 
+Map.metaClass.addNested = { Map rhs ->
+  def lhs = delegate
+  rhs.each { k, v -> lhs[k] = lhs[k] in Map ? lhs[k].addNested(v) : v }
+  lhs
+}
+
 @Library("Infrastructure@all-your-base")
 
 import uk.gov.hmcts.contino.GradleBuilder
@@ -100,6 +106,6 @@ withPipeline(type, product, component) {
   }
 
   onNonPR() {
-    loadVaultSecrets(commonSecrets + apiTestSecrets)
+    loadVaultSecrets(commonSecrets.addNested(apiTestSecrets))
   }
 }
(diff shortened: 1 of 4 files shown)


infrastructure/main.tf
```diff
@@ -14,7 +14,7 @@ locals {
 
   s2s_rg       = "rpe-service-auth-provider-${local.local_env}"
   s2s_url      = "http://${local.s2s_rg}.service.core-compute-${local.local_env}.internal"
-  dm_store_url = "http://dm-store-${local.local_env}.service.core-compute-${local.local_env}.internal"
+  dm_store_url = "${local.local_env == "sandbox" ? "false" : "http://dm-store-${local.local_env}.service.core-compute-${local.local_env}.internal"}"
```
nitinprabhu: we don't need this logic as iddm store is built on sandbox we will have to update this later.

Jenkinsfile_CNP
```diff
@@ -13,29 +25,34 @@ def component = "processor"
...
 
-
-List<LinkedHashMap<String, Object>> commonSecrets = [
-  secret('test-s2s-secret', 'TEST_S2S_SECRET'),
-  secret('s2s-secret', 'S2S_SECRET'),
-  secret('test-private-key-der', 'TEST_PRIVATE_KEY_DER'),
+def commonSecrets = [
+  's2s-${env}': [
+    secret('microservicekey-bulk-scan-processor-tests', 'TEST_S2S_SECRET'),
+    secret('microservicekey-bulk-scan-processor-tests', 'S2S_SECRET')
```
nitinprabhu: let's add a comment here as discussed

infrastructure/main.tf
doncem: yep. was testing out the behaviour. will revert

PR #783, opened 2019-08-27
Use deprecatred version of load secrets

nitinprabhu:


nitinprabhu:
> i don't think this was our issue to be reverted

Please check build link it failed on Aat deploy stage although passed on pr.

PR #784, opened 2019-08-28
Send heartbeat messages to orchestrator

rkondratowicz:


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/HeartbeatTask.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(HeartbeatTask.class);
+
+    private final ServiceBusHelper serviceBusHelper;
+
+    public HeartbeatTask(@Qualifier("envelopes-helper") ServiceBusHelper serviceBusHelper) {
+        this.serviceBusHelper = serviceBusHelper;
+    }
+
+    @Scheduled(cron = "0 0/10 * * * *")
```
rkondratowicz: unless we want to have different setting of different envs I think this can stay as it is.

PR #790, opened 2019-08-30
Merge secrets - not use map merger

doncem:

[Update Jenksinfiles to solve deprecated vault sets](https://tools.hmcts.net/jira/browse/BPS-775)

### Change description ###

Still not seeing the secret needed for functional test. Secret is present in vault. Will just use full blown secret list for both: PRs and master builds

```
[ ] Yes
[x] No
```


doncem:
let's hope this will solve for now #792

PR #804, opened 2019-09-04
Delete files only after envelope gets completed

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-788


rkondratowicz:
good find @doncem

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/Processor.java
```diff
@@ -66,7 +67,11 @@ protected void processParsedEnvelopeDocuments(
             return;
         }
 
-        deleteBlob(envelope, cloudBlockBlob);
+        try {
+            cloudBlockBlob.breakLease(0);
```
nitinprabhu: Any reason we are breaking the lease here ? Worth adding a comment regarding the same.

rkondratowicz: This part was being done in (now deleted) `deleteBlob(...)` method.
Keeping it as it was.

rkondratowicz:
Closing, going with alternative solution from https://github.com/hmcts/bulk-scan-processor/pull/815

PR #821, opened 2019-09-11
Remove old payment fields from input model

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-772

### DIFF
--- a/src/main/resources/metafile-schema.json
+++ b/src/main/resources/metafile-schema.json
@@ -154,164 +154,18 @@
       "items": {
         "id": "#/properties/payments/items",
         "type": "object",
-        "oneOf": [
-          {
-            "properties": {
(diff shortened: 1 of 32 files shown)


src/main/resources/metafile-schema.json
```diff
@@ -154,164 +154,18 @@
...
+            "type": "string",
+            "title": "The unique identifier for the payment",
+            "pattern": "^[0-9]+$"
           }
-        ]
+        },
+        "required": [
+          "document_control_number"
+        ],
+        "additionalProperties": true
```
rkondratowicz: so that old format is still accepted

rkondratowicz:
@nitinprabhu it's already included
`should_parse_envelope_data_with_payments`

PR #845, opened 2019-09-18
Add config for payments feature toggle

aliveni:
https://tools.hmcts.net/jira/browse/BPS-813

### Change description ###
Just adding the config in this pr. 
In the next pr, use the config to process payments.

```
[ ] Yes
[x ] No
```


infrastructure/variables.tf
```diff
@@ -180,3 +180,7 @@ variable "appinsights_instrumentation_key" {
   description = "Instrumentation key of the App Insights instance this webapp should use. Module will create own App Insights resource if this is not provided"
   default     = ""
 }
+
+variable "process_payments_enabled" {
+  default = "true"
```
nitinprabhu: may be disable it for now and turn it on when needed ?

aliveni: you mean disable by default and enable on aat and demo?
we have to enable on aat and demo for QA

doncem: There's always edgy case whether one should default to `true` or `false`

aliveni:
> You would need to update helm config too. Can be done later as the app is not using it yet :)

@nitinprabhu updated helm config

src/main/resources/application.yaml
```diff
@@ -133,6 +134,7 @@ containers:
       jurisdiction: BULKSCAN
       poBox: BULKSCANPO
       ocrValidationUrl: ${OCR_VALIDATION_URL_BULKSCAN_SAMPLE_APP}
+      payments-enabled: true
```
doncem: perhaps best to stick to common naming convention at least within same config scope even though i **do** prefer hyphens

aliveni:
> Where's the chart bump gone?

it's overridden by the other pr (merged to master) 😞 Will update shortly

PR #849, opened 2019-09-19
Bps 798 read blob from configured container

mdayican:

https://tools.hmcts.net/jira/browse/BPS-798

### Change description ###

read from configured BLOB container

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/ContainerNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+public class ContainerNotFoundException extends RuntimeException {
+
+    private static final long serialVersionUID = -3411204344653345638L;
+
+    public ContainerNotFoundException() {
```
doncem: Perhaps would be handy for us to know which container is it and then we can stick in instead of `BLOB`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
```diff
@@ -84,10 +86,22 @@ public CloudBlobContainer getContainer(String containerName) throws URISyntaxExc
...
-        return StreamSupport
+        List<CloudBlobContainer> cloudBlobContainerList = StreamSupport
             .stream(cloudBlobClient.listContainers().spliterator(), false)
             .filter(c -> !c.getName().endsWith(REJECTED_CONTAINER_NAME_SUFFIX))
+            .filter(this::filterBySelectedContainer)
             .collect(toList());
+
+        if (cloudBlobContainerList.isEmpty()) {
+            throw new ContainerNotFoundException();
+        }
```
doncem: hmm. this potentially is "risky" change. Can we return empty list instead and log an error in case not found such container if specified in config?

mdayican: changed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/ContainerNotFoundException.java
mdayican: > Perhaps would be handy for us to know which container is it and then we can stick in instead of `BLOB`?

delete the exception

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
```diff
@@ -84,10 +85,22 @@ public CloudBlobContainer getContainer(String containerName) throws URISyntaxExc
...
     public List<CloudBlobContainer> listInputContainers() {
-        return StreamSupport
+        List<CloudBlobContainer> cloudBlobContainerList = StreamSupport
             .stream(cloudBlobClient.listContainers().spliterator(), false)
             .filter(c -> !c.getName().endsWith(REJECTED_CONTAINER_NAME_SUFFIX))
+            .filter(this::filterBySelectedContainer)
             .collect(toList());
+
+        if (cloudBlobContainerList.isEmpty()) {
+            log.error("NO Blob container found, configuration for selected container: {}", properties.getBlobSelectedContainer());
```
nitinprabhu: Probably it should be "Container not found for configured container name {} ?

nitinprabhu: Also you can verify this error log in output capture ? something similar to https://github.com/hmcts/bulk-scan-processor/blob/d7c42b665edcea8a916976020290f46ad534bd87/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ErrorNotificationHandlerTest.java#L180-L185

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
```diff
@@ -84,10 +85,22 @@ public CloudBlobContainer getContainer(String containerName) throws URISyntaxExc
...
+        if (cloudBlobContainerList.isEmpty()) {
+            log.error("Container not found for configured container name : {}", properties.getBlobSelectedContainer());
+        }
+
+        return cloudBlobContainerList;
+    }
+
+    private boolean filterBySelectedContainer(CloudBlobContainer container) {
+        return SELECT_ALL_CONTAINER.equalsIgnoreCase(properties.getBlobSelectedContainer())
+            || properties.getBlobSelectedContainer().equals(container.getName());
```
nitinprabhu: `properties.getBlobSelectedContainer()` is repeated thrice could we please extract it ?

mdayican: > `properties.getBlobSelectedContainer()` is repeated thrice could we please extract it ?

changed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
mdayican: added

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
doncem: this still logs everytime. even if `ALL` is in application config

PR #861, opened 2019-09-25
AKS cleanup - Prepare to migrate APP to AKS on AAT

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-818

### Change description ###

- Moved sensible default to. `values.yaml`

- Removed s2s vault dependency and read secrets from bulk-scan vault instead.

- Copied postgres db password as a secret to `bulk-scan` vault to be used for flyway migration.

- API Management configuration is still in terraform and not moved to AKS configuration.

```
[ ] Yes
[x ]

### CODE REVIEW

File: charts/bulk-scan-processor/Chart.yaml [hunk 1]
```diff
@@ -1,6 +1,6 @@
 name: bulk-scan-processor
 home: https://github.com/hmcts/bulk-scan-processor
-version: 0.1.4
+version: 0.1.5
```
adusumillipraveen: Not using latest java chart ?

charts/bulk-scan-processor/values.aat.template.yaml
```diff
@@ -15,3 +15,24 @@ java:
     NOTIFICATIONS_TO_ORCHESTRATOR_TASK_DELAY: "3000"
     PROCESS_PAYMENTS_ENABLED: "true"
     STORAGE_BLOB_SELECTED_CONTAINER: "ALL"
+    BULK_SCANNING_DB_USER_NAME: bulk_scanner@bulk-scan-processor-aat
```
adusumillipraveen: AAT has too many values, is it not possible to template them like https://github.com/hmcts/draft-store/blob/master/charts/draft-store-service/values.yaml#L6 in values.yaml so that u dont' need to override in every environment.

charts/bulk-scan-processor/Chart.yaml
nitinprabhu: this is chart version :) I will do update Docker file and java chart version as a separate PR(might include Java 11 as well)

PR #873, opened 2019-10-02
Update flyway strategy

nitinprabhu:
### Change description ###

- Currently AKS staging container fails to start up due to `org.flywaydb.core.internal.exception.FlywaySqlException:
Unable to obtain connection from database: The server requested password-based authentication, but no password was provided.`

- FLYWAY_NOOP_STRATEGY has to be true which was set to default false.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/charts/bulk-scan-processor/values.yaml
+++ b/charts/bulk-scan-processor/values.yaml
@@ -27,7 +27,7 @@ java:
     BULK_SCANNING_DB_PORT: "5432"
     FLYWAY_URL: jdbc:postgresql://bulk-scan-processor-{{ .Values.global.environment }}.postgres.database.azure.com:5432/bulk_scan?sslmode=require
     FLYWAY_USER: bulk_scanner@bulk-scan-processor-{{ .Values.global.environment }}.postgres.database.azure.com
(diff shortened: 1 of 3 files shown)


charts/bulk-scan-processor/values.yaml
```diff
@@ -27,7 +27,7 @@ java:
     BULK_SCANNING_DB_PORT: "5432"
     FLYWAY_URL: jdbc:postgresql://bulk-scan-processor-{{ .Values.global.environment }}.postgres.database.azure.com:5432/bulk_scan?sslmode=require
     FLYWAY_USER: bulk_scanner@bulk-scan-processor-{{ .Values.global.environment }}.postgres.database.azure.com
-    FLYWAY_NOOP_STRATEGY: "false"
+    FLYWAY_NOOP_STRATEGY: "true"
```
rkondratowicz: why are we not running migrations?

nitinprabhu: for preview it will be false no db migration for other envs it will be true i.e run migration provided db migration step is present.

PR #887, opened 2019-10-07
Read the correct string to be set as write processed queue env

nitinprabhu:
### Change description ###

- Read secret value was set instead of write for processed queue envelope write string.

- In output.tfvars correct value is used hence works fine on AAT Functional tests on ASE.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -32,7 +32,7 @@ def combinedSecrets = [
     secret('storage-account-primary-key', 'TEST_STORAGE_ACCOUNT_KEY'),
     secret('test-private-key-der', 'TEST_PRIVATE_KEY_DER'),
     secret('storage-account-name', 'TEST_STORAGE_ACCOUNT_NAME'),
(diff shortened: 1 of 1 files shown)


Jenkinsfile_CNP
```diff
@@ -32,7 +32,7 @@ def combinedSecrets = [
     secret('storage-account-primary-key', 'TEST_STORAGE_ACCOUNT_KEY'),
     secret('test-private-key-der', 'TEST_PRIVATE_KEY_DER'),
     secret('storage-account-name', 'TEST_STORAGE_ACCOUNT_NAME'),
-    secret('processed-envelopes-queue-listen-connection-string', 'PROCESSED_ENVELOPES_QUEUE_WRITE_CONN_STRING')
+    secret('processed-envelopes-queue-send-connection-string', 'PROCESSED_ENVELOPES_QUEUE_WRITE_CONN_STRING')
```
timja: oops

PR #897, opened 2019-10-11
Added app insights secret loading from vault config

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-841

### Change description ###

- This PR will need https://github.com/hmcts/bulk-scan-shared-infrastructure/pull/142 to get merged as it sets secret in vault.

- It will load the secret from vault and as an env var to be used by the application.

- It will be done only in AAT/PROD as we override the key vault config in preview.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/charts/bulk-scan-processor/values.yaml
+++ b/charts/bulk-scan-processor/values.yaml
@@ -45,7 +45,7 @@ java:
         - processed-envelopes-queue-listen-connection-string
         - storage-account-name
         - storage-account-primary-key
-        - AppInsightsInstrumentationKey
+        - app-insights-instrumentation-key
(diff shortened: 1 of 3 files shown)


charts/bulk-scan-processor/values.yaml
```diff
@@ -45,6 +45,7 @@ java:
         - processed-envelopes-queue-listen-connection-string
         - storage-account-name
         - storage-account-primary-key
+        - AppInsightsInstrumentationKey
```
doncem: see, hyphens would be so nice :) (mentioned in the pr review linked in description here)

PR #907, opened 2019-10-16
Stop logging OCR-specific validation errors

lgonczar:
### Change description ###

Stop logging OCR-specific validation errors from the transformation service. Those errors might contain sensitive data. Our service doesn't care about them - it's enough to log the fact that OCR is invalid according to the validation service.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
@@ -101,7 +101,7 @@ public class OcrValidator {
                 throw new OcrValidationException(
                     "OCR validation service returned OCR-specific errors. "
                         + "Document control number: " + docWithOcr.documentControlNumber + ". "
-                        + "Envelope: " + envelope.zipFileName + ". "
+                        + "Envelope: " + envelope.zipFileName + "."
                 );
             case WARNINGS:
                 log.info(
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -99,10 +99,9 @@ private void handleValidationResponse(
         switch (res.status) {
             case ERRORS:
                 throw new OcrValidationException(
-                    "Ocr validation failed. "
-                        + "Document OCR: " + docWithOcr.documentControlNumber + ". "
-                        + "Envelope: " + envelope.zipFileName + ". "
-                        + "Errors: " + Strings.join(res.errors, ", ")
+                    "OCR validation service returned OCR-specific errors. "
+                        + "Document control number: " + docWithOcr.documentControlNumber + ". "
+                        + "Envelope: " + envelope.zipFileName + ". ");
```
lgonczar: We log this exception and also include it in event data.

The only concern that I have is that now the document provider will not know what exactly is wrong with OCR data. They will be able to check that with the service, though. Thoughts?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidatorTest.java
```diff
@@ -283,7 +283,7 @@ public void should_throw_an_exception_if_service_responded_with_error_response()
         // then
         assertThat(err)
             .isInstanceOf(OcrValidationException.class)
-            .hasMessageContaining("Error!");
+            .hasMessageContaining("OCR validation service returned OCR-specific errors");
```
rkondratowicz: might be good to also assert that it DOES NOT contain "Error!" anymore. Ideally with a comment or assertj's description.
Or change "Error!" to something like "OCR error that should not be included in exception message" ;)

lgonczar: That would be excessive - there's a whole world of stuff that we don't want in this exception.

PR #925, opened 2019-10-29
aks: add config for error notifications

rkondratowicz:


charts/bulk-scan-processor/values.yaml
```diff
@@ -32,6 +32,7 @@ java:
     DELETE_REJECTED_FILES_CRON: "0 0/10 * * * *"
     INCOMPLETE_ENVELOPES_TASK_CRON: "0 0 * * * *"
     DELETE_REJECTED_FILES_TTL: PT72H
+    ERROR_NOTIFICATIONS_ENABLED: "false" // only enabled on prod
```
rkondratowicz: to be overwritten in flux

PR #943, opened 2019-11-11
Log storage account URI

lgonczar:
### Change description ###

Log storage account URI when retrieving SAS token. This is needed for diagnosing HMCTS demo problems.

```
[ ] Yes
[x] No
```


lgonczar:
> i guess no harm is it? otherwise will have to fiddle with env vars in charts etc, which will be pain

No, it's just there to so that we can be sure which storage account the application is talking to.

PR #986, opened 2019-12-09
Allow encoded OCR data to contain any character from Base64 set

lgonczar:
### Change description ###

Allow encoded OCR data to contain any character from Base64 set. Before this change, two characters were missing (`+` and `/`)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/metafile-schema.json
+++ b/src/main/resources/metafile-schema.json
@@ -105,7 +105,7 @@
           "ocr_data": {
             "id": "#/properties/scannable_items/items/properties/ocr_data",
             "type": ["string", "null"],
-            "pattern": "^[0-9a-zA-Z=]*$"
+            "pattern": "^[0-9a-zA-Z=+/]*$"
           },
           "file_name": {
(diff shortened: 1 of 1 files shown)


lgonczar:
> Can we please add a test ?

We have tests verifying that we reject when OCR is not Base64, when it's Base64 but otherwise invalid and that we convert it properly when it's valid. If we want to be stricter than that, I think we should remove the pattern from this field rather than than check if we accept all characters from Base64 set.

src/main/resources/metafile-schema.json
```diff
@@ -105,7 +105,7 @@
           "ocr_data": {
             "id": "#/properties/scannable_items/items/properties/ocr_data",
             "type": ["string", "null"],
-            "pattern": "^[0-9a-zA-Z=]*$"
+            "pattern": "^[0-9a-zA-Z=+/]*$"
```
rkondratowicz: perhaps would be worth adding a "title" field here (or any other if there's a more appropriate one) do explain we expect base64 encoded string here.

PR #1013, opened 2020-01-09
Enable storage proxy for all environments except preview

nitinprabhu:
# Change description ###

- Going forward with IP whitelisting we would now need proxy enabled not just for demo but for all environments except preview to run smoke tests.

```
[ ] Yes
[ x] No
```


charts/bulk-scan-processor/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: bulk-scan-processor
 apiVersion: v1
 home: https://github.com/hmcts/bulk-scan-processor
-version: 0.2.16
+version: 0.2.17
```
nitinprabhu: flux PR https://github.com/hmcts/cnp-flux-config/pull/1772

PR #1127, opened 2020-04-05
Added zipDeleted parameter to query

apereverzin:

### Change description ###
Added zipDeleted parameter to the query to make it possible to exclude already deleted envelopes from the result, otherwise such envelopes will be fetched again and again.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
@@ -140,7 +140,7 @@ public class EnvelopeRepositoryTest {
         );
 
         // when
-        final List<Envelope> result = repo.findByContainerAndStatus(container1, Status.COMPLETED);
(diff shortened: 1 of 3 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeRepositoryTest.java
```diff
@@ -152,6 +152,30 @@ public void should_find_complete_envelopes_by_container() {
...
+    @Test
+    public void should_find_complete_envelopes_by_container_excluding_already_deleted() {
+        // given
+        final String container1 = "container1";
+        dbHas(
+            envelope("X", Status.COMPLETED, container1),
+            envelope("Y", Status.PROCESSED, "container2"),
+            envelope("X", Status.COMPLETED, container1, true),
+            envelope("X", Status.PROCESSED, container1),
+            envelope("Z", Status.COMPLETED, "container3")
```
rkondratowicz, line 164: perhaps it would be more clear if the other 4 would pass `false` as parameter

PR #1139, opened 2020-04-07
Added delete-complete-files properties

apereverzin:

### Change description ###
cnp-flux-config: https://github.com/hmcts/cnp-flux-config/pull/2866
Cron expression is to run this job once per day at 8pm. TBD

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -191,7 +191,7 @@ scheduling:
       ttl: ${DELETE_REJECTED_FILES_TTL}
     delete-complete-files:
       enabled: ${DELETE_COMPLETE_FILES_ENABLED}
-      cron: ${DELETE_COMPLETE_FILES_CRON}
+      cron: ${DELETE_COMPLETE_FILES_CRON:false}
 
 envelope-access:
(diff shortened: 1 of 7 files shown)


apereverzin:
Fixed

src/main/resources/application.yaml
```diff
@@ -190,7 +190,8 @@ scheduling:
       cron: ${DELETE_REJECTED_FILES_CRON}
       ttl: ${DELETE_REJECTED_FILES_TTL}
     delete-complete-files:
-      enabled: false
+      enabled: ${DELETE_COMPLETE_FILES_ENABLED:false}
```
nitinprabhu, line 193: can we remove default ?

apereverzin: having default value makes it bullet proof. other properties have default values. why do you suggest removing default?

doncem: there are some conflicts in bootstrapping those application configs in during flux deployments on different environments. the working solution is to remove defaults

apereverzin:
PR-1142

PR #1142, opened 2020-04-09
Config delete job

apereverzin:

### Change description ###
cnp-flux-config: hmcts/cnp-flux-config#2866

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -191,7 +191,7 @@ scheduling:
       ttl: ${DELETE_REJECTED_FILES_TTL}
     delete-complete-files:
       enabled: ${DELETE_COMPLETE_FILES_ENABLED}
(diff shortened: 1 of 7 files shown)


src/main/resources/application.yaml
```diff
@@ -190,7 +190,8 @@ scheduling:
       cron: ${DELETE_REJECTED_FILES_CRON}
       ttl: ${DELETE_REJECTED_FILES_TTL}
     delete-complete-files:
-      enabled: false
+      enabled: ${DELETE_COMPLETE_FILES_ENABLED}
+      cron: ${DELETE_COMPLETE_FILES_CRON:false}
```
nitinprabhu, line 194: default value false ? Also we should not need default value

apereverzin: I am just restoring the build, will fix this after the build passes

apereverzin:
PR-1157

PR #1157, opened 2020-04-14
Conf delete

apereverzin:

### Change description ###
cnp-flux-config: hmcts/cnp-flux-config#2866

```
[ ] Yes
[x] No
```


charts/bulk-scan-processor/values.aat.template.yaml
```diff
@@ -5,3 +5,4 @@ java:
   devApplicationInsightsInstrumentKey: "f666440f-43c0-4abb-9c20-2ce175a374f0"
   environment:
     STORAGE_BLOB_SELECTED_CONTAINER: "ALL"
+    DELETE_COMPLETE_FILES_CRON: "0/10 * * * * *"
```
nitinprabhu, line 8: Is a reason we need this here. This config will only apply to staging

src/main/resources/application.yaml
```diff
@@ -190,7 +190,8 @@ scheduling:
       cron: ${DELETE_REJECTED_FILES_CRON}
       ttl: ${DELETE_REJECTED_FILES_TTL}
     delete-complete-files:
-      enabled: false
+      enabled: ${DELETE_COMPLETE_FILES_ENABLED:false}
```
nitinprabhu, line 193: we should remove default value

charts/bulk-scan-processor/values.aat.template.yaml
apereverzin: Removed

charts/bulk-scan-processor/values.preview.template.yaml
```diff
@@ -37,6 +37,7 @@ java:
     PROCESS_PAYMENTS_ENABLED: "false"
     OCR_VALIDATION_URL_BULKSCAN_SAMPLE_APP: "http://bulk-scan-sample-app-aat.service.core-compute-aat.internal"
     DELETE_REJECTED_FILES_ENABLED: "false"
+    DELETE_COMPLETE_FILES_ENABLED: "false"
```
nitinprabhu, line 40: if we are running functional tests then we will need this enabled else functional test will fail

charts/bulk-scan-processor/values.aat.template.yaml
nitinprabhu: we should not have any over rides here as this is only for staging

apereverzin: It is "0 0/10 * * * *" in values.yaml

PR #1162, opened 2020-04-16
Add service to upload envelope documents

doncem:

[Processor: create document upload job](https://tools.hmcts.net/jira/browse/BPS-717)

### Change description ###

Orchestrating upload task with single service. A combination of existing functionality in 2 different places:

- Main task
- Re-upload task

```
[ ] Yes
[x] No
```


doncem:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/StuckEnvelopesService.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/HMCTS/bulk-scan-processor/pullRequest?prid=5371415&bid=17606168)

doncem:
I'm more favouring a rename now so will rename to `UploadEnvelopesService` straight away

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/StuckEnvelopesService.java
```diff
@@ -0,0 +1,205 @@
...
+        this.eventRepository = eventRepository;
+        this.documentProcessor = documentProcessor;
+        this.envelopeProcessor = envelopeProcessor;
+    }
+
+    public void reProcessEnvelopes() {
+        envelopeRepository
+            .findByStatus(CREATED)
+            .stream()
+            // can be moved to query instead. up for a change later or now
```
rkondratowicz, line 84: it's only temporary (until upload gets removed from the main job) so I think it's fine to leave it like that

doncem: yes, realised and updated the comment as well

doncem:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 21
           

Complexity increasing per file
==============================
- src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java  3
- src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsServiceTest.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/HMCTS/bulk-scan-processor/pullRequest?prid=5371415&bid=17606168)

doncem:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 17
           

Complexity increasing per file
==============================
- src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java  3
- src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsServiceTest.java  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/HMCTS/bulk-scan-processor/pullRequest?prid=5371415&bid=17606168)

doncem:
codacy is all nonsense. just saying

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,209 @@
...
+    private final BlobManager blobManager;
+    private final ZipFileProcessor zipFileProcessor;
+    private final ProcessEventRepository eventRepository;
+    private final DocumentProcessor documentProcessor;
+    private final EnvelopeProcessor envelopeProcessor;
+
+    public UploadEnvelopeDocumentsService(
+        // only process envelopes after cool down period
+        // can be removed once uploading feature is removed from main job
+        @Value("${scheduling.task.upload-documents.cool-down-minutes}") long coolDownPeriod,
```
rkondratowicz, line 67: 👍

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,209 @@
...
+                );
+
+                boolean isUploaded = resultOption
+                    .map(result -> uploadParsedZipFileName(envelope, result.getPdfs()))
+                    .orElse(false);
+
+                // only update envelope state if no errors occurred during zip file processing
+                if (resultOption.isPresent()) {
+                    envelopeProcessor.handleEvent(envelope, isUploaded ? DOC_UPLOADED : DOC_UPLOAD_FAILURE);
+                }
```
rkondratowicz, line 115: could just be
```java
if (resultOption.isPresent()) {
    boolean isUploaded = uploadParsedZipFileName(envelope, result.getPdfs()));
    envelopeProcessor.handleEvent(envelope, isUploaded ? DOC_UPLOADED : DOC_UPLOAD_FAILURE);
}
```
I think

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,209 @@
...
+            log.warn(
+                "Rejecting blob - invalid signature. File: {}, Container: {}",
+                zipFileName,
+                containerName,
+                exception
+            );
+
+            handleEventRelatedError(DOC_SIGNATURE_FAILURE, containerName, zipFileName, exception.getMessage());
+
+            blobManager.tryMoveFileToRejectedContainer(zipFileName, containerName, null);
```
rkondratowicz, line 159: is this needed here? The javadoc for the class says it's only uploading the files, not doing any validation

src/main/resources/application.yaml
```diff
@@ -174,9 +174,14 @@ scheduling:
...
   lock_at_most_for: ${SCHEDULING_LOCK_AT_MOST_FOR:PT10M} # 10 minutes in ISO-8601
   task:
+    # 1 - scan storage for new envelopes and process them
     scan:
       delay: ${SCAN_DELAY:30000} # In milliseconds
       enabled: ${SCAN_ENABLED:false}
+    # 2 - upload all documents for successfully scanned envelopes
+    upload-documents:
+      cool-down-minutes: 20 # see UploadEnvelopeDocumentsService for explanation
+    # 3 - send notification to orchestrator once all documents are uploaded
```
rkondratowicz, line 184: 👍

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
doncem: since process still involves extraction etc - need to cover this extra bit even though it is most likely not going to be breached. but let's be wild and assume blob can be tampered during the window of creating envelope and uploading pdfs :shrug: 

note. once signature feature will be removed - this section will be gone as well. so, need to keep it for the time being

lgonczar: We shouldn't do any rejections in this job. At this stage of the process envelopes have already been verified and if they've been modified since then, that shouldn't lead to a rejection, but an error, which we should be aware of, so that we could decide what to do with the envelope. It shouldn't be covered by the automated rejection process. We can prevent tampering with envelopes by restricting permissions in SAS tokens (they will prevent much more bad things). I think in this case DocSignatureFailureException should be treated just like any exception.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+        } catch (IOException exception) {
+            log.error("Failure reading zip. File: {}, Container: {}", zipFileName, containerName, exception);
+
+            handleEventRelatedError(DOC_FAILURE, containerName, zipFileName, exception.getMessage());
+        }
```
lgonczar, line 158: I think it would be best to catch Exception instead of IOException here, change the message to a more general one ("Failed to process zip..." or sth) and allow other blobs to be processed.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+
+        return Optional.empty();
+    }
+
+    private boolean uploadParsedZipFileName(Envelope envelope, List<Pdf> pdfs) {
+        try {
+            documentProcessor.uploadPdfFiles(pdfs, envelope.getScannableItems());
+
+            log.info("Uploaded pdfs. File {}, container: {}", envelope.getZipFileName(), envelope.getContainer());
```
lgonczar, line 167: Worth adding envelope ID.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
lgonczar: How about `minimumEnvelopeAge` or sth? It wasn't clear to me what `coolDownPeriod` meant.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    private void processByContainer(String containerName, List<Envelope> envelopes) {
+        log.info("Processing envelopes in {} container. Envelopes found: {}", containerName, envelopes.size());
+
+        try {
+            CloudBlobContainer blobContainer = blobManager.getContainer(containerName);
+
+            envelopes.forEach(envelope ->
+                getBlobInputStream(
```
lgonczar, line 101: When processing an envelope, a lease must be acquired on it, so that multiple processes won't upload PDFs for it.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+            containerName,
+            zipFileName,
+            event
+        );
+
+        processEvent.setReason(reason);
+        eventRepository.saveAndFlush(processEvent);
+
+        log.info(
+            "Zip {} from {} marked as {}",
```
lgonczar, line 197: It would be ideal to have enveloper ID here, as well.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+
+            envelope.setUploadFailureCount(envelope.getUploadFailureCount() + 1);
+            envelopeRepository.saveAndFlush(envelope);
+            // no need to create event. envelopeProcessor::handleEvent does that
+
+            return false;
+        }
+    }
+
+    private void handleEventRelatedError(Event event, String containerName, String zipFileName, String reason) {
```
lgonczar, line 186: Is there a way to generate event in just one place? We also use EnvelopeProcessor for that.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+                exception
+            );
+
+            handleEventRelatedError(DOC_SIGNATURE_FAILURE, containerName, zipFileName, exception.getMessage());
+
+            blobManager.tryMoveFileToRejectedContainer(zipFileName, containerName, null);
+        } catch (IOException exception) {
+            log.error("Failure reading zip. File: {}, Container: {}", zipFileName, containerName, exception);
+
+            handleEventRelatedError(DOC_FAILURE, containerName, zipFileName, exception.getMessage());
```
lgonczar, line 157: I think any failure in this job could result in `DOC_UPLOAD_FAILURE`, as uploading the docs the goal of the task.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
lgonczar, line 186: Could this be done by EnvelopeProcessor instead? It's already used for that. Currently, there's some inconsistence in handling. Whenever we've using EnvelopeProcessor, we don't save reason, but we're doing it here.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,203 @@
...
+            log.error(
+                "Failed to upload PDF files to Document Management. File: {}, Container: {}",
+                envelope.getZipFileName(),
+                envelope.getContainer(),
+                exception
+            );
+
+            envelope.setUploadFailureCount(envelope.getUploadFailureCount() + 1);
+            envelopeRepository.saveAndFlush(envelope);
+            // no need to create event. envelopeProcessor::handleEvent does that
```
lgonczar, line 180: It could be clearer if we moved that call here.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
doncem: ah yes, because it's a different job now

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
lgonczar: Btw very well thought with this parameter

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,243 @@
...
+            log.error(
+                "Failed to upload PDF files to Document Management. File: {}, Container: {}, Envelope ID: {}",
+                envelope.getZipFileName(),
+                envelope.getContainer(),
+                envelope.getId(),
+                exception
+            );
+
+            envelope.setUploadFailureCount(envelope.getUploadFailureCount() + 1);
+            Status.fromEvent(DOC_UPLOAD_FAILURE).ifPresent(envelope::setStatus);
```
rkondratowicz, line 212: what if it's not present? would that be normal?
How about
```java
envelope.setStatus(UPLOAD_FAILURE);
```
>

lgonczar: I agree, nothing better than explicit setting. We know what status to set, no need to retrieve it.

doncem: copy paste methodology used. makes more sense to set directly ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,243 @@
...
+                        getBlobInputStream(client, containerName, envelope.getZipFileName(), envelope.getId())
+                    ).flatMap(inputStream ->
+                        processInputStream(inputStream, containerName, envelope.getZipFileName(), envelope.getId())
+                    ).map(result ->
+                        uploadParsedZipFileName(envelope, result.getPdfs())
+                    ).filter(isUploaded -> isUploaded).ifPresent(voidTrue ->
+                        envelopeProcessor.handleEvent(envelope, DOC_UPLOADED)
+                    );
+
+                blobClient.ifPresent(this::breakLease);
```
lgonczar, line 117: I don't think representing this logic as a chain of transformations is the clearest way. Also, in case of an unhandled exception:
- one blob can block other blobs from being processed
- lease won't be broken

I propose that individual methods return specific values and throw exceptions instead of returning empty optionals. Then this code look like this:
```
CloudBlockBlob blobClient;

try {
    blobClient = getCloudBlockBlob(
        blobContainer,
        envelope.getZipFileName(),
        envelope.getId()
    );

    blobManager.acquireLease(blobClient, containerName, envelope.getZipFileName())
    BlobInptuStream inputStream =
        getBlobInputStream(blobClient, containerName, envelope.getZipFileName(), envelope.getId());
        
    ProcessResult result = processInputStream(inputStream, containerName, envelope.getZipFileName(), envelope.getId());
    boolean isUploaded = uploadParsedZipFileName(envelope, result.getPdfs())

    if (isUploaded) {
        envelopeProcessor.handleEvent(envelope, DOC_UPLOADED)
    }
} catch (Exception exc) {
    // log error saying we failed to upload this envelope
} finally {
    if (blobClient != null) {
        blobClient.breakLease();
    }
}
```

lgonczar: Worth checking what other people think here

doncem: unchained as suggested with tiny internal exception to handle logging messages the way it was before - keeping them separate instead of generic "i failed" message

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,219 @@
...
+            if (lease.isPresent()) {
+                BlobInputStream bis = getBlobInputStream(blobClient, containerName, zipFileName, envelopeId);
+                ZipFileProcessingResult result = processInputStream(bis, containerName, zipFileName, envelopeId);
+
+                uploadParsedZipFileName(envelope, result.getPdfs());
+
+                envelopeProcessor.handleEvent(envelope, DOC_UPLOADED);
+            }
+        } catch (UploadFailures failure) {
+            log.error(failure.getMessage(), failure.getCause());
```
lgonczar, line 96: Individual specific exceptions are great - they make it easier to find out what happened and it's good we're throwing them. On this level I would still catch `Exception` here to be sure we covered everything (now and after future modifications, possibly faulty), and log with general error message. The cause will still result in all details being available. We will see that the job failed and we'll be able to find out why - both error messages will be available in logs.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,219 @@
...
+
+    public void processByContainer(String containerName, List<Envelope> envelopes) {
+        log.info("Processing envelopes in {} container. Envelopes found: {}", containerName, envelopes.size());
+
+        try {
+            CloudBlobContainer blobContainer = blobManager.getContainer(containerName);
+
+            envelopes.forEach(envelope -> processEnvelope(blobContainer, envelope));
+        } catch (URISyntaxException | StorageException exception) {
+            log.error("Unable to get client for {} container", containerName, exception);
```
lgonczar, line 72: Not sure this log message will be accurate in all cases, as we're not just getting the container in the `try` clause. Also, it would be best to handle all exceptions, including unchecked ones, so that we can continue with other containers. How about sth like this?
```
} catch (Exception exception) {
    log.error("An error occurred when trying to upload envelope documents for container %s", containerName, exception);
}
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
doncem: I can add generic one as well - no problem

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -0,0 +1,219 @@
...
+    private void tryBreakLease(CloudBlockBlob blobClient) {
+        try {
+            blobClient.breakLease(0);
+        } catch (StorageException exception) {
+            // we will expire lease anyway. no need to escalate to error
+            log.warn("Failed to break the lease for {}", blobClient.getName(), exception);
+        }
+    }
+
+    private static class UploadFailures extends RuntimeException {
```
lgonczar, line 214: Best to follow the existing naming convention for exceptions. How about `UploadFailureException` or `FailedUploadException` or `DocumentUploadException` or sth like this?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
doncem: This will be accurate. This exception is achieved **only** when trying to get client for container. All other URI and Storage exceptions are caught individually in separate methods. However - I will add generic exception as mentioned/discussed in previous review comment ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
doncem: I knew this will be a topic ;). Therefore left for suggestion in the review. Thanks @lgonczar, will do the first option. The reason I did such name is because I would like to keep it private only for internal use. Short and simple

doncem:
Pipeline got stuck - will use fresh start

PR #1163, opened 2020-04-17
Extracted file deletion functionality out of the main job

apereverzin:

### Change description ###
Extracted file deletion functionality out of main scheduled jobs

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/EnvelopeDeletionTest.java
```diff
@@ -63,7 +63,7 @@ public void should_delete_zip_file_after_successful_ingestion() throws Exception
         filesToDeleteAfterTest.add(destZipFilename);
 
         await("file should be deleted")
-            .atMost(scanDelay + 40_000, TimeUnit.MILLISECONDS)
+            .atMost(scanDelay + 150_000, TimeUnit.MILLISECONDS)
```
lgonczar, line 66: 150 is too much for a functional test to wait. Why is this value so big?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
```diff
@@ -67,6 +68,11 @@ public void should_complete_envelope_referenced_by_queue_message() throws Except
         EnvelopeResponse updatedEnvelope = getEnvelope(zipFilename);
         assertThat(updatedEnvelope.getScannableItems()).hasSize(2);
         assertThat(updatedEnvelope.getScannableItems()).allMatch(item -> item.ocrData == null);
+
+        await("file should be deleted")
+            .atMost(scanDelay + 40_000, TimeUnit.MILLISECONDS)
```
lgonczar, line 73: `scanDelay` doesn't apply here - it's the delay between two consecutive runs of blob processing job.

Worth, by analogy to how it's done elsewhere, introducing a constant representing maximum allowed time for completed blob deletion (probably even much shorter, as deletion job runs every 10 seconds).

lgonczar: Also, just before `UUID envelopeId = getEnvelope(zipFilename).getId();` we need to make an assertion checking that file exists.

lgonczar: Having had more thoughts, we can't test the new deletion process here, as this test will only pass because of the old process deleting the file.

This assertion then duplicates the test which has been deleted. I think we shouldn't make changes to these existing tests (those verifying that file has been deleted) at this stage and do it only when deletion bit has been removed from the main job. As long as the main job is responsible for deletion, that should be covered by tests.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
```diff
@@ -19,12 +19,14 @@
...
 import static com.jayway.awaitility.Awaitility.await;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.hamcrest.Matchers.is;
 
 public class ProcessedEnvelopeMessageHandlingTest extends BaseFunctionalTest {
 
     private static final long MAX_MESSAGE_PROCESSING_TIME_MILLIS = 40_000;
     private static final long MAX_ENVELOPE_FINALISATION_TIME_MILLIS = 10_000;
+    private static final int DELETE_TIMEOUT = 20_000;
```
lgonczar, line 29: Could we name it using the same convention as the two other constants, i.e. `MAX_BLOB_DELETION_TIME_MILLIS` or sth like that?

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
```diff
@@ -54,6 +56,7 @@ public void should_complete_envelope_referenced_by_queue_message() throws Except
             .pollInterval(500, TimeUnit.MILLISECONDS)
             .until(() -> hasNotificationBeenSent(zipFilename));
 
+        assertThatCode(() -> getEnvelope(zipFilename)).doesNotThrowAnyException();
```
lgonczar, line 59: I had an actual blob in mind, not envelope in DB. Just to make sure it exists before we send message to `processed-envelopes` queue.

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
apereverzin: @lgonczar done

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
apereverzin: @lgonczar Fixed

src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/ProcessedEnvelopeMessageHandlingTest.java
```diff
@@ -49,24 +51,30 @@ public void should_complete_envelope_referenced_by_queue_message() throws Except
...
 
         EnvelopeResponse updatedEnvelope = getEnvelope(zipFilename);
         assertThat(updatedEnvelope.getScannableItems()).hasSize(2);
         assertThat(updatedEnvelope.getScannableItems()).allMatch(item -> item.ocrData == null);
+
+        await("file should be deleted")
+            .atMost(DELETE_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS)
+            .pollInterval(2, TimeUnit.SECONDS)
+            .until(() -> testHelper.storageHasFile(inputContainer, zipFilename), is(false));
     }
```
lgonczar, line 78: Sorry, I didn't notice it before, but we've remove all the assertions for `Envelope.isZipDeleted()`. We need to put it here.

lgonczar: Sorry, not here, but in an appropriate integration test.

PR #1171, opened 2020-04-21
Count summary

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1188

### Change description ###
Updated SQL query for count summary.  Updated test is failing with the old query.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
     @Query(
         nativeQuery = true,
         value = "SELECT\n"
-            + "  container,\n"
-            + "  date,\n"
+            + "  first_events.container AS container,\n"
+            + "  first_events.date AS date,\n"
             + "  count(*) AS received,\n"
-            + "  SUM(CASE WHEN event IN ('DOC_FAILURE', 'FILE_VALIDATION_FAILURE', 'DOC_SIGNATURE_FAILURE') THEN 1 ELSE 0 END) AS rejected\n"
+            + "  SUM(CASE WHEN rejection_events.id IS NOT NULL THEN 1 ELSE 0 END) AS rejected\n"
```
doncem, line 21: `rejection_events.id` could just have number and case/when/then here would become obsolete

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
...
             + "FROM (\n"
             + "  SELECT DISTINCT on (container, zipfilename)\n"
-            + "  container, zipfilename, date(createdat), event\n"
+            + "    container, zipfilename, date(createdat), event\n"
             + "  FROM process_events\n"
             + "  ORDER BY container, zipfilename, date(createdat) ASC\n"
```
doncem, line 26: we could just use `createdat` instead and use `date()` function later in the group by clause

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
...
-            + "GROUP BY container, date\n"
-            + "HAVING date = :date"
+            + "LEFT JOIN (\n"
+            + "  SELECT DISTINCT on (container, zipfilename)\n"
+            + "    id, container, zipfilename\n"
+            + "  FROM process_events\n"
+            + "  WHERE event IN ('DOC_FAILURE', 'FILE_VALIDATION_FAILURE', 'DOC_SIGNATURE_FAILURE')\n"
+            + ") AS rejection_events\n"
+            + "ON rejection_events.container = first_events.container\n"
+            + "AND rejection_events.zipfilename = first_events.zipfilename\n"
```
doncem, line 35: not minding the spacings but looks like tthere's no need to expand to left join at all. how about moving ID to first subquery and but instead of ID just using case/when/then condition on `event` field like in the `where` clause of second subquery? then it will for sure be 0s and 1s in that collumn

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
...
+            + "GROUP BY first_events.container, first_events.date\n"
+            + "HAVING date = :date\n"
```
doncem, line 37: function `date()` can simply go lastly here and all other dates will be dealt in full - keeping the correct order of events

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
lgonczar: True, wouldn't be an id, but some count column, always being one. It would work equally well. Would it be clearer? I don't know, because you'd have to check what that column is in the joined select to understand what we're summing. Let's see what others think and I'm ok with either one.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
...
             + "  count(*) AS received,\n"
-            + "  SUM(CASE WHEN event IN ('DOC_FAILURE', 'FILE_VALIDATION_FAILURE', 'DOC_SIGNATURE_FAILURE') THEN 1 ELSE 0 END) AS rejected\n"
+            + "  SUM(CASE WHEN rejection_events.id IS NOT NULL THEN 1 ELSE 0 END) AS rejected\n"
             + "FROM (\n"
             + "  SELECT DISTINCT on (container, zipfilename)\n"
-            + "  container, zipfilename, date(createdat), event\n"
+            + "    container, zipfilename, createdat\n"
             + "  FROM process_events\n"
-            + "  ORDER BY container, zipfilename, date(createdat) ASC\n"
+            + "  ORDER BY container, zipfilename, createdat ASC\n"
```
lgonczar, line 26: Why is date(...) dropped?

lgonczar: You're right, this is necessary.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -15,18 +15,26 @@
...
+            + "LEFT JOIN (\n"
+            + "  SELECT DISTINCT on (container, zipfilename)\n"
+            + "    id, container, zipfilename\n"
+            + "  FROM process_events\n"
+            + "  WHERE event IN ('DOC_FAILURE', 'FILE_VALIDATION_FAILURE', 'DOC_SIGNATURE_FAILURE')\n"
+            + ") AS rejection_events\n"
+            + "ON rejection_events.container = first_events.container\n"
+            + "AND rejection_events.zipfilename = first_events.zipfilename\n"
+            + "GROUP BY first_events.container, date(first_events.createdat)\n"
+            + "HAVING date(first_events.createdat) = :date\n"
```
lgonczar, line 37: There are two things not optimal in this bit. Partly it was like that in the previous version, but worth improving, as we're doing a full database scan in this query.
* We don't need to calculate the date in this query, at all, because we receive it in the parameter
* Instead of HAVING we should have a condition in WHERE in first_events subquery - this way the join will be far less costly. Currently, the whole table data is involved in the query and the more records we add the more of a problem this will become.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
lgonczar: Having had more thoughts, we should probably remove createdat from result of this subquery (we should still order by it) and in the main SELECT we should just include :date from the parameter.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
lgonczar: I also realised we have a missing index - for createdat.

PR #1174, opened 2020-04-21
Remove the wait check config

aliveni:
https://tools.hmcts.net/jira/browse/BPS-803

### Change description ###
Flux config PR - https://github.com/hmcts/cnp-flux-config/pull/3029

```
[ ] Yes
[x] No
```


charts/bulk-scan-processor/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 name: bulk-scan-processor
 apiVersion: v1
 home: https://github.com/hmcts/bulk-scan-processor
-version: 0.2.20
+version: 0.2.21
```
doncem, line 4: not a crucial/breaking change. we can bundle up the chart change version bump. unless pipeline actually breaks (afaik the plan was to break)

PR #1181, opened 2020-04-23
Remove ASE deployment for demo

nitinprabhu:

https://tools.hmcts.net/jira/browse/BPS-1164

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -151,6 +151,9 @@ module "bulk-scan" {
     OCR_VALIDATION_URL_BULKSCAN_SAMPLE_APP = "${var.ocr_validation_url_bulkscan_sample_app}"
     OCR_VALIDATION_URL_PROBATE             = "${var.ocr_validation_url_probate}"
 
+    NO_NEW_ENVELOPES_TASK_ENABLED     = "false"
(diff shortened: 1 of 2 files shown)


infrastructure/main.tf
```diff
@@ -151,6 +151,9 @@ module "bulk-scan" {
     OCR_VALIDATION_URL_BULKSCAN_SAMPLE_APP = "${var.ocr_validation_url_bulkscan_sample_app}"
     OCR_VALIDATION_URL_PROBATE             = "${var.ocr_validation_url_probate}"
 
+    NO_NEW_ENVELOPES_TASK_ENABLED     = "false"
+    PUBLICLAW_ENABLED                 = "true"
```
nitinprabhu, line 155: Added those app settings for backward compatibility with ASE not needed once we completely move to AKS

doncem: but those won't be applied as there's no ASE deployment present after this merge isn't it?

nitinprabhu: yup just kept it there so both master and demo are in sync and just in case we need to revert we can do it smoothly :)

PR #1185, opened 2020-04-24
Populate poBox in error notification message

lgonczar:
### Change description ###

Populate poBox in error notification message, so far always set to null.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -326,12 +326,21 @@ private void sendErrorMessageToQueue(
         );
     }
 
+    private String getPoBox(String containerName) {
+        return containerMappings
+            .getMappings()
+            .stream()
+            .filter(m -> m.getContainer().equals(containerName))
+            .findFirst()
+            .get()
```
rkondratowicz, line 335: perhaps `.map(...).orElseThrow(...)` would be safer

lgonczar: Effect is the same, in both cases we've also got enough information to find out what happened and the saught-for item is the item without which the rejection couldn't have happened. But I agree that there's some value. It will be a bit more obvious in logs.

PR #1190, opened 2020-04-27
Count summary report query

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1188

### Change description ###
After Friday's discussion this PR suggest an update to count summary report query to improve performance. Original suggestion to use

WHERE date(createdat) = :date 

does not work because it is possible that the first event for an envelope was created before the date in the query and the envelope should be included into the report for that date. But not using WHERE clause causes inefficiency of the query. The suggestion is to introduce one more query parameter 'earliest' which means the earliest possible date of the first event related to an envelope; currently it is coded as 30 days before the date of the report query. The WHERE clause looks like

WHERE date(createdat) >= :earliest AND date(createdat) <= :date

Reducing number of days (i.e. making 'earliest' closer to 'date') improves performance of the query.

This PR is just a performance improvement of the query fixing the original bug. Up to you should it be applied or not.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeCountSummaryRepositoryTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeCountSummaryRepositoryTest.java
@@ -12,6 +12,7 @@ import uk.gov.hmcts.reform.bulkscanprocessor.helper.reports.countsummary.Item;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.common.Event;
 
 import java.time.Instant;
+import java.time.LocalDate;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/EnvelopeCountSummaryRepository.java
```diff
@@ -16,25 +16,29 @@
...
             + "AND rejection_events.zipfilename = first_events.zipfilename\n"
-            + "GROUP BY first_events.container, date(first_events.createdat)\n"
-            + "HAVING date(first_events.createdat) = :date\n"
+            + "GROUP BY first_events.container, first_events.date\n"
+            + "HAVING date = :date\n"
     )
-    List<EnvelopeCountSummaryItem> getReportFor(@Param("date") LocalDate date);
+    List<EnvelopeCountSummaryItem> getReportFor(
+        @Param("earliest") LocalDate earliest,
+        @Param("date") LocalDate date
```
rkondratowicz, line 42: cryptic parameters. are those 'from' and 'to'?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/EnvelopeCountSummaryRepositoryTest.java
```diff
@@ -66,9 +70,11 @@ public void should_filter_by_date() {
             event("X", DOC_PROCESSED_NOTIFICATION_SENT),
             event("Y", DOC_PROCESSED_NOTIFICATION_SENT)
         );
+        LocalDate date = now().minusDays(1);
+        LocalDate earliest = date.atStartOfDay().minusDays(30).atZone(EUROPE_LONDON_ZONE_ID).toLocalDate();
```
rkondratowicz, line 74: is this different than `date.minusDays(30)`?

apereverzin:
Optimization is out of scope of the task

PR #1179, opened 2020-04-23
Add service for document uploads

doncem:
### JIRA link (if applicable) ###

[Create document upload job](https://tools.hmcts.net/jira/browse/BPS-717)

### Change description ###

Re-creating #1162 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1191, opened 2020-04-27
Add `upload-documents` task

doncem:

### Change description ###

Part of config and envelope bundle by container was already introduced in #1179. Moved here for fewer dependencies on service along with application config.

Covered fully with unit tests and integration test verifying the happy path

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
```diff
@@ -0,0 +1,113 @@
...
+        getTask(0).run();
+
+        // then
+        verifyNoMoreInteractions(envelopeRepository);
+
+        // and
+        verify(uploadService, times(2)).processByContainer(anyString(), anyList());
+    }
+
+    private UploadEnvelopeDocumentsTask getTask(long minimumEnvelopeAge) {
```
nitinprabhu, line 84: can we please rename when I read it was bit confusing to me ?
may be `uploadTaskWithEnvelopeAge(..)`

doncem: wasn't really too fuss about the name. especially when this is temporary. as soon as upload bits are removed from scan task - this will be redundant setup

PR #1192, opened 2020-04-27
Cron expression for delete job in aat

apereverzin:

### Change description ###
Flux: https://github.com/hmcts/cnp-flux-config/pull/3105
Made it every 10 sec

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-processor/values.aat.template.yaml
+++ b/charts/bulk-scan-processor/values.aat.template.yaml
@@ -5,4 +5,3 @@ java:
   devApplicationInsightsInstrumentKey: "f666440f-43c0-4abb-9c20-2ce175a374f0"
   environment:
     STORAGE_BLOB_SELECTED_CONTAINER: "ALL"
-    DELETE_COMPLETE_FILES_CRON: "0/10 * * * * *"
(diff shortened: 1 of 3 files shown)


charts/bulk-scan-processor/values.aat.template.yaml
```diff
@@ -5,3 +5,4 @@ java:
   devApplicationInsightsInstrumentKey: "f666440f-43c0-4abb-9c20-2ce175a374f0"
   environment:
     STORAGE_BLOB_SELECTED_CONTAINER: "ALL"
+    DELETE_COMPLETE_FILES_CRON: "0/10 * * * * *"
```
nitinprabhu, line 8: this applies only to staging pod and not to all aat pods

apereverzin: @nitinprabhu should it be in values.yaml? in this case it will also be in prod - right?

nitinprabhu: for prod we can override in flux yaml if the value needs to be different. May be worth checking with team what we need.

I would day every 10 mins on prod and other environment 10 secs ?

lgonczar: I'd make sure flux for prod says every 10 minutes before this PR gets merged.

lgonczar:
Please make sure flux change gets in first

PR #1196, opened 2020-04-28
Remove not used terraform config

rkondratowicz:


rkondratowicz:
> How about the file itself? Can we also delete it at this stage?

More changes (general resources cleanup) to come.

PR #1197, opened 2020-04-28
Refactor use of processor abstract class

doncem:

### Change description ###

Slightly related to jira ticket as will have to remove `processed` status which quite a chunky task and left it for another PR during work on this one

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -183,16 +183,15 @@ private void processZipFile(
         if (leaseId.isPresent()) {
             // Zip file will include metadata.json and collection of pdf documents
             try (ZipInputStream zis = loadIntoMemory(cloudBlockBlob, zipFilename)) {
-                registerEvent(ZIPFILE_PROCESSING_STARTED, container.getName(), zipFilename, null);
+                handleEventRelatedError(ZIPFILE_PROCESSING_STARTED, container.getName(), zipFilename, null, null);
```
rkondratowicz, line 186: `ZIPFILE_PROCESSING_STARTED` is not an error afaik

doncem: this will be updated to just use `EnvelopeProcessor::createEvent` in the following PR(s) ;). Don't want to change any extra namings/logic here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -236,7 +237,7 @@ private ZipFileProcessingResult processZipFileContent(
 
             Optional<OcrValidationWarnings> ocrValidationWarnings = this.ocrValidator.assertOcrDataIsValid(envelope);
 
-            Envelope dbEnvelope = toDbEnvelope(envelope, containerName, ocrValidationWarnings);
+            dbEnvelope = toDbEnvelope(envelope, containerName, ocrValidationWarnings);
 
             result.setEnvelope(envelopeProcessor.saveEnvelope(dbEnvelope));
```
rkondratowicz, line 242: the only thing that is used from `dbEnvelope` (which is passed to `handleEventRelatedError`) is its ID, but is it ever set in case of an error?

doncem: probably not, you are right. i'll review this

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
doncem: squashed that local change as it was not too big of a change and still related to event creations

PR #1199, opened 2020-04-28
Enable new `upload-documents` task

doncem:

### Change description ###

Enabling new task introduced in #1191. Will link this one in flux PR in a bit

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-processor/values.yaml
+++ b/charts/bulk-scan-processor/values.yaml
@@ -15,6 +15,8 @@ java:
     OCR_VALIDATION_URL_FINREM: http://finrem-cos-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
     SCAN_ENABLED: "true"
     SCAN_DELAY: "4000"
+    UPLOAD_TASK_ENABLED: "true"
+    UPLOAD_TASK_DELAY: "4000"
     REUPLOAD_ENABLED: "true"
     REUPLOAD_DELAY: "4000"
     NOTIFICATIONS_TO_ORCHESTRATOR_TASK_ENABLED: "true"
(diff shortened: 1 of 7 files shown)


charts/bulk-scan-processor/values.yaml
```diff
@@ -15,6 +15,8 @@ java:
     OCR_VALIDATION_URL_FINREM: http://finrem-cos-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
     SCAN_ENABLED: "true"
     SCAN_DELAY: "4000"
+    UPLOAD_TASK_ENABLED: "true"
+    UPLOAD_TASK_DELAY: "4000"
```
nitinprabhu, line 19: may be on PR you want less delay ?

doncem: true, haven't checked preview template ;)

doncem: hm. there's nothing there. deletion only amended to run every 10 seconds. maybe i'll just add there a delay of 2s

lgonczar: https://github.com/hmcts/cnp-flux-config/pull/3139 already in place, nice.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTask.java
```diff
@@ -35,6 +42,8 @@ public UploadEnvelopeDocumentsTask(
         this.uploadService = uploadService;
     }
 
+    @Scheduled(fixedDelayString = "${scheduling.task." + TASK_NAME + ".delay}")
+    @SchedulerLock(name = TASK_NAME) // so to not upload documents multiple times
```
lgonczar, line 46: Should the task name be the name of the reupload task until it's gone?

lgonczar:
Looks good. I have just one question about scheduler lock.

In the future, it would be good to acquire locks on blobs, so that we could have multiple instances uploading envelopes at the same time. This would be an improvement compared to current reupload task.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTask.java
lgonczar: I just realised these two jobs look for envelopes in different statuses. All good.

doncem: yep. about to submit a pr for reupload complete re-write ;)

doncem:
locks are acquired in service. but yes - perhaps best to leave for future review once everything settles. it should work as is now with shedlock it's safer for 1st version

PR #1200, opened 2020-04-28
Clean up sas token test

rkondratowicz:
Spotted while cleaning up signature stuff.
Not sure why it was there in the first place.


infrastructure/test_config.tf
```diff
@@ -20,8 +20,3 @@ resource "azurerm_key_vault_secret" "s2s_secret" {
   name         = "s2s-secret"
   value        = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 }
-
-data "azurerm_key_vault_secret" "test_private_key_der" {
-  key_vault_id = "${data.azurerm_key_vault.key_vault.id}"
-  name         = "test-private-key-der"
-}
```
doncem, line 27: this was stored possibly for retrieving sas token manually on local machine. but yes, not needed for service

PR #1202, opened 2020-04-28
Check if envelope exists after lease on blob is acquired

lgonczar:
### Change description ###

Check if envelope exists after lease on blob is acquired. Otherwise race conditions may lead to multiple processes trying to process the same file (one of them rejecting it, the other one processing successfully).

```
[ ] Yes
[x] No
```


README.md
```diff
@@ -5,7 +5,7 @@
 [![codecov](https://codecov.io/gh/hmcts/bulk-scan-processor/branch/master/graph/badge.svg)](https://codecov.io/gh/hmcts/bulk-scan-processor)
 [![Known Vulnerabilities](https://snyk.io/test/github/hmcts/bulk-scan-processor/badge.svg)](https://snyk.io/test/github/hmcts/bulk-scan-processor)
 
-## Purpose
+## Purpose.
```
lgonczar, line 8: This will go away after successful build

PR #1210, opened 2020-04-30
Dont create an event when deleting completed file fails

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1223


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -120,29 +114,7 @@ private boolean tryProcessCompleteEnvelope(CloudBlobContainer container, Envelop
 
         } catch (Exception ex) {
             log.error("Failed to process file. {}", loggingContext, ex);
-            registerDeleteFailureEvent(envelope, ex.getMessage());
```
nitinprabhu, line 123: We could have alternatively register event only when it is not a storage exception for acquiring lease

PR #1211, opened 2020-04-30
Amend tests for upload documents task

doncem:
### Change description ###

Every now and then unit test fails because envelope is not created in the past. Making sure creation date is set manually to make tests valid.

This config is temporary and all these changes will be reverted in the future

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
@@ -66,7 +66,7 @@ class UploadEnvelopeDocumentsTaskTest {
     // will verify grouping by container and throwing different error
     // so both exception branches are covered in a single test
     @Test
-    void should_do_nothing_when_failing_to_get_container_client() {
+    void should_call_service_twice_when_2_different_containers_are_present_in_the_list() {
         // given
         List<Envelope> envelopes = Arrays.asList(
             getEnvelope(CONTAINER_1, MINIMUM_ENVELOPE_AGE - 1),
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
```diff
@@ -89,25 +92,37 @@ private UploadEnvelopeDocumentsTask getTask(long minimumEnvelopeAge) {
...
+            Field field = Envelope.class.getDeclaredField("createdAt");
+            field.setAccessible(true);
+            field.set(envelope, now().minus(amountOfMinutesToSubtract, MINUTES));
+
+            return envelope;
+        } catch (NoSuchFieldException | SecurityException exception) {
+            throw new RuntimeException("Could not access 'createdAt' field in envelope", exception);
+        } catch (IllegalArgumentException | IllegalAccessException exception) {
+            throw new RuntimeException("Could not update 'createdAt' field in envelope", exception);
+        }
```
lgonczar, line 122: This can probably be just one `catch(Exception exception)`

doncem: I'm being a good boy here and explicitly defining whats wrong with what where

nitinprabhu: indeed or can throws Exception as it is private method and just used above :)

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
```diff
@@ -30,6 +32,7 @@
     private static final String CONTAINER_1 = "container-1";
     private static final String CONTAINER_2 = "container-2";
     private static final String ZIP_FILE_NAME = "zip-file-name";
+    private static final long MINIMUM_ENVELOPE_AGE = 20;
```
nitinprabhu, line 35: seconds ? may be `MINIMUM_ENVELOPE_AGE_IN_SECS` ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
doncem: Helped me out actually as I forgot set access and got specific message. Plus - I'll remove those pretty soon anyway

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
doncem: in minutes ;)

PR #1213, opened 2020-04-30
Introduce `TestStorageHelper` for storage boilerplate actions

doncem:
### Change description ###

When removing/refactoring re-upload task repeated actions cam into play so extracting this into separate helper class. We can use this in all other tests so to reduce the clutter of all this setup

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
@@ -1,0 +1,109 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.util;
+
+import com.microsoft.azure.storage.CloudStorageAccount;
+import com.microsoft.azure.storage.StorageException;
+import com.microsoft.azure.storage.blob.CloudBlobClient;
+import com.microsoft.azure.storage.blob.CloudBlobContainer;
+import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import org.testcontainers.containers.DockerComposeContainer;
+
+import java.io.File;
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+
+import static uk.gov.hmcts.reform.bulkscanprocessor.helper.DirectoryZipper.zipDir;
+
+public class TestStorageHelper {
+
+    private static TestStorageHelper INSTANCE;
+
+    public static final String CONTAINER_NAME = "bulkscan";
+    public static final String ZIP_FILE_NAME = "1_24-06-2018-00-00-00.zip";
+
+    private static DockerComposeContainer<?> dockerContainer;
+    public static CloudBlobClient cloudBlobClient;
+    private CloudBlobContainer testContainer;
+
+    private TestStorageHelper() {
+        // empty constructor
+    }
+
+    public static TestStorageHelper getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new TestStorageHelper();
+        }
+
+        return INSTANCE;
+    }
+
+    public static void createDocker() {
+        dockerContainer = new DockerComposeContainer<>(
+            new File("src/integrationTest/resources/docker-compose.yml")
+        ).withExposedService("azure-storage", 10000);
+    }
+
+    public static void startDocker() {
+        if (dockerContainer == null) {
+            throw new RuntimeException(
+                "Cannot start container. Did you create one with "
+                    + TestStorageHelper.class.getSimpleName()
+                    + "::createDocker in a static context?"
+            );
+        }
+
+        try {
+            dockerContainer.start();
+            cloudBlobClient = CloudStorageAccount
+                .parse("UseDevelopmentStorage=true")
+                .createCloudBlobClient();
+        } catch (InvalidKeyException | URISyntaxException exception) {
+            throw new RuntimeException("Unable to setup docker environment", exception);
+        }
+    }
+
+    public static void stopDocker() {
+        dockerContainer.stop();
+    }
+
+    public void createContainer() {
+        try {
+            testContainer = cloudBlobClient.getContainerReference(CONTAINER_NAME);
+            testContainer.createIfNotExists();
+        } catch (URISyntaxException | StorageException exception) {
+            throw new RuntimeException("Unable to create container", exception);
+        }
+    }
+
+    public void deleteContainer() {
+        try {
+            testContainer.deleteIfExists();
+        } catch (StorageException exception) {
+            throw new RuntimeException("Unable to delete container", exception);
+        }
+    }
+
+    public void upload(String directory) {
+        try {
+            CloudBlockBlob blockBlobReference = testContainer.getBlockBlobReference(ZIP_FILE_NAME);
+
+            // Blob need to be deleted as same blob may exists if previously uploaded blob was not deleted
+            // due to doc upload failure
+            if (blockBlobReference.exists()) {
+                blockBlobReference.breakLease(0);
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,109 @@
...
+    public static TestStorageHelper getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new TestStorageHelper();
+        }
+
+        return INSTANCE;
+    }
+
+    public static void createDocker() {
+        dockerContainer = new DockerComposeContainer<>(
```
nitinprabhu, line 40: why not use Class rule
```
@ClassRule
public static DockerComposeContainer environment =
    new DockerComposeContainer(new File("src/test/resources/compose-test.yml"))
            .withExposedService("redis_1", REDIS_PORT)
            .withExposedService("elasticsearch_1", ELASTICSEARCH_PORT);

```

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,109 @@
...
+        } catch (InvalidKeyException | URISyntaxException exception) {
+            throw new RuntimeException("Unable to setup docker environment", exception);
+        }
+    }
+
+    public static void stopDocker() {
+        dockerContainer.stop();
+    }
+
+    public void createContainer() {
```
nitinprabhu, line 68: let's specifically call this method createBulkscanContainer ?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
doncem: `ClassRule` is junit4 and we should finally make progress to junit5 instead ;). it's tricky as it needs a lot of other reviews to be done beforehand ;)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,109 @@
...
+            throw new RuntimeException(
+                "Cannot start container. Did you create one with "
+                    + TestStorageHelper.class.getSimpleName()
+                    + "::createDocker in a static context?"
+            );
+        }
+
+        try {
+            dockerContainer.start();
+            cloudBlobClient = CloudStorageAccount
```
nitinprabhu, line 56: method says start docker but doing more than just starting :)

nitinprabhu: this should be outside or method should be something else

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,109 @@
...
+import static uk.gov.hmcts.reform.bulkscanprocessor.helper.DirectoryZipper.zipDir;
+
+public class TestStorageHelper {
+
+    private static TestStorageHelper INSTANCE;
+
+    public static final String CONTAINER_NAME = "bulkscan";
+    public static final String ZIP_FILE_NAME = "1_24-06-2018-00-00-00.zip";
+
+    private static DockerComposeContainer<?> dockerContainer;
```
nitinprabhu, line 23: do we really need generics here ? can be just `DockerComposeContainer`

doncem: Just solving the warning here ;). Class is generic

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
doncem: Well. It is "StorageHelper" afterall. Could call `setUp` but it's just too generic I guess. I don't mind what it is now but if you have something specific in mind - let me know ;)

nitinprabhu: I think it's better to have method like `initialise` or `setUp` where you can club both `createDocker` and `startDocker` and get rid of below 
```
if (dockerContainer == null) {
             throw new RuntimeException(
                 "Cannot start container. Did you create one with "
                     + TestStorageHelper.class.getSimpleName()
                     + "::createDocker in a static context?"
             );
         }
```

doncem: was random suggestion ^ ;). with if statement.

plus, it's generic class. helper. setup sounds more test related. i'll rename `initialiseStorage` ;)

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.io.File;
+import java.net.URISyntaxException;
+import java.security.InvalidKeyException;
+
+import static uk.gov.hmcts.reform.bulkscanprocessor.helper.DirectoryZipper.zipDir;
+
+public class TestStorageHelper {
+
+    private static TestStorageHelper INSTANCE;
```
nitinprabhu, line 18: why can't we use proper names :) this is so generic

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
```diff
@@ -0,0 +1,106 @@
...
+        dockerContainer.start();
+    }
+
+    private static void initializeStorage() {
+        try {
+            cloudBlobClient = CloudStorageAccount
+                .parse("UseDevelopmentStorage=true")
+                .createCloudBlobClient();
+        } catch (InvalidKeyException | URISyntaxException exception) {
+            throw new RuntimeException("Unable to setup docker environment", exception);
```
nitinprabhu, line 52: docker environment ? do we need try catch here ?

doncem: try catch so to reduce exception signatures in test classes

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
doncem: i thought that's a normal approach when setting up instances of a class privately

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/util/TestStorageHelper.java
nitinprabhu: ok is this message valid `Unable to setup docker environment` ?

doncem: Now that method name is different, will replace with `unable to initialise storage` ;)

PR #1216, opened 2020-04-30
Fix logging version

mdayican:
### Change description ###

bump logging version

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -174,7 +174,8 @@ def versions = [
   junitPlatform: '1.6.2',
   mockitoJupiter: '3.3.3',
   springBoot: springBoot.class.package.implementationVersion,
-  springfoxSwagger: '2.9.2'
+  springfoxSwagger: '2.9.2',
+  reformLogging: '5.1.5'
 ]
 
 dependencyManagement {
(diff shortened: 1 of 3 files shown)


mdayican:
> this is certainly not ready for a review. there are lot's of integration tests depending on logs and those need to be reviewed. i started locally but had to abandon it
> 
> This is not as straight forward as one might believe ;)

most of the log verification tests were for notification and they are removed.
only one test failing should_do_nothing_when_failing_to_get_container_client() which was not failing in my local and seems not related to logs.

doncem:
> only one test failing should_do_nothing_when_failing_to_get_container_client() which was not failing in my local and seems not related to logs.

that's mentioned in our pull requests channel on slack and i did create a PR to address it (the name is wrong too): #1211

mdayican:
> > only one test failing should_do_nothing_when_failing_to_get_container_client() which was not failing in my local and seems not related to logs.
> 
> that's mentioned in our pull requests channel on slack and i did create a PR to address it (the name is wrong too): #1211

so there is no other failing test

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
```diff
@@ -60,27 +52,6 @@ void should_do_nothing_when_envelope_is_not_yet_ready_to_be_processed() {
         verifyNoMoreInteractions(envelopeRepository);
     }
 
-    // will verify grouping by container and throwing different error
```
mdayican, line 63: removed this test for now. fix is #1211

PR #1221, opened 2020-05-01
Added index for envelope based on container and zip file

nitinprabhu:
### Change description ###

- Thanks to @lgonczar for his analysis to find missing indexes in envelopes table.

- To start with below query gets executed every time we process an envelope.

```
   @Query("select e from Envelope e"
    )
```

```
[ ] Yes
[ x] No
```


lgonczar:
Your index is probably the most important by far. Another index we're going to need would be for the query retrieving envelopes to delete. That query filters by container, status and isDeleted. The index could be just on `isDeleted`, because there's very few envelopes with this value being false at any time. I can see CPU usage goes up to 20% every ten minutes and it's likely because of that query.

PR #1226, opened 2020-05-01
Create indexes on envelope_id foreign key where present

lgonczar:
### Change description ###

Create indexes on envelope_id foreign key in tables that contain this column

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V051__Add_foreign_key_indexes.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX scannable_items_envelope_id_idx ON scannable_items (envelope_id);
+CREATE INDEX non_scannable_items_envelope_id_idx ON non_scannable_items (envelope_id);
```
nitinprabhu, line 2: although we don't have non scannable items data currently but worth adding it now so that when it starts coming we don't have to change anything

PR #1239, opened 2020-05-06
rename fields of ProcessedEnvelope

mdayican:
### Change description ###

rename fields of ProcessedEnvelope.
orchestrator changes will go after this

```
[ ] Yes
[x] No
```


mdayican:
> it should be one of those IDs. strict XOR. because we are migrating from `id` to `envelope_id`. same goes to log messages.
> 
> don't purely count on me, verify with another reviewer

I am keeping all Ids now, than orchestrator start sending envelope_id  and come back here and delete id

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/msg/ProcessedEnvelope.java
```diff
@@ -9,16 +9,22 @@
...
+        @JsonProperty("envelope_id") UUID envelopeId,
+        @JsonProperty("ccd_id") Long ccdId,
+        @JsonProperty("envelope_ccd_action") String envelopeCcdAction
     ) {
         this.id = id;
-        this.processedCcdReference = processedCcdReference;
-        this.processedCcdType = processedCcdType;
+        this.envelopeId = envelopeId;
+        this.ccdId = ccdId;
+        this.envelopeCcdAction = envelopeCcdAction;
```
lgonczar, line 28: Shouldn't `processedCcdReference` map to the same field as `ccdId` and `processedCcdType` to the same field as `envelopeCcdAction`? This way we could still reference the same fields, the change would be minimal. Once the orchestrator starts sending the new fields, we'd just need to remove the old ones.

mdayican: processor does not do anything with new fields.

lgonczar:
> it should be one of those IDs. strict XOR. because we are migrating from `id` to `envelope_id`. same goes to log messages.
> 
> don't purely count on me, verify with another reviewer

As just discussed with Mustafa, orchestrator will always include `id` in the message, as long as the processor uses it. The two other fields aren't used yet, anyway.

PR #1240, opened 2020-05-06
Fix the handling of rejected files

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1206

### Change description ###

```
[ ] Yes
[x] No
```


lgonczar:
In this case there's no need for any retries on the queue level - we just need to fail the process when we can't send a message.

apereverzin:
Not necessary

PR #1241, opened 2020-05-07
Handling sending message error

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1206

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -323,22 +324,38 @@ private void handleInvalidFileError(
...
+            ErrorMapping.getFor(cause.getClass()),
+            (errorCode) -> trySendErrorMessageToQueueAndMoveFileToRejectedContainer(
+                zipFilename,
+                containerName,
+                cause, eventId,
+                leaseId,
+                errorCode
+            ),
+            () -> blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId)
+        );
```
lgonczar, line 337: It should be more sth like this:
```
Optionals.ifPresentOrElse(
...
{
   send zip file
   move to rejected container when no error so far
},
() -> throw exception because we don't have necessary error code mapping (this is an existing bug)
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -323,22 +324,40 @@ private void handleInvalidFileError(
...
+        try {
+            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+            blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId);
+        } catch (Exception exc) {
+            log.error(
+                "Error sending notification to the queue."
+                    + "File name: " + zipFilename + " "
+                    + "Container: " + containerName,
+                exc
+            );
```
lgonczar, line 359: All we need to do here is not catch the exception and let the process fail.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -323,22 +324,41 @@ private void handleInvalidFileError(
...
+            (errorCode) -> trySendErrorMessageAndMoveFileToRejectedContainer(
+                zipFilename,
+                containerName,
+                cause, eventId,
+                leaseId,
+                errorCode
+            ),
+            () -> {
+                throw new ConfigurationException("Error code mapping not found for " + cause.getClass().getName());
+            }
```
rkondratowicz, line 338: unrelated

apereverzin: This is what Leszek asked to change

lgonczar: I asked for this after some changes were made in this PR, resulting in the lack of failure in case of misconfiguration.

All we need for this story, is not catching the exception.

apereverzin: ok, I will revert this change

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -308,22 +309,39 @@ private void handleInvalidFileError(
...
+        Optionals.ifPresentOrElse(
+            ErrorMapping.getFor(cause.getClass()),
+            (errorCode) -> trySendErrorMessageAndMoveFileToRejectedContainer(
+                zipFilename,
+                containerName,
+                cause, eventId,
+                leaseId,
+                errorCode
+            ),
+            () -> blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId)
```
lgonczar, line 321: > I asked for this after some changes were made in this PR, resulting in the lack of failure in case of misconfiguration.
Sorry, I was wrong - the problem was there even before your changes. We should throw an exception here. The most important change in this PR is to not catch exceptions thrown when we fail to send message.

apereverzin: Oh! I have thought you meant to raise another task for this

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -308,22 +309,41 @@ private void handleInvalidFileError(
...
+            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+            blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId);
+        } catch (Exception exc) {
+            log.error(
+                "Error sending notification to the queue."
+                    + "File name: " + zipFilename + " "
+                    + "Container: " + containerName,
+                exc
+            );
+            throw exc;
```
lgonczar, line 345: I missed this change.

We should either not catch this exception at all or throw a new exception with our custom message. We can't log an error and then throw an exception, because that will result in two error log entries in App Insights, representing the same thing.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -308,22 +310,38 @@ private void handleInvalidFileError(
...
+        String leaseId,
+        ErrorCode errorCode
+    ) {
+        try {
+            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+            blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId);
+        } catch (Exception exc) {
+            final String msg = "Error sending notification to the queue."
+                + "File name: " + zipFilename + " "
+                + "Container: " + containerName;
```
lgonczar, line 342: The message is about sending notification to the queue, but the try part is about more than that. Perhaps we should remove the call to blobManager from this method.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -308,22 +310,38 @@ private void handleInvalidFileError(
...
+                errorCode
+            ),
+            () -> {
+                throw new ConfigurationException("Error code mapping not found for " + cause.getClass().getName());
+            }
+        );
+    }
 
-        blobManager.tryMoveFileToRejectedContainer(zipFilename, containerName, leaseId);
+    private void trySendErrorMessageAndMoveFileToRejectedContainer(
```
lgonczar, line 328: Now that this method throws exception on failure, best to remove `try` prefix. And if we remove the call to blob manager from here, ale the name part related with that bit.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForDisabledService.java
```diff
@@ -75,4 +81,36 @@ public void should_reject_file_when_service_is_disabled() throws Exception {
         errorWasSent(SAMPLE_ZIP_FILE_NAME, ErrorCode.ERR_SERVICE_DISABLED);
     }
 
+    @Test
+    public void should_not_move_file_if_send_message_throws_exception_when_service_is_disabled() throws Exception {
```
lgonczar, line 85: How about something more general, e.g. `should_abandon_rejecting_file_when_sending_error_notification_fails`? The reasons of why we're trying to reject the file is irrelevant here.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForDisabledService.java
```diff
@@ -75,4 +81,36 @@ public void should_reject_file_when_service_is_disabled() throws Exception {
...
+        // given
+        byte[] zipBytes = zipDir("zipcontents/ok");
```
lgonczar, line 87: Could we maybe use an invalid file instead of using the assumption that the service is disabled? The test code doesn't make the service disabled - it relies on some external condition. It would be best if the test itself could say "look, this is the bit that will cause the rejection".

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForDisabledService.java
apereverzin: Done. And new test class created

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForDisabledService.java
apereverzin: Created new test class BlobProcessorTaskTestForFailingNotification

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
apereverzin: Done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
apereverzin: Done

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -304,24 +306,40 @@ private void handleInvalidFileError(
...
-
+        Optionals.ifPresentOrElse(
+            ErrorMapping.getFor(cause.getClass()),
+            (errorCode) -> sendErrorMessage(
+                zipFilename,
+                containerName,
+                cause, eventId,
+                leaseId,
+                errorCode
+            ),
```
lgonczar, line 317: Suggestion: it would be best to put `blobManager.tryMoveFileToRejectedContainer(...` here. The code will still work as it is, but it would be clearer that we only proceed with both actions if the mapping is there.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailingNotification.java
```diff
@@ -0,0 +1,67 @@
...
+import static uk.gov.hmcts.reform.bulkscanprocessor.model.common.Event.ZIPFILE_PROCESSING_STARTED;
+
+@IntegrationTest
+@RunWith(SpringRunner.class)
+@TestPropertySource(properties = {
+    "containers.mappings[0].container=bulkscan",
+    "containers.mappings[0].jurisdiction=BULKSCAN",
+    "containers.mappings[0].poBox=BULKSCANPO",
+    "containers.mappings[0].paymentsEnabled=false",
+    "containers.mappings[0].enabled=false"
```
lgonczar, line 28: These properties seem irrelevant to what we're testing here. Can they be removed?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailingNotification.java
```diff
@@ -0,0 +1,67 @@
...
+            zipFileProcessor,
+            containerMappings,
+            ocrValidator,
+            serviceBusHelper,
+            paymentsEnabled
+        );
+    }
+
+    @Test
+    public void should_abandon_rejected_file_when_sending_error_notification_fails() throws Exception {
```
lgonczar, line 48: The file is not rejected yet. How about `should_abandon_processing_invalid_file...`?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailingNotification.java
```diff
@@ -0,0 +1,67 @@
...
+@IntegrationTest
+@RunWith(SpringRunner.class)
+@TestPropertySource(properties = {
+    "containers.mappings[0].container=bulkscan",
+    "containers.mappings[0].jurisdiction=BULKSCAN",
+    "containers.mappings[0].poBox=BULKSCANPO",
+    "containers.mappings[0].paymentsEnabled=false",
+    "containers.mappings[0].enabled=false"
+})
+public class BlobProcessorTaskTestForFailingNotification extends ProcessorTestSuite<BlobProcessorTask> {
```
lgonczar, line 30: By convention, `Test` should go to the end of the test class name.

apereverzin: There are quite a lot of test classes in this package named BlobProcessorTaskTestFor..., and this approach makes sense in my opinion

doncem: we have few of those. i guess shouldn't be harmful here either

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailingNotification.java
apereverzin: ok

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForFailingNotification.java
lgonczar: Aaa, I see, and all of them are for BlobProcessorTask. Fair enough. Might need to rename them all on one go later.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -304,22 +306,40 @@ private void handleInvalidFileError(
...
-                    );
-                }
-            });
+        Optionals.ifPresentOrElse(
+            ErrorMapping.getFor(cause.getClass()),
+            (errorCode) -> {
+                sendErrorMessage(
+                    zipFilename,
+                    containerName,
+                    cause, eventId,
```
lgonczar, line 315: Rafał won't like this line

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -304,22 +306,40 @@ private void handleInvalidFileError(
...
+        String containerName,
+        Exception cause,
+        Long eventId,
+        String leaseId,
+        ErrorCode errorCode
+    ) {
+        try {
+            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+        } catch (Exception exc) {
+            final String msg = "Error sending notification to the queue."
```
lgonczar, line 338: Sorry to be requesting so many changes for this PR. It would be best for this message to be sth like `Error sending error notification...`, because we have notification by which we understand the ones we send to the orchestrator.

apereverzin: No worries at all.

PR #1242, opened 2020-05-07
add new ccd columns to envelope table

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1191

### Change description ###

add new ccd columns to envelope table
ccdId is long but we prefer to save as varchar. ?!?

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V053__Add_ccd_id_and_action_to_envelopes.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE envelopes
+  ADD COLUMN ccdId VARCHAR(50) NULL,
```
apereverzin, line 2: If the value is long, shouldn't the column type be bigint?

mdayican: that is the question, Leszek recommended it and I agreed it might be better.

rkondratowicz: it's an external ID, safer to have it as varchar

src/main/resources/db/migration/V053__Add_ccd_id_and_action_to_envelopes.sql
```diff
@@ -0,0 +1,3 @@
...
+  ADD COLUMN envelopeCcdAction VARCHAR(50) NULL;
```
rkondratowicz, line 3: new line character please

PR #1245, opened 2020-05-10
Add an assertion checking the body of envelopes endpoint response

lgonczar:
### Change description ###

In functional/smoke tests, add an assertion checking the body of envelopes endpoint response. This will help see what's wrong with the response in case the status is different than expected.

Also, changed the name in SyntheticTest-Source header to be accurate.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
+++ b/src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
@@ -41,6 +41,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 public class TestHelper {
 
+    private static final String TEST_SOURCE_NAME = "Bulk Scan Processor tests";
     private static final Random RANDOM = new Random();
 
     public String s2sSignIn(String s2sName, String s2sSecret, String s2sUrl) {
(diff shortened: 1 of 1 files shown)


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/TestHelper.java
```diff
@@ -113,7 +115,7 @@ public String getSasToken(String containerName, String testUrl) throws Exception
             .relaxedHTTPSValidation()
             .baseUri(testUrl)
             .header(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
-            .header(SyntheticHeaders.SYNTHETIC_TEST_SOURCE, "Bulk Scan Processor functional test")
+            .header(SyntheticHeaders.SYNTHETIC_TEST_SOURCE, TEST_SOURCE_NAME)
```
mdayican, line 118: can we add method name or sth defines the test so when it fails we can track it easily
like TEST_SOURCE_NAME + " s2sSignIn"

lgonczar: We'll find the value from this header in AppInsights requests, along with information about which endpoints were hit.

PR #1250, opened 2020-05-11
start saving ccd id and ccd action to db

mdayican:
### Change description ###

start saving ccd id and ccd action to db.
id is replaced with envelope_id. orchestrator has already sending envelope_id since friday.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1251, opened 2020-05-11
change type

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1191

### Change description ###

changing  type of ccdID and will change the type in orchestrator.
and changing the json field mapping for now so when it receives long type "ccd_id" it can ignore.
and at the end this will merge with https://github.com/hmcts/bulk-scan-processor/pull/1250

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/msg/ProcessedEnvelope.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/msg/ProcessedEnvelope.java
@@ -12,14 +12,14 @@ public class ProcessedEnvelope {
 
     public final UUID envelopeId;
 
-    public final Long ccdId;
+    public final String ccdId;
 
     public final String envelopeCcdAction;
 
     public ProcessedEnvelope(
         @JsonProperty("id") UUID id,
         @JsonProperty("envelope_id") UUID envelopeId,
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/in/msg/ProcessedEnvelope.java
```diff
@@ -12,14 +12,14 @@
...
-    public final Long ccdId;
+    public final String ccdId;
 
     public final String envelopeCcdAction;
 
     public ProcessedEnvelope(
         @JsonProperty("id") UUID id,
         @JsonProperty("envelope_id") UUID envelopeId,
-        @JsonProperty("ccd_id") Long ccdId,
+        @JsonProperty("xx_ccd_id") String ccdId,
```
rkondratowicz, line 22: is that prefix intentional?

mdayican: yes. as orchestrator sending this field it may cause conversion problem

mdayican: unknown field are ignored

PR #1255, opened 2020-05-11
Create basic end-to-end test verifying exception envelope processing

lgonczar:

https://tools.hmcts.net/jira/browse/BPS-1198
https://tools.hmcts.net/jira/browse/BPS-1194

### Change description ###

Create basic end-to-end test verifying exception envelope processing. This test will be extended later (exact scope to be decided). For now it's just serving as an example end-to-end test.

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/BaseFunctionalTest.java
```diff
@@ -85,16 +87,32 @@ public void setUp() throws Exception {
...
+            inputContainer,
+            files,
+            metadataFile,
+            destZipFilename,
+            operationContext
+        );
+    }
+
+    protected EnvelopeResponse waitForEnvelopeToBeInStatus(String fileName, List<Status> awaitedStatuses) {
+        String s2sToken = testHelper.s2sSignIn(this.s2sName, this.s2sSecret, this.s2sUrl);
```
nitinprabhu, line 102: probably this should be done only once so we can reuse. Also another thing is intermittently we do get 503 from S2S so worth doing it once ? Not strictly for this PR may be

lgonczar: Good idea, but best to do it as a separate change. Currently, a test retrieve the token only once. To improve this, we'd need to introduce some sort of global cache.

PR #1271, opened 2020-05-18
Format envelopes query

rkondratowicz:
In preparation for upcoming change to this query.

Formatting, so that each column is in separate line.
(then it will also be easier to see what changed)

https://tools.hmcts.net/jira/browse/BPS-1240


rkondratowicz:
build passed, github not refreshed

PR #1278, opened 2020-05-19
Log loading zip to memory

rkondratowicz:
start, finish, file size


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -215,8 +216,14 @@ private void processZipFile(
     }
 
     private ZipInputStream loadIntoMemory(CloudBlockBlob cloudBlockBlob, String zipFilename) throws StorageException {
+        log.info("Loading {} into memory.", zipFilename);
```
lgonczar, line 219: How about `Loading file {} into memory.`? Easier to search.

lgonczar: Same with the other log

rkondratowicz: fixed

PR #1287, opened 2020-05-21
Return envelope's CCD id and action from `/envelopes` endpoint

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-1240


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/EnvelopeResponse.java
```diff
@@ -79,7 +85,9 @@ public EnvelopeResponse(
         @JsonProperty("classification") String classification,
         @JsonProperty("scannable_items") List<ScannableItemResponse> scannableItems,
```
nitinprabhu, line 86: Although not part of this PR but I realised that `ScannableItemResponse` is leaking OCR data which we should not as sometimes OCR data may not be cleared if the processed notification is not received

PR #1295, opened 2020-05-26
Improve test resources navigation

rkondratowicz:
Remove constant and string concatenation so that it's possible to navigate to loaded resource file directly in IntelliJ


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/model/mapper/EnvelopeMapperTest.java
```diff
@@ -32,11 +32,10 @@
 import static uk.gov.hmcts.reform.bulkscanprocessor.model.common.DocumentType.OTHER;
 
 public class EnvelopeMapperTest {
-    private static final String METAFILE_BASE_PATH = "/metafiles/valid/";
 
     @Test
     public void should_map_zip_envelope_correctly() throws Exception {
-        InputEnvelope zipEnvelope = getEnvelopeFromMetafile(METAFILE_BASE_PATH + "with-scannable-items.json");
+        InputEnvelope zipEnvelope = getEnvelopeFromMetafile("/metafiles/valid/with-scannable-items.json");
```
rkondratowicz, line 38: you can now cmd+click on that string to open the json file.

PR #1307, opened 2020-05-28
Filter out failed envelopes in sql query

rkondratowicz:
instead of doing that in code.


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/UploadEnvelopeDocumentsTaskTest.java
```diff
@@ -84,28 +81,6 @@ void should_call_service_twice_when_2_different_containers_are_present_in_the_li
         verify(uploadService, times(2)).processByContainer(anyString(), anyList());
```
apereverzin, line 81: Better should be `times(envelopes .size())` instead of `times(2)`

rkondratowicz: It's not about number of envelopes, it's number of containers.

PR #1311, opened 2020-05-28
Remove obsolete reupload properties

rkondratowicz:
Flux changes will follow.

https://tools.hmcts.net/jira/browse/BPS-718

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -82,10 +82,8 @@ module "bulk-scan" {
     S2S_NAME   = "${var.s2s_name}"
     S2S_SECRET = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 
-    REUPLOAD_BATCH     = "${var.reupload_batch}"
-    REUPLOAD_DELAY     = "${var.reupload_delay}"
     REUPLOAD_MAX_TRIES = "${var.reupload_max_tries}"
-    REUPLOAD_ENABLED   = "${var.reupload_enabled}"
+
     SCAN_DELAY         = "${var.scan_delay}"
     SCAN_ENABLED       = "${var.scan_enabled}"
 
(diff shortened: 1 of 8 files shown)


infrastructure/main.tf
```diff
@@ -82,10 +82,8 @@ module "bulk-scan" {
     S2S_NAME   = "${var.s2s_name}"
     S2S_SECRET = "${data.azurerm_key_vault_secret.s2s_secret.value}"
 
-    REUPLOAD_BATCH     = "${var.reupload_batch}"
-    REUPLOAD_DELAY     = "${var.reupload_delay}"
     REUPLOAD_MAX_TRIES = "${var.reupload_max_tries}"
```
rkondratowicz, line 85: this one stays, I will rename it separately (to match the job name)

doncem: it is not used at all - can be removed without regret

src/main/resources/application.yaml
```diff
@@ -181,9 +181,6 @@ scheduling:
       enabled: ${DELETE_COMPLETE_FILES_ENABLED}
       cron: ${DELETE_COMPLETE_FILES_CRON}
     reupload:
-      batch: ${REUPLOAD_BATCH:20} # fetch 20 envelopes at once to process them. 0 for all
-      delay: ${REUPLOAD_DELAY:1800000} # In milliseconds, 30min
-      enabled: ${REUPLOAD_ENABLED:false}
       max_tries: ${REUPLOAD_MAX_TRIES:5}
```
doncem, line 184: i guess it will be merged to `upload` section as there is no `reupload` task anymore

rkondratowicz: yup, I guess my comment on `main.tf` should be here instead

PR #1313, opened 2020-05-28
Added indexes on envelopes

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1227

### Change description ###
Please, refer https://tools.hmcts.net/confluence/display/RBS/SQL+queries+for+envelopes+table for detailed description

```
[ ] Yes
[x] No
```


src/main/resources/db/migration/V057__Add_createdat_date_index_on_envelopes.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX envelopes_createdat_status_idx ON envelopes((createdat::DATE));
```
rkondratowicz, line 1: should `status` be in the index name?

apereverzin: Fixed

src/main/resources/db/migration/V057__Add_createdat_date_index_on_envelopes.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX envelopes_createdat_idx ON envelopes((createdat::DATE));
```
doncem, line 1: probably don't need extra pair of braces

apereverzin: there are 2 options: ON envelopes((createdat::DATE)) or ON envelopes(date(createdat))
I prefer the first one

doncem: ah. ok, makes sense

apereverzin:
Removed index on status, in my local tests it gives some not very significant performance improvement compared to using just index on jurisdiction AND status

doncem:
> Removed index on status, in my local tests it gives some not very significant performance improvement compared to using just index on jurisdiction AND status

perhaps we need to review queries we use and after reviewing them make a decision what's best for db to include certain indices

src/main/resources/db/migration/V054__Add_jurisdiction_status_index_on_envelopes.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX envelopes_jurisdiction_status_idx ON envelopes(jurisdiction, status);
```
rkondratowicz, line 1: which queries will use this index?

apereverzin: Please, refer https://tools.hmcts.net/confluence/display/RBS/SQL+queries+for+envelopes+table

apereverzin:
PR-1324 due to conflict in flyway script names

PR #1314, opened 2020-05-29
Rename max upload retries property

rkondratowicz:
https://tools.hmcts.net/jira/browse/BPS-718

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -172,7 +172,7 @@ scheduling:
     upload-documents:
       delay: ${UPLOAD_TASK_DELAY} # In milliseconds
       enabled: ${UPLOAD_TASK_ENABLED}
-      max_tries: ${UPLOAD_MAX_TRIES}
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yaml
```diff
@@ -172,6 +172,7 @@ scheduling:
     upload-documents:
       delay: ${UPLOAD_TASK_DELAY} # In milliseconds
       enabled: ${UPLOAD_TASK_ENABLED}
+      max_tries: ${UPLOAD_MAX_TRIES}
```
doncem, line 175: perhaps `max-retries` makes more sense does it not?

rkondratowicz: Well, it depends.
Currently this value is set to 5. I assume the intent was to try 5 times.
Not fail once and then retry 5 times, ie 6 attempts in total.

nitinprabhu: Not sure if max tries is the correct word ? may be max_upload_attempts?

PR #1323, opened 2020-06-02
Acquire lease before blob deletion

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1224

### Change description ###
Before deleting each blob (both processed and rejected), bulk-scan-processor should acquire lease on that blob, so that no other process can work on that file.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,13 +56,22 @@ public void run() {
...
-                                blob.delete(INCLUDE_SNAPSHOTS, null, null, null);
-                                log.info("Deleted rejected file {}", blob.getName());
+                                Optional<String> leaseId = blobManager.acquireLease(
+                                    blob,
+                                    container.getName(),
+                                    zipFileName
+                                );
+                                if (leaseId.isPresent()) {
+                                    AccessCondition accessCondition = generateLeaseCondition(leaseId.get());
+                                    blob.delete(INCLUDE_SNAPSHOTS, accessCondition, null, null);
```
doncem, line 72: in v12 the delete function calls another one with all 4 arguments passed. last one being `Context.NONE`. is it same/similar with this one?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -93,11 +97,24 @@ private boolean tryProcessCompleteEnvelope(CloudBlobContainer container, Envelop
...
+                    container.getName(),
+                    envelope.getZipFileName()
+                );
+                if (leaseId.isPresent()) {
+                    AccessCondition accessCondition = generateLeaseCondition(leaseId.get());
+                    deleted = cloudBlockBlob.deleteIfExists(
+                        NONE,
+                        accessCondition,
+                        null,
+                        null
```
doncem, line 111: same comment here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
apereverzin: The point is to use the access condition. CloudBlockBlob::delete(<no arguments>) calls internally delete(...) with 4 parameters with NULL access condition.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTaskTest.java
```diff
@@ -52,16 +54,19 @@ public void setUp() {
     public void should_mark_as_deleted_complete_envelope() throws Exception {
         // given
         final String containerName1 = "container1";
+        final String leaseId = "LEASEID";
```
doncem, line 57: Maybe we can capture this lease id so to verify it was actually used when deleting?

-- edit --

oh this is integration testing. erm. capturing can be moved to unit test then ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -93,11 +97,24 @@ private boolean tryProcessCompleteEnvelope(CloudBlobContainer container, Envelop
             CloudBlockBlob cloudBlockBlob = container.getBlockBlobReference(envelope.getZipFileName());
 
             if (cloudBlockBlob.exists()) {
-                deleted = cloudBlockBlob.deleteIfExists();
-                if (deleted) {
-                    log.info("File deleted. {}", loggingContext);
-                } else {
-                    log.info("File has not been deleted. {}", loggingContext);
+                Optional<String> leaseId = blobManager.acquireLease(
```
mdayican, line 100: do we need to add check metadata to delete ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,13 +56,22 @@ public void run() {
...
+                                    log.info("Deleted rejected file {}", blob.getName());
```
lgonczar, line 73: Worth adding container name, as the envelope is identified by both.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
apereverzin: these files are already in COMPLETED state so all lease expirations (if any) have been handled

mdayican: Meta data can stay there even it is completed. I think you do not need to check metadata.

apereverzin: This is what I have meant

mdayican: blobManager.acquireLease( ) method checks metadata?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,13 +56,26 @@ public void run() {
                     .listBlobs(null, true, EnumSet.of(SNAPSHOTS), null, null)
                     .forEach(listItem -> {
                         try {
+                            final String zipFileName = FilenameUtils.getName(listItem.getUri().toString());
                             CloudBlockBlob blob = container.getBlockBlobReference(
-                                FilenameUtils.getName(listItem.getUri().toString())
+                                zipFileName
                             );
 
                             if (canBeDeleted(blob)) {
```
mdayican, line 64: do we need this `canBeDeleted` ?

doncem: looks like no longer needed :). happy about it as it should then reduce this crazy level of nested ifs

apereverzin: this will change business logic - canBeDeleted method checks if enough time passed after rejection.

apereverzin: and I slightly refactored the code to reduce nesting

lgonczar: It is required - rejected files need to stay there for a while before we delete them.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
apereverzin: Done

PR #1329, opened 2020-06-04
Log DCNs

rkondratowicz:
Right after metadata gets successfully parsed.
https://tools.hmcts.net/jira/browse/BPS-1207

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
@@ -49,6 +49,7 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.zip.ZipInputStream;
 
+import static java.util.stream.Collectors.joining;
 import static org.apache.commons.io.IOUtils.toByteArray;
 import static uk.gov.hmcts.reform.bulkscanprocessor.model.common.Event.ZIPFILE_PROCESSING_STARTED;
 import static uk.gov.hmcts.reform.bulkscanprocessor.model.mapper.EnvelopeMapper.toDbEnvelope;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -241,6 +242,12 @@ private void processZipFileContent(
 
             InputEnvelope envelope = envelopeProcessor.parseEnvelope(result.getMetadata(), zipFilename);
 
+            log.info(
+                "Parsed envelope. Payment DCNs: {}. Document DCNs: {}",
```
nitinprabhu, line 246: worth including zip file name and jurisdiction here ?

rkondratowicz: oops. fixed.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -241,6 +242,14 @@ private void processZipFileContent(
 
             InputEnvelope envelope = envelopeProcessor.parseEnvelope(result.getMetadata(), zipFilename);
 
+            log.info(
+                "Parsed envelope. File name: {}. Container: {}. Payment DCNs: {}. Document DCNs: {}",
+                zipFilename,
+                containerName,
+                envelope.payments.stream().map(payment -> payment.documentControlNumber).collect(joining(",")),
```
lgonczar, line 249: Can payments be null?

lgonczar: Just checked, null will be converted to an empty list.

PR #1333, opened 2020-06-08
Refactor Error Mappings

aliveni:
https://tools.hmcts.net/jira/browse/BPS-974

### Change description ###
- Added 2 subclasses `InvalidMetafileException` and `ZipFileProcessingFailedException` to represent error codes `ERR_METAFILE_INVALID` and `ERR_ZIP_PROCESSING_FAILED`
- Use `instance of` to get the error mapping for the exception class
- Add logging when no error mapping found.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/errornotifications/ErrorMapping.java
```diff
@@ -55,6 +35,10 @@ private ErrorMapping() {
     }
 
     public static Optional<ErrorCode> getFor(Class<?> clazz) {
-        return Optional.ofNullable(mapping.get(clazz));
+        return mapping.entrySet()
+            .stream()
+            .filter(entry -> clazz.isInstance(entry.getKey()))
+            .map(Map.Entry::getValue)
+            .findFirst();
```
rkondratowicz, line 42: may be worth adding a comment on the `mapping` stating that the order matters

PR #1336, opened 2020-06-09
appinsight - trace schedule runs as new request

mdayican:
### JIRA link (if applicable) ###

https://tools.hmcts.net/jira/browse/BPS-1201

### Change description ###

trace all schedules runs in appinsight.
very similar to : hmcts/send-letter-service#433

*with shedlock Tests were working local without any waits and it was working with waits in the pipeline with waits apart from deletetasktest. At the end tests are isolated from shedlock.

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1343, opened 2020-06-11
send trace id for envelope queue messages

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1201

### Change description ###

send trace id for envelope queue messages.
aid is track that message in appinsight until is completely finished.
we can use any name, in their example they used  "ParentId".
https://docs.microsoft.com/en-us/azure/azure-monitor/app/custom-operations-tracking

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
@@ -3,6 +3,9 @@ package uk.gov.hmcts.reform.bulkscanprocessor.services.servicebus;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.microsoft.applicationinsights.internal.util.LocalStringsUtils;
+import com.microsoft.applicationinsights.web.internal.RequestTelemetryContext;
+import com.microsoft.applicationinsights.web.internal.ThreadContext;
 import com.microsoft.azure.servicebus.IQueueClient;
 import com.microsoft.azure.servicebus.Message;
 import com.microsoft.azure.servicebus.MessageBody;
@@ -12,6 +15,7 @@ import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.InvalidMessageException;
 import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.ServiceBusConnectionTimeoutException;
 import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.Msg;
 
+import java.util.HashMap;
 import java.util.concurrent.CompletableFuture;
 import javax.annotation.PreDestroy;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
```diff
@@ -69,7 +75,17 @@ Message mapToBusMessage(Msg msg) {
         busMessage.setMessageId(msg.getMsgId());
         busMessage.setMessageBody(getMsgBodyInBytes(msg));
         busMessage.setLabel(msg.getLabel());
-
+        busMessage.setProperties(new HashMap<String,String>());
+        RequestTelemetryContext context = ThreadContext.getRequestTelemetryContext();
```
mdayican, line 79: when we start assigning request telemetry `getHttpRequestTelemetry ` should not be null

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
```diff
@@ -69,7 +75,17 @@ Message mapToBusMessage(Msg msg) {
...
+        String parentId;
+        if (context != null
+            && context.getHttpRequestTelemetry() != null
+            && context.getHttpRequestTelemetry().getId() != null) {
+            parentId = context.getHttpRequestTelemetry().getId();
+        } else {
+            parentId = LocalStringsUtils.generateRandomIntegerId();
```
nitinprabhu, line 86: do we have to create `HttpRequestTelemetry` if `context.getHttpRequestTelemetry()` is null ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
```diff
@@ -69,7 +75,17 @@ Message mapToBusMessage(Msg msg) {
...
+        }
+        busMessage.getProperties().put(PARENT_ID, parentId);
```
nitinprabhu, line 88: how and where we are going to use this id ?

mdayican: reader will get this and put to telemetry request. and if reader process produce any queue message we can use this id as parent id.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
mdayican: it is in other pr. when tasks run we will create telemetry request by AOP.
https://github.com/hmcts/bulk-scan-processor/pull/1336

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/servicebus/ServiceBusHelper.java
mdayican: Message producers will try to keep this id if it is a result of another request. we can introduce another id like span-id(operation-id) it will change with each process.
for example if this envelope has payment it will create a message in payment queue with same parent-id but different  span-id (operation-id)?

mdayican:
later we will plan this change

PR #1344, opened 2020-06-11
IGNORE

mdayican:
### Change description ###

test scannable items order

```
[ ] Yes
[ ] No
```


mdayican:
tested scannable items order, loose the order after save to db

PR #1346, opened 2020-06-12
maintain document order

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1267

### Change description ###

for now we do not need,
maintain document order

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/db/migration/V057__Add_document_order_column_to_scannable_items.sql
+++ b/src/main/resources/db/migration/V057__Add_document_order_column_to_scannable_items.sql
@@ -1,0 +1,2 @@
+ALTER TABLE scannable_items
+  ADD COLUMN documentOrder INTEGER DEFAULT 0;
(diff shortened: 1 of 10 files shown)


src/main/resources/db/migration/V057__Add_document_order_column_to_scannable_items.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE scannable_items
+  ADD COLUMN documentOrder INTEGER DEFAULT 0;
```
lgonczar, line 2: Best to add not null contraint.

mdayican: how we can sort out existing ones?

lgonczar: I thought that's why you added the default. Ultimately we shouldn't even have default value for this column, as we'll be always setting it from code.

rkondratowicz:
stale - closing

PR #1349, opened 2020-06-15
send ocr errors in notification messages

mdayican:

https://tools.hmcts.net/jira/browse/BPS-910

### Change description ###

we need to send messages to exela but not log in anywhere.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrValidationException.java
```diff
@@ -1,7 +1,21 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
 
 public class OcrValidationException extends InvalidMetafileException {
+
+    //might contain sensitive data
```
rkondratowicz, line 5: formatting

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -103,10 +103,11 @@ private void handleValidationResponse(
...
         switch (res.status) {
             case ERRORS:
+                String message =  "OCR validation service returned OCR-specific errors. "
+                    + "Document control number: " + docWithOcr.documentControlNumber + ". "
+                    + "Envelope: " + envelope.zipFileName + ".";
                 throw new OcrValidationException(
-                    "OCR validation service returned OCR-specific errors. "
-                        + "Document control number: " + docWithOcr.documentControlNumber + ". "
-                        + "Envelope: " + envelope.zipFileName + "."
+                    message, message +" Errors: " + res.errors
```
rkondratowicz, line 110: formatting

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrValidationException.java
```diff
@@ -1,7 +1,21 @@
...
+    }
+
+    public OcrValidationException(String message, String detailMessage) {
+        super(message);
+        this.detailMessage = detailMessage;
+    }
+
+    public String getDetailMessage() {
+        return detailMessage;
     }
```
rkondratowicz, line 20: Looks like this is always used and never the 'normal' message.
Can we just put the detailed message to message field?

mdayican: we do not want to log that is the reason

mdayican: log.error("exception ", ex) will print the message

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -103,10 +103,11 @@ private void handleValidationResponse(
...
         switch (res.status) {
             case ERRORS:
+                String message = "OCR validation service returned OCR-specific errors. "
+                    + "Document control number: " + docWithOcr.documentControlNumber + ". "
+                    + "Envelope: " + envelope.zipFileName + ".";
                 throw new OcrValidationException(
-                    "OCR validation service returned OCR-specific errors. "
-                        + "Document control number: " + docWithOcr.documentControlNumber + ". "
-                        + "Envelope: " + envelope.zipFileName + "."
+                    message, message + " Errors: " + res.errors
```
apereverzin, line 110: Would be better just to have errors in detailed message. Also not sure if writing list (res.errors) will be readable enough.

mdayican: we need document control and filename, it will print the error  like this: ["", ""]

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -360,7 +361,11 @@ private void sendErrorMessage(
         ErrorCode errorCode
     ) {
         try {
-            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+            String message =
+                cause instanceof OcrValidationException
+                    ? ((OcrValidationException) cause).getDetailMessage()
+                    : cause.getMessage();
+            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, message);
```
apereverzin, line 368: As of the comment below for me looks better:
String message = cause.getMessage();
if (cause instanceof OcrValidationException) {
    message += cause.getDetailMessage();
}

mdayican: changed the logic

apereverzin: not sure I understand what you mean

mdayican: refresh the page, I changed with if.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -103,10 +103,12 @@ private void handleValidationResponse(
...
             case ERRORS:
+                String message = "OCR validation service returned OCR-specific errors. "
+                    + "Document control number: " + docWithOcr.documentControlNumber + ". "
+                    + "Envelope: " + envelope.zipFileName + ".";
                 throw new OcrValidationException(
-                    "OCR validation service returned OCR-specific errors. "
-                        + "Document control number: " + docWithOcr.documentControlNumber + ". "
-                        + "Envelope: " + envelope.zipFileName + "."
+                    message,
+                    message + "Errors: " + res.errors
```
lgonczar, line 111: Good idea with the fields that contains detailed information.

This message is only good as the official `message` in OcrValidationException. `detailMessage` should just contain errors, as the message that we have is only good for our logging. In particular, the `OCR validation service returned OCR-specific errors` bit will not be understood by our supplier.

Errors should also be properly formatted, i.e. we should explicitly build a string from them using String.join(...) or something similar.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -360,7 +361,11 @@ private void sendErrorMessage(
         ErrorCode errorCode
     ) {
         try {
-            sendErrorMessageToQueue(zipFilename, containerName, eventId, errorCode, cause);
+            String message = cause.getMessage();
+            if (cause instanceof OcrValidationException) {
+                message = ((OcrValidationException) cause).getDetailMessage();
+            }
```
lgonczar, line 367: This is generic code looking for a very specific exception. We need this additional field which won't be logged, so how about adding it to `InvalidEnvelopeException` instead of `OcrValidationException`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
lgonczar: Or we could start `detailMessage` with something like: `Service {} returned following OCR field validation errors:`. Still without document control number and envelope, because those will be known.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
mdayican: ocr exception extends InvalidMetafileException, but we can add. but I did not see any benefit adding there.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
mdayican: `OCR validation service returned OCR-specific errors. Document control number: 852352234039153252048. Envelope: 59423325200994520_15-02-2019-08-08-19.zip. Errors: [first_name is missing, last_name is missing]`

this how it looks with our service

mdayican: > Or we could start `detailMessage` with something like: `Service {} returned following OCR field validation errors:`. Still without document control number and envelope, because those will be known.

does provider care about service?

PR #1356, opened 2020-06-18
Delete DOC_PROCESSED events

aliveni:
https://tools.hmcts.net/jira/browse/BPS-517

### Change description ###
DOC_PROCESSED event is no longer used in the processor. 
Delete all entries with `DOC_PROCESSED` event type from `process_events` table. 

```
[ ] Yes
[x] No
```


lgonczar:
I think we should do things like this through an Ops ticket. Is that correct, @mdayican ?

mdayican:
> I think we should do things like this through an Ops ticket. Technically the same outcome, but I heard that was the recommendation for some reason. Is that correct, @mdayican ?

last time when I was modifying the notification status, I created ops ticket as I want to see the count and filenames, I think @apereverzin told it is recommended.

PR #1360, opened 2020-06-18
add new parent exception

mdayican:
### Change description ###

add new parent exception which has  detailmessage.
suppressed the sonar warning  "Inheritance tree of classes should not be too deep".
better to talk with team and create another ticket to tidy up exceptions. 
to keep it simple this can be the first pr.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/ZipFileProcessingFailedException.java
```diff
@@ -3,7 +3,7 @@
 /**
  * Base class for all exceptions that represent failed zip file processing situation.
  */
-public abstract class ZipFileProcessingFailedException extends RuntimeException {
+public abstract class ZipFileProcessingFailedException extends RejectionException {
```
lgonczar, line 6: This one also needs to be an interface

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/ZipNameNotMatchingMetaDataException.java
```diff
@@ -1,6 +1,6 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
 
-public class ZipNameNotMatchingMetaDataException extends InvalidMetafileException {
+public class ZipNameNotMatchingMetaDataException extends RejectionException implements InvalidMetafileException {
```
lgonczar, line 3: All exceptions that previously extended `InvalidMetadatafileException` should now extend `InvalidEnvelopeException`. It, in turn, along with `PaymentsDisabledException` and `ServiceDisabledException`, should extend `RejectionException`.

mdayican: so InvalidEnvelopeException will be interface?

lgonczar: On a second thought, maybe we could get rid of `InvalidEnvelopeException` and just replace it with `RejectionException`. Technically, it doesn't add any benefit.

lgonczar: > so InvalidEnvelopeException will be interface?

I think it should be completely replaced by `RejectionException`.

PR #1367, opened 2020-06-19
fix staging  notification conf

mdayican:
### Change description ###

fix staging  notification conf

```
[ ] Yes
[x] No
```


mdayican:
we use aat queue

PR #1368, opened 2020-06-19
Use non staging for external notification service

nitinprabhu:


nitinprabhu:
Merging as it is used only on AAT staging

PR #1373, opened 2020-06-22
Add pact test for sample-app OCR validation

rkondratowicz:
Just one contract (for happy path) in this PR.
Regarding authentication, I propose going with a hardcoded s2s token - sample app will stub s2s client when verifying the contract.
Read more about this here: https://docs.pact.io/provider/handling_auth

Changes related to publishing contract will be added in next PRs.

Jira ticket: https://tools.hmcts.net/jira/browse/BPS-1279


rkondratowicz:
reopening - jenkins helm issue

PR #1376, opened 2020-06-23
Feature/add pact test for sample app ocr validation

rkondratowicz:
Just one contract (for happy path) in this PR.
Regarding authentication, I propose going with a hardcoded s2s token - sample app will stub s2s client when verifying the contract.
Read more about this here: https://docs.pact.io/provider/handling_auth

Changes related to publishing contract will be added in next PRs.

Jira ticket: https://tools.hmcts.net/jira/browse/BPS-1279


build.gradle
```diff
@@ -314,6 +327,12 @@ dependencies {
   smokeTestImplementation sourceSets.test.runtimeClasspath
   smokeTestImplementation group: 'io.rest-assured', name: 'rest-assured'
   functionalTestImplementation group: 'com.typesafe', name: 'config', version: '1.4.0'
+
+  contractTestImplementation sourceSets.main.runtimeClasspath
+  contractTestImplementation sourceSets.test.runtimeClasspath
+  contractTestImplementation group: 'io.rest-assured', name: 'rest-assured'
+  contractTestImplementation group: 'au.com.dius', name: 'pact-jvm-consumer-java8', version: '4.0.10'
```
doncem, line 334: is this needed?

build.gradle
```diff
@@ -56,6 +64,11 @@ test {
   failFast = true
 }
 
+task contract(type: Test, description: 'Runs the functional tests.', group: 'Verification') {
```
mdayican, line 67: description is wrong?

build.gradle
rkondratowicz: fixed

build.gradle
rkondratowicz: oops, fixed

PR #1377, opened 2020-06-23
Rejection exception and tidy up Exceptions

mdayican:
### Change description ###

add new parent exception which has detailmessage.
errormaping removed it is in RejectionException

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/RejectionException.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.ErrorCode;
+
+public abstract class RejectionException extends RuntimeException {
```
lgonczar, line 5: This is a suggestion only. I probably proposed this name, but I'm not sure it's descriptive enough. Maybe `EnvelopeRejectionException`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/RejectionException.java
```diff
@@ -0,0 +1,36 @@
...
+
+    // might contain sensitive data
+    private final String detailMessage;
+    private final ErrorCode errorCode;
+
+    public RejectionException(ErrorCode errorCode, String message) {
+        super(message);
+        this.detailMessage = message;
+        this.errorCode = errorCode;
+    }
```
lgonczar, line 15: This one might be a bit risky, because developers might not see the distinction between `message` and `detailMessage`. Maybe we could drop this constructor?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/RejectionException.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+import uk.gov.hmcts.reform.bulkscanprocessor.model.out.msg.ErrorCode;
+
+public abstract class RejectionException extends RuntimeException {
+
+    // might contain sensitive data
+    private final String detailMessage;
```
lgonczar, line 8: How about `detailedMessage` or `errorDescription`?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/RejectionException.java
lgonczar, line 5: Nice change moving error code to the exceptions.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/RejectionException.java
mdayican: if we drop we need to change all the other subclasses. for now we can keep like this and add comment.

lgonczar: ok

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/EnvelopeRejectionException.java
```diff
@@ -0,0 +1,37 @@
...
+
+public abstract class EnvelopeRejectionException extends RuntimeException {
+
+    // might contain sensitive data
+    private final String errorDescription;
+    private final ErrorCode errorCode;
+
+    public EnvelopeRejectionException(ErrorCode errorCode, String message) {
+        super(message);
+        this.errorDescription = message;
```
nitinprabhu, line 13: we can add comment here as below  :) `//use message as error description`

PR #1385, opened 2020-06-26
Publish error notification message with container and zip file name

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1293

### Change description ###
In Bulk scan processor, publish error notification message to queue with container name and zip file name as message id

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/ErrorNotificationSender.java
```diff
@@ -60,7 +61,7 @@ private void sendErrorMessageToQueue(
```
apereverzin, line 61: Should be:
String messageId = StringUtils.joinWith("_", containerName, zipFilename);
This messageId is used in log.info(...)

aliveni: Fixed it

aliveni:
> Just a question in router we are sending envelope id as message id and processor is already sending envelope id which is unique anyways so do we need this change ?

As discussed, zip file can be rejected before creating the envelope.

PR #1392, opened 2020-06-30
Event creation

apereverzin:

### Change description ###
Moved event creation from FileErrorHandler to FileContentProcessor. This makes code cleaner and allows to better understand the workflow.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileContentProcessorTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileContentProcessorTest.java
@@ -35,11 +35,12 @@ class FileContentProcessorTest {
     private static final String FILE_NAME = "file1.zip";
     private static final String CONTAINER_NAME = "container";
     private static final String LEASE_ID = "leaseID";
-    private static final String DCN = "dcn";
     private static final String POBOX = "pobox";
     private static final String BULKSCAN = "bulkscan";
     private static final String CASE_NUMBER = "case_number";
     private static final String CASE_REFERENCE = "case_reference";
+    private static final Long EVENT_ID = 1L;
+    public static final String MSG = "msg";
 
     @Mock
     private EnvelopeProcessor envelopeProcessor;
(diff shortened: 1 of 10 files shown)


lgonczar:
I'd like to propose an alternative solution: renaming `FileErrorHandler` to `FileRejector`/`FileRejectionService`/`EnvelopeReje...` - this would make it clear what it does. Creating an event would then be a natural part of rejection and doesn't need to be extracted.

apereverzin:
@lgonczar The point of this PR is to make clear where failure event is created.Before it was created in FileContentProcessor and in FileErrorHandler, now only in FileContentHandler, this makes code more clear. I agree with renaming to FileRejector, and also suggest to rename _handleInvalidFileError_ method to _handleInvalidFile_.

apereverzin:
@lgonczar The point of this PR is to make clear where failure event is created.Before it was created in FileContentProcessor and in FileErrorHandler, now only in FileContentHandler, this makes code more clear. I agree with renaming to FileRejector, and also suggest to rename _handleInvalidFileError_ method to _handleInvalidFile_.

lgonczar:
> @lgonczar The point of this PR is to make clear where failure event is created.Before it was created in FileContentProcessor and in FileErrorHandler, now only in FileContentHandler, this makes code more clear. I agree with renaming to FileRejector, and also suggest to rename _handleInvalidFileError_ method to _handleInvalidFile_.

Fair enough

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileContentProcessorTest.java
```diff
@@ -35,11 +35,12 @@
     private static final String FILE_NAME = "file1.zip";
     private static final String CONTAINER_NAME = "container";
     private static final String LEASE_ID = "leaseID";
-    private static final String DCN = "dcn";
     private static final String POBOX = "pobox";
     private static final String BULKSCAN = "bulkscan";
     private static final String CASE_NUMBER = "case_number";
     private static final String CASE_REFERENCE = "case_reference";
+    private static final Long EVENT_ID = 1L;
+    public static final String MSG = "msg";
```
lgonczar, line 43: Can we keep this change outside this PR? I don't agree with creating static variables without any particular meaning rather than being used multiple times as any value. We can move that discussion into another PR, not to block this one.

PR #1396, opened 2020-06-30
Existing envelope

apereverzin:

### Change description ###
Code improvement - made result of checking envelope existence Optional rather than using null and some brushing up.

Also noticed that envelope existence runs 2 times - before acquiring lease and after acquiring lease. I assume this is justified but whoever will be reviewing this PR please confirm this workflow is correct.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
+++ b/src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
@@ -69,8 +69,8 @@ class BlobProcessorTaskTest {
         given(blob.getUri()).willReturn(URI.create("file.zip"));
         given(container.getBlockBlobReference("file.zip")).willReturn(cloudBlockBlob);
         given(container.getName()).willReturn("cont");
-        given(envelopeProcessor.getEnvelopeIdByFileAndContainer("cont", "file.zip"))
-            .willReturn(Optional.empty());
+        given(envelopeProcessor.getEnvelopeByFileAndContainer("cont", "file.zip"))
+            .willReturn(null);
         given(cloudBlockBlob.exists()).willReturn(true);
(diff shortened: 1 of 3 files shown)


rkondratowicz:
I don't see this change as an improvement.
There's more code now and then new method `tryLeaseAndProcessZipFile` that wraps `leaseAndProcessZipFile` adds confusion imho.

apereverzin:
Using Optional instead of checking for null and introducing a method instead of _else - if - else_ - this makes logic cleaner imho.

Please, also reply to my question in the description

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTest.java
```diff
@@ -69,8 +69,8 @@ void processBlobs_should_not_call_envelopeProcessor_if_failed_to_load_file() thr
         given(blob.getUri()).willReturn(URI.create("file.zip"));
         given(container.getBlockBlobReference("file.zip")).willReturn(cloudBlockBlob);
         given(container.getName()).willReturn("cont");
-        given(envelopeProcessor.getEnvelopeByFileAndContainer("cont", "file.zip"))
-            .willReturn(null);
+        given(envelopeProcessor.getEnvelopeIdByFileAndContainer("cont", "file.zip"))
+            .willReturn(Optional.empty());
```
doncem, line 73: do we need a test with non empty optional too?

apereverzin: Not right now. BlobProcessorTaskTest always contained this only method effectively making BlobProcessorTask not unit tested (all code coverage was provided only by integration tests). After refactoring BlobProcessorTask will become unit testable and more methods will be added to BlobProcessorTaskTest.

apereverzin:
Not necessary

PR #1401, opened 2020-07-07
File names extractor

apereverzin:

### Change description ###
Keep refactoring BlobProcessorTask - introduced FileNamesExtractor component.

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuite.java
```diff
@@ -124,6 +126,8 @@ public void setUp() throws Exception {
 
         blobManager = new BlobManager(cloudBlobClient, blobManagementProperties);
 
+        fileNamesExtractor = new FileNamesExtractor();
```
rkondratowicz, line 129: can it not be autowired?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -49,16 +46,19 @@
...
+    private final FileNamesExtractor fileNamesExtractor;
+
     private final EnvelopeProcessor envelopeProcessor;
 
     private final FileContentProcessor fileContentProcessor;
 
     public BlobProcessorTask(
         BlobManager blobManager,
-        EnvelopeProcessor envelopeProcessor,
+        FileNamesExtractor fileNamesExtractor, EnvelopeProcessor envelopeProcessor,
```
rkondratowicz, line 57: formatting

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractorTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    void should_shuffle_multiple_file_names() {
+        // given
+        ListBlobItem blob0 = mock(ListBlobItem.class);
+        ListBlobItem blob1 = mock(ListBlobItem.class);
+        ListBlobItem blob2 = mock(ListBlobItem.class);
+        ListBlobItem blob3 = mock(ListBlobItem.class);
+        ListBlobItem blob4 = mock(ListBlobItem.class);
```
rkondratowicz, line 69: can we use `var`?

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuite.java
apereverzin: FileNamesExtractor depends on _scheduling.task.scan.enabled_ property (along with BlobProcessorTask) which is false in integration tests.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+                if (Strings.isNullOrEmpty(fileName)) {
+                    log.error("Cannot extract filename from list blob item. URI: {}", b.getUri());
+                } else {
+                    zipFilenames.add(fileName);
+                }
+            });
+        Collections.shuffle(zipFilenames);
+        return zipFilenames;
+    }
+}
```
doncem, line 38: looks like a utility class. just shuffling some list elements around. can avoid autowiring and can be unit tested as well

doncem: plus it's not strictly bound to scan task - if needed, perhaps can be applied to any list blob

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+@Component
+@ConditionalOnProperty(value = "scheduling.task.scan.enabled", matchIfMissing = true)
```
doncem, line 16: utility classes don't need spring aop annotations ;)

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+public class FileNamesExtractor {
```
doncem, line 17: utility classes are final

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final Logger log = LoggerFactory.getLogger(FileNamesExtractor.class);
+
+    private FileNamesExtractor() {
+        // Utility class
+    }
+
+    public static List<String> getZipFileNamesFromContainer(CloudBlobContainer container) {
```
doncem, line 24: since it's a utility, perhaps something like `shuffleContainerBlobs` will suit? mind it's a list of strings which signifies file names so something along those words would be ideal :upside_down_face:

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -76,29 +73,19 @@ public void processBlobs() {
...
         for (String zipFilename : zipFilenames) {
             tryProcessZipFile(container, zipFilename);
         }
 
         log.info("Finished processing blobs for container {}", container.getName());
     }
 
+    List<String> getFileNames(CloudBlobContainer container) {
+        return getZipFileNamesFromContainer(container);
+    }
```
doncem, line 87: why? it can be unit tested on it's own no need for nested method

apereverzin: In BlogProcessorTaskTest it has to be mocked. Tried to use PowerMock and JMockit for mocking static call, both have some issues when being used with JUnit5

doncem: but why mock? can just use same thing as in utility unit test:

```java
given(container.listBlobs()).willReturn(singletonList(blob));
given(blob.getUri()).willReturn(URI.create("file.zip"));
```

apereverzin: Ok.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileNamesExtractorTest.java
```diff
@@ -0,0 +1,82 @@
...
+        assertThat(zipFileNames).hasSize(blobs.size());
+        var isShuffled = false;
+        // ensure resulting file names are not in the same order as file names of original blobs
+        for (var i = 0; i < blobs.size(); i++) {
+            if (!zipFileNames.get(i).equals("file" + i + ".zip")) {
+                isShuffled = true;
+                break;
+            }
+        }
+        assertThat(isShuffled).isTrue();
```
nitinprabhu, line 80: can we not use `containsExactlyInAnyOrder` ?

apereverzin: No. Shuffling file names is essential part of the logic and the result is unpredictable.

doncem: i think the key point here is `InAnyOrder` :wink:

nitinprabhu: I am not sure if I understand that. 
input list -> zip1, zip2 ,zip3
after shuffling -> zip3, zip1, zip2 
so why we can't use `containsExactlyInAnyOrder`

apereverzin: Because zip1, zip2, zip3 will also pass but in reality it is not shuffled but it must be.

apereverzin: `InAnyOrder` includes also `inTheSameOrder`, but we need `InDifferentOrder`.

PR #1406, opened 2020-07-10
Added logs to processor

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1297

### Change description ###
There is 'PDFs found in ...' log entry marking the end of processing zip file, but there is no log entry marking the beginning of the processing. This PR fills this gap make lease expiration problem better analysable.

Logs added to ZipFileProcessor and UploadEnvelopeDocumentsService

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
@@ -88,6 +88,8 @@ public class UploadEnvelopeDocumentsService {
         CloudBlockBlob blob = null;
         Optional<String> leaseId = Optional.empty();
 
+        log.info("Processing envelope. Container {}, File {}", containerName, zipFileName);
+
         try {
             blob = getCloudBlockBlob(blobContainer, zipFileName, envelopeId);
             leaseId = blobManager.acquireLease(blob, containerName, envelope.getZipFileName());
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/UploadEnvelopeDocumentsService.java
```diff
@@ -88,6 +88,8 @@ private void processEnvelope(CloudBlobContainer blobContainer, Envelope envelope
         CloudBlockBlob blob = null;
         Optional<String> leaseId = Optional.empty();
 
+        log.info("Processing envelope. Container {}, File {}", containerName, zipFileName);
```
nitinprabhu, line 91: We already have processing zip file in the logs 
image

apereverzin: ZipFileProcessor::process which produces this log entry is called from BlobProcessorTask (exactly - from FileContentProcessor) and from UploadEnvelopeDocumentsTask (exactly - from UploadEnvelopeDocumentsService). In the first case there is a log entry from the calling method, in the second case - no, unless this new line is merged. With this new line reading logs and finding exact workflows will be less confusing.

rkondratowicz: should the existing one be deleted then?

apereverzin: No. The existing log is in BlobProcessorTask, the method writing this log entry is called by 2 callers, one of these callers writes a log entry, the other caller does not. This PR is fixing this - both callers will write log entries and thus the whole workflow will be traceable.

apereverzin:
Not necessary.

PR #1415, opened 2020-07-16
upgrade java chart version

mdayican:
### Change description ###

upgrade java chart version,
we do not receive flux test success notifications.

```
[ ] Yes
[x] No
```

### DIFF
--- a/charts/bulk-scan-processor/requirements.yaml
+++ b/charts/bulk-scan-processor/requirements.yaml
@@ -1,6 +1,6 @@
 dependencies:
   - name: java
-    version: ~2.15.0
+    version: ~3.1.0
     repository: '@hmctspublic'
   - name: servicebus
     version: ~0.2.2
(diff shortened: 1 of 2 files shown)


charts/bulk-scan-processor/requirements.yaml
```diff
@@ -1,6 +1,6 @@
 dependencies:
   - name: java
-    version: ~2.15.0
+    version: ~3.1.0
```
doncem, line 3: that's a major upgrade. perhaps need to have a look at the diff and apply changes accordingly. e.g. - there's no need for readiness flag in values.yaml. requirement.yaml is deprecated. stuff like that. are we sure we are ready to jump on v3 here?

doncem: i know it's in itself (requirements :D) but still - worth looking over rather than bumping to v3 blindly

PR #1416, opened 2020-07-16
remove blob existence check for Func test

mdayican:
### Change description ###

in flux func. test that check can fail.

```
[ ] Yes
[x] No
```


mdayican:
> Could you please add why this is being removed ?

in aat another process can pick up.

PR #1436, opened 2020-08-03
Feature: build queue clients using access key and namespace

doncem:

[Improve the way Service Bus clients are created](https://tools.hmcts.net/jira/browse/BPS-771)

### Change description ###

all but 1 queue is not yet available to be built using access key. this is addressed as well.

~this pr introduces a different way (compared to before and to payment processor which only had one queue client hmcts/bulk-scan-payment-processor#264). in case reviewers see it as a problem or anything - can go back to old way. but first - i'm interesting to see the pipeline run as well~

^ didn't work. will revert re-work

```
[ ] Yes
[x] No
```


src/main/resources/application.yaml
```diff
@@ -98,11 +98,11 @@ queues:
     connection-string: ${QUEUE_PROCESSED_ENVELOPES_READ}
     queue-name: ${QUEUE_PROCESSED_ENVELOPES_NAME}
   notifications:
-    access-key: ${QUEUE_NOTIFICATIONS_SEND_ACCESS_KEY}
+    access-key: ${QUEUE_NOTIFICATIONS_SEND_ACCESS_KEY:}
```
aliveni, line 101: do we need the default value here?

aliveni: same for `namespace-override`

doncem: we do as notification is not yet merged - it lives in `reform-scan` domain, not `bulk-scan`

PR #1437, opened 2020-08-03
Check for incomplete envelopes every 15min

rkondratowicz:
So that we get notified earlier.

https://github.com/hmcts/cnp-flux-config/pull/5192


infrastructure/variables.tf
```diff
@@ -117,7 +117,7 @@ variable "reports_cron" {
 # region incomplete envelopes monitoring
 
 variable "incomplete_envelopes_cron" {
-  default     = "0 0 * * * *"
+  default     = "0 */15 * * * *"
```
nitinprabhu, line 120: I think this config is not used but worth keeping them in sync or completely get rid of this ?

doncem: yes this was only used for ASE app. completely irrelevant right now

rkondratowicz: we can remove all ASE related stuff on a PR dedicated to that subject.

PR #1441, opened 2020-08-04
upgrade azure blob-lib and use in SAS retrieval

mdayican:

### Change description ###

upgrade azure blob-lib and just use in SAS retrieval

```
[ ] Yes
[x] No
```


mdayican:
stuck in pipeline

PR #1442, opened 2020-08-04
Upgrade blob lib for sas token

mdayican:

### Change description ###

Upgrade blob lib for sas token.
For now old and new version will be there and we will use new version just in sas token creation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/HttpConfiguration.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/HttpConfiguration.java
@@ -1,16 +1,23 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.config;
 
+import com.azure.core.http.HttpClient;
+import com.azure.core.http.ProxyOptions;
+import com.azure.core.http.netty.NettyAsyncHttpClientBuilder;
 import feign.Client;
 import feign.httpclient.ApacheHttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/config/HttpConfiguration.java
```diff
@@ -44,4 +51,30 @@ private CloseableHttpClient getHttpClient() {
             .setDefaultRequestConfig(config)
             .build();
     }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "storage", name = "proxy_enabled", havingValue = "true")
```
doncem, line 56: with proxy, dunno - false or true occasion for `match if missing`?

mdayican: I think we do not need proxy. default value is false in property. for now I am leaving it like this.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/SasTokenGeneratorService.java
```diff
@@ -13,55 +12,46 @@
...
             throw new UnableToGenerateSasTokenException(e);
         }
     }
 
-    private SharedAccessBlobPolicy createSharedAccessPolicy(String serviceName) {
-        TokenConfig config = getTokenConfigForService(serviceName);
-
-        SharedAccessBlobPolicy policy = new SharedAccessBlobPolicy();
-        policy.setPermissions(EnumSet.of(WRITE, LIST));
-        policy.setSharedAccessExpiryTime(Date.from(now().plusSeconds(config.getValidity()).toInstant(ZoneOffset.UTC)));
```
nitinprabhu, line 62: what about the timezone ?

mdayican: OffsetDateTime  is A date-time with an offset from UTC/Greenwich .

doncem: depends on servers. i guess our pods are in UK South

nitinprabhu: I hope the timezone is set correctly otherwise this would block our consumer of file uploads

mdayican: I will set to UTC explicitly

PR #1443, opened 2020-08-05
use new lib in contributor

mdayican:

### Change description ###

use new lib in contributor

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/info/BlobContainerInfoContributor.java
```diff
@@ -15,18 +15,18 @@
...
     }
 
     @Override
     public void contribute(Builder builder) {
         builder.withDetail(
             "containers",
-            StreamSupport.stream(client.listContainers().spliterator(), false)
-                .map(CloudBlobContainer::getName)
+            StreamSupport.stream(client.listBlobContainers().spliterator(), false)
+                .map(BlobContainerItem::getName)
```
doncem, line 29: looks like a perfect time to get rid of extra dependencies used here, right? we can just `client.listContainers().stream().map().collect()` ;)

PR #1445, opened 2020-08-05
upgrade blob lib part 3 - use in Rejected files report

mdayican:

### Change description ###

upgrade blob lib part 3 - use in Rejected files report

new Blob Client `BlobServiceClient` added to `BlobManager` so constructor changed and to make the PR small I added new method `getRejectedContainers` same as  `listRejectedContainers` (still using in `CleanUpRejectedFilesTask`, will remove in next PR and rename method `get` to `list`)

As  `BlobManager` change cause changes in related tests.
Just `RejectedFilesReportTest` is using the new client, others assigned with null.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/RejectedFilesReportService.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/RejectedFilesReportService.java
@@ -36,7 +36,7 @@ public class RejectedFilesReportService {
 
     private Stream<BlobItem> getBlobs(BlobContainerClient container) {
         ListBlobsOptions listOptions = new ListBlobsOptions();
-        listOptions.getDetails().setRetrieveSnapshots(false);
+        listOptions.getDetails().setRetrieveSnapshots(true);
         return container
                 .listBlobs(listOptions, null).stream();
     }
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/RejectedFilesReportService.java
```diff
@@ -26,23 +23,21 @@ public RejectedFilesReportService(BlobManager blobManager) {
...
-    private Stream<ListBlobItem> getBlobs(CloudBlobContainer container) {
-        return StreamSupport.stream(
-            container
-                .listBlobs(null, true, EnumSet.of(SNAPSHOTS), null, null)
-                .spliterator(),
-            false
-        );
+    private Stream<BlobItem> getBlobs(BlobContainerClient container) {
+        ListBlobsOptions listOptions = new ListBlobsOptions();
+        listOptions.getDetails().setRetrieveSnapshots(false);
```
nitinprabhu, line 39: does existing code retrieves snapshots ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/RejectedFilesReportService.java
```diff
@@ -26,23 +23,21 @@ public RejectedFilesReportService(BlobManager blobManager) {
...
+                        listItem.getName(),
+                        container.getBlobContainerName()
                     )))
             .collect(toList());
     }
 
-    private Stream<ListBlobItem> getBlobs(CloudBlobContainer container) {
-        return StreamSupport.stream(
-            container
-                .listBlobs(null, true, EnumSet.of(SNAPSHOTS), null, null)
```
nitinprabhu, line 43: what about snapshots in the new code ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/RejectedFilesReportService.java
mdayican: previous code returns snapshots , we should return snapshots

PR #1448, opened 2020-08-07
use new lib in rejected file cleaning

mdayican:

### Change description ###

use new lib in rejected file cleaning.
LeaseAcquirer introduced from blob router.
some part moved to #1451 and merged.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
@@ -1,7 +1,9 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 
+import com.azure.storage.blob.BlobClient;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.models.BlobItem;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
@@ -15,8 +17,15 @@ import uk.gov.hmcts.reform.bulkscanprocessor.tasks.processor.BlobManager;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static uk.gov.hmcts.reform.bulkscanprocessor.services.storage.LeaseMetaDataChecker.LEASE_EXPIRATION_TIME;
+import static uk.gov.hmcts.reform.bulkscanprocessor.util.TimeZones.EUROPE_LONDON_ZONE_ID;
 
 @IntegrationTest
 public class CleanUpRejectedFilesTaskTest {
@@ -62,26 +71,61 @@ public class CleanUpRejectedFilesTaskTest {
     }
 
     @Test
-    public void should_delete_old_files() throws Exception {
+    public void should_delete_old_files_with_snapshots() throws Exception {
         // given
         // there are two files in rejected container
-        rejectedContainer.getBlobClient("foo.zip")
-            .upload(new ByteArrayInputStream("some content".getBytes()),"some content".getBytes().length);
-        rejectedContainer.getBlobClient("bar.zip")
-            .upload(new ByteArrayInputStream("some content".getBytes()),"some content".getBytes().length);
-
+        uploadFile(rejectedContainer, "foo.zip", "content_foo_1");
+        uploadFile(rejectedContainer, "bar.zip", "content_bar_2");
 
         assertThat(rejectedContainer.listBlobs()).hasSize(2); // sanity check
 
-        // one of them has a snapshot
-        rejectedContainer
-            .getBlobClient("bar.zip")
-            .createSnapshot();
+        // one of them has a snapshot and expired lease metadata
+        BlobClient blobClient = rejectedContainer
+            .getBlobClient("bar.zip");
+
+        blobClient.createSnapshot();
+        Map<String, String> blobMetaData = new HashMap<>();
+        blobMetaData.put(LEASE_EXPIRATION_TIME, LocalDateTime.now(EUROPE_LONDON_ZONE_ID).minusSeconds(10).toString());
+        blobClient.setMetadata(blobMetaData);
 
         // when
         new CleanUpRejectedFilesTask(blobManager, leaseAcquirer, "PT0H").run();
 
         // then
         assertThat(rejectedContainer.listBlobs()).isEmpty();
+    }
+
+    @Test
+    public void should_not_delete_old_file_if_metadata_lease_not_expired() throws Exception {
+        // given
+        // there are two files in rejected container
+        uploadFile(rejectedContainer, "foo.zip", "content_foo");
+        uploadFile(rejectedContainer, "bar.zip", "content_bar");
+
+        assertThat(rejectedContainer.listBlobs()).hasSize(2); // sanity check
+
+        // one of them has vslid lease metadata
+        BlobClient blobClient = rejectedContainer
+            .getBlobClient("bar.zip");
+
+        Map<String, String> blobMetaData = new HashMap<>();
+        blobMetaData.put(LEASE_EXPIRATION_TIME, LocalDateTime.now(EUROPE_LONDON_ZONE_ID).plusSeconds(90).toString());
+        blobClient.setMetadata(blobMetaData);
+
+        // when
+        new CleanUpRejectedFilesTask(blobManager, leaseAcquirer, "PT0H").run();
+
+        // then
+        assertThat(rejectedContainer.listBlobs()).isNotEmpty();
+        List<BlobItem> listBlobItem = rejectedContainer.listBlobs().stream().collect(Collectors.toList());
+        assertThat(listBlobItem.size()).isEqualTo(1);
+        assertThat(listBlobItem.get(0).getName()).isEqualTo("bar.zip");
+
+    }
+
+    private void uploadFile(BlobContainerClient containerClient, String fileName, String content) {
+        byte[] byteContent = content.getBytes();
+        containerClient.getBlobClient(fileName)
+            .upload(new ByteArrayInputStream(byteContent), byteContent.length);
     }
 }
--- a/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
@@ -1,6 +1,5 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 
-import com.azure.storage.blob.BlobClient;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobServiceClient;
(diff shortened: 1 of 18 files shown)


mdayican:
> How about we try and split this so that team can be super assured everything is covered nicely?
> 
> Looks like components can be introduced first and then use them in the job:
> 
> 1. `LeaseMetaDataChecker`
> 2. `LeaseAcquirer`
> 3. Use of components 1 and 2 in `CleanUpRejectedFilesTask`
> 
> Another note on 2: looks like a copy. Which is OK but, e.g. this caught my eye:
> 
> ```java
>             if (!isReady) {
>                 release(leaseClient, blobClient);
>             } else {
>                 onLeaseSuccess.accept(leaseId);
>                 if (releaseLease) {
>                     release(leaseClient, blobClient);
>                 }
>             }
> ```
> 
> one has a `if` wrapper for release condition the other doesn't. we should apply on both occasions should we not? in that case (or even current) why not create custom leaseclient which can be autoclosable interface so that one should not worry about that if condition and `try (leaseclient = get it from provider)` will do the rest. one might argue about logging context but custom client can keep the representative blob client along with that boolean flag
> 
> that's just a brainstorm ^

1)can be spilt but this way we can see the usage and with integration test of CleanUpRejectedFilesTask assures they are working. I think we team can cope with it.
2) before doing anything we need to lease the blob(in current code we skipped that part) and if we can not mark metadata we need to releaser lease. Other is depend on operation. if blob is moved or deleted if we try to release  lease it will fail.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +54,60 @@ public void run() {
...
-        Date createdTime = blob.getProperties().getLastModified();
-        Date cutoff = Date.from(Instant.now().minus(this.ttl));
-
-        return createdTime.before(cutoff);
+    private boolean canBeDeleted(BlobItem blobItem) {
+        return blobItem
+            .getProperties()
+            .getLastModified()
+            .plus(ttl)
+            .isBefore(now());
```
doncem, line 111: this is offset now. it'll use whatever servers have configured. should we include timezone then? this also assuming `getLastModified()` is of same type

mdayican: I thought it should be same. I guess last modified and now()  both using the default.

mdayican: same code with blob-router,  but seems better to add zone UTC.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
```diff
@@ -44,33 +62,76 @@ public static void tearDownContainer() {
...
-            .getBlockBlobReference("bar.zip")
-            .createSnapshot();
+        // one of them has a snapshot and expired lease metadata
+        BlobClient blobClient = rejectedContainer
+            .getBlobClient("bar.zip");
+
+        blobClient.createSnapshot();
+        Map<String, String> blobMetaData = new HashMap<>();
+        blobMetaData.put(LEASE_EXPIRATION_TIME, LocalDateTime.now(EUROPE_LONDON_ZONE_ID).minusSeconds(10).toString());
+        blobClient.setMetadata(blobMetaData);
```
nitinprabhu, line 95: comment would be nice here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +56,67 @@ public void run() {
...
-            });
+            .getRejectedContainers()
+            .forEach(this::cleanUpContainer);
 
         log.info("Finished {} job", TASK_NAME);
     }
 
-    private void tryDeleteFile(CloudBlobContainer container, ListBlobItem listItem) {
-        final String zipFileName = FilenameUtils.getName(listItem.getUri().toString());
+    private void cleanUpContainer(BlobContainerClient containerClient) {
```
nitinprabhu, line 65: cleanUpContainer ? are you deleting files here then worth naming it accordingly ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +56,67 @@ public void run() {
...
-        blob.downloadAttributes();
-
-        Date createdTime = blob.getProperties().getLastModified();
-        Date cutoff = Date.from(Instant.now().minus(this.ttl));
-
-        return createdTime.before(cutoff);
+    //ToDo remove log
+    private boolean canBeDeleted(BlobItem blobItem) {
+        OffsetDateTime lastModified = blobItem.getProperties().getLastModified();
+        OffsetDateTime now = now(UTC);
```
nitinprabhu, line 111: UTC ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: deleting files in that container. Same method  as Blob router.

mdayican: `deleteFilesInRejectedContainer`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +56,67 @@ public void run() {
...
+
+        containerClient
+            .listBlobs(listOptions, null)
+            .stream()
+            .filter(this::canBeDeleted)
+            .map(blobItem -> containerClient.getBlobClient(blobItem.getName()))
+            .forEach(blobClient -> leaseAcquirer.ifAcquiredOrElse(
+                blobClient,
+                leaseId -> delete(blobClient, leaseId),
+                errorCode -> {}, // nothing to do if blob not found in rejected container
```
apereverzin, line 78: Should be logged, I assume

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
```diff
@@ -44,33 +62,76 @@ public static void tearDownContainer() {
...
 
-        // one of them has a snapshot
-        rejectedContainer
-            .getBlockBlobReference("bar.zip")
-            .createSnapshot();
+        // one of them has a snapshot and expired lease metadata
+        BlobClient blobClient = rejectedContainer
+            .getBlobClient("bar.zip");
+
+        blobClient.createSnapshot();
```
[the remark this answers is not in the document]
mdayican: should be sth like that: 
` rejectedContainer
            .getBlobClient("bar.zip")
            .createSnapshot()
            .setMetadata(blobMetaData);
`
will add.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: `LeaseAcquirer` logs it.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: @doncem warned about it and added. in blob router we use defaults. 
I think it is better to add UTC as for `getLastModified` it says:
`A DateTimeOffset containing the blob's last-modified time, in UTC format.`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +56,67 @@ public void run() {
...
+                false
+            ));
+
+        log.info("Finished removing rejected files. Container: {}", containerName);
+    }
+
+    private void delete(BlobClient blobClient, String leaseId) {
```
apereverzin, line 85: deleteBlob ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: https://github.com/hmcts/blob-router-service/blob/1bdc43734e38a4239be806c3eafe0c8e138faa96/src/main/java/uk/gov/hmcts/reform/blobrouter/services/RejectedBlobChecker.java#L25

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
```diff
@@ -53,58 +56,67 @@ public void run() {
...
+        OffsetDateTime now = now(UTC);
+        OffsetDateTime lastTimeToStay = lastModified.plus(ttl);
+        log.info(
+            "Blob's last modified: {}, now time: {}, lastTimeToStay: {}",
+            lastModified,
+            now,
+            lastTimeToStay
+        );
+        return lastTimeToStay
+            .isBefore(now(UTC));
```
apereverzin, line 120: No reason to break this line

apereverzin: Any reason for using lastTimeToStay instead of cutoff?

mdayican: types are different that is why this part changed. It is just for logging after I check the logs I will remove the `vars`.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: London zone used

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTask.java
mdayican: Lodnon zone used.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/CleanUpRejectedFilesTaskTest.java
```diff
@@ -44,33 +61,78 @@ public static void tearDownContainer() {
...
+        this.rejectedContainer = blobServiceClient.getBlobContainerClient(("test-rejected"));
+        if (!this.rejectedContainer.exists()) {
+            this.rejectedContainer.create();
+        }
+    }
 
-        this.rejectedContainer = cloudBlobClient.getContainerReference("test-rejected");
-        this.rejectedContainer.createIfNotExists();
+    @AfterEach
+    protected void deleteAllContainers() {
```
nitinprabhu, line 74:
```suggestion
    protected void deleteTestContainer() {
```

PR #1450, opened 2020-08-10
Add content type to headers for OCR validation client call

doncem:
### Change description ###

e2e tests continuously fails. checked latest build and this caught my attention:

> 415 Unsupported Media Type: [{"timestamp":"2020-08-09T09:47:07.206+0000","status":415,"error":"Unsupported Media Type","message":"Content type 'application/xml;charset=UTF-8' not supported","path":"/forms/D8/validate-ocr"}]

**note**: application/xml

```
[ ] Yes
[x] No
```


rkondratowicz:
it would be nice to update tests

PR #1454, opened 2020-08-12
Scheduled daily reports

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1367

### Change description ###

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/email/ReportSenderTest.java
```diff
@@ -106,4 +108,8 @@ private JavaMailSender getMailSender() {
 
         return mailSender;
     }
+
+    private LocalDate getYesterday() {
+        return LocalDate.now().minusDays(1);
+    }
```
rkondratowicz, line 114: It's called once.
Isn't
```
LocalDate yesterday = LocalDate.now().minusDays(1);
```
clear?

apereverzin: This approach is similar to what I used in https://github.com/hmcts/blob-router-service/pull/598 and I assume it is justified improving readability

apereverzin: ok, I will inline this call

PR #1456, opened 2020-08-12
use new version of azure-blob-api in DeleteCompleteFilesTask

mdayican:

### Change description ###
use new version of azure-blob-api in DeleteCompleteFilesTask

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
@@ -101,7 +101,7 @@ public class DeleteCompleteFilesTask {
 
             BlobClient blobClient = container.getBlobClient(envelope.getZipFileName());
 
-            if (Boolean.TRUE.equals(blobClient.exists())) {
+            if (Boolean.TRUE == blobClient.exists()) {
 
                 leaseAcquirer.ifAcquiredOrElse(
                     blobClient,
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
@@ -101,7 +101,7 @@ public class DeleteCompleteFilesTask {
 
             BlobClient blobClient = container.getBlobClient(envelope.getZipFileName());
 
-            if (Boolean.TRUE == blobClient.exists()) {
+            if (Boolean.TRUE.equals(blobClient.exists())) {
 
                 leaseAcquirer.ifAcquiredOrElse(
                     blobClient,
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
@@ -1,8 +1,11 @@
 package uk.gov.hmcts.reform.bulkscanprocessor.tasks;
 
-import com.microsoft.azure.storage.AccessCondition;
-import com.microsoft.azure.storage.blob.CloudBlobContainer;
-import com.microsoft.azure.storage.blob.CloudBlockBlob;
+import com.azure.core.http.rest.Response;
+import com.azure.core.util.Context;
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.models.BlobRequestConditions;
+import com.azure.storage.blob.models.DeleteSnapshotsOptionType;
 import net.javacrumbs.shedlock.spring.annotation.SchedulerLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -80,58 +85,72 @@ private void processCompleteFiles(CloudBlobContainer container) {
...
 
-            CloudBlockBlob cloudBlockBlob = container.getBlockBlobReference(envelope.getZipFileName());
+            BlobClient blobClient = container.getBlobClient(envelope.getZipFileName());
 
-            if (cloudBlockBlob.exists()) {
-                Optional<String> leaseId = blobManager.acquireLease(
-                    cloudBlockBlob,
-                    container.getName(),
-                    envelope.getZipFileName()
+            if (Boolean.TRUE.equals(blobClient.exists())) {
```
apereverzin, line 104: is _equals_ necessary?

doncem: same question

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -80,58 +85,72 @@ private void processCompleteFiles(CloudBlobContainer container) {
...
+
+                leaseAcquirer.ifAcquiredOrElse(
+                    blobClient,
+                    leaseId -> delete(blobClient, leaseId),
+                    DeleteCompleteFilesTask::throwBlobDeleteException,
```
doncem, line 109: looks like failure case is caught and logged in `delete` function. perhaps we just can get rid of `if exists` statement and just use this line for logging information that blob has already been deleted or smth

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
mdayican: > is _equals_ necessary?

no

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
mdayican: you mean instead of checking existence we can directly try to delete ?

doncem: Yea i am hoping we can avoid making extra call and just try delete straight away

mdayican: but in that case we depend on azure exceptions which I do not want now. Lets see how this goes and we can refactor later.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
mdayican: removed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
doncem: sounds good to me. still - does it need exception throw? it wasn't throwing before

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
mdayican: Sonar did not liked without equals.

mdayican: https://sonarcloud.io/project/issues?id=Verification%3Abulk-scan-processor&open=AXPnlJAnYbrraxsQDZRm&pullRequest=1456&resolved=false&types=BUG

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
mdayican: in this case it should throw,  we mark the file deleted if there is no error.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -80,58 +85,63 @@ private void processCompleteFiles(CloudBlobContainer container) {
...
-                log.info("Marked envelope as deleted. {}", loggingContext);
-            }
+            envelope.setZipDeleted(true);
+            envelopeRepository.saveAndFlush(envelope);
+            log.info("Marked envelope as deleted. {}", loggingContext);
 
-            return deleted;
+            return true;
 
         } catch (Exception ex) {
```
apereverzin, line 123: The logic looks simpler than it used to be regarding returning _true_ or _false_ and it makes sense to catch exception (and log it) in the calling method rather than check if the result is _true_ or _false_.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/DeleteCompleteFilesTask.java
```diff
@@ -80,58 +85,63 @@ private void processCompleteFiles(CloudBlobContainer container) {
...
             log.error("Failed to process file. {}", loggingContext, ex);
             return false;
         }
     }
+
+    private void delete(BlobClient blobClient, String leaseId) {
```
apereverzin, line 129: _deleteBlob_ definitely looks a better method name

PR #1472, opened 2020-08-18
clear metadata after success

mdayican:

### Change description ###

clear blob metadata after successful operation.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseAcquirer.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseAcquirer.java
@@ -65,7 +65,7 @@ public class LeaseAcquirer {
             if (isReady) {
                 onLeaseSuccess.accept(leaseId);
                 if (releaseLease) {
-                    release(leaseClient, blobClient);
+                    releaseAll(leaseClient, blobClient);
                 }
             }
         } catch (BlobStorageException exc) {
@@ -90,6 +90,21 @@ public class LeaseAcquirer {
         } catch (BlobStorageException exc) {
             logger.warn(
                 "Could not release the lease with ID {}. Blob: {}, container: {}",
+                leaseClient.getLeaseId(),
+                blobClient.getBlobName(),
+                blobClient.getContainerName(),
+                exc
+            );
+        }
+    }
+
+    private void releaseAll(BlobLeaseClient leaseClient, BlobClient blobClient) {
+        try {
+            leaseMetaDataChecker.clearMetaData(blobClient, leaseClient.getLeaseId());
+            leaseClient.releaseLease();
+        } catch (BlobStorageException exc) {
+            logger.warn(
+                "Could not release lease or clear metadata, lease with ID {}. Blob: {}, container: {}",
                 leaseClient.getLeaseId(),
                 blobClient.getBlobName(),
                 blobClient.getContainerName(),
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseMetaDataChecker.java
```diff
@@ -72,4 +73,14 @@ private boolean isMetaDataLeaseExpired(String leaseExpirationTime) {
                 .isBefore(LocalDateTime.now(EUROPE_LONDON_ZONE_ID)); // check if lease expired
         }
     }
+
+    public void clearMetaData(BlobClient blobClient, String leaseId) {
+        Map<String, String> blobMetaData = new HashMap<>();
+        blobMetaData.put(LEASE_EXPIRATION_TIME, null);
+        blobClient.setMetadataWithResponse(
+            blobMetaData,
```
nitinprabhu, line 81: can we not just pass null here instead of map with expiry time null ?

mdayican:
> why do we need to clear metadata ? If you just release the lease after successful operation then next task would be to delete it anyways ?

deleting using the same logic. it checks metadata as well.

nitinprabhu:
> > why do we need to clear metadata ? If you just release the lease after successful operation then next task would be to delete it anyways ?
> 
> deleting using the same logic. it checks metadata as well.

I mean what is the advantage of doing it this way ? Is the complexity reduced or we have less code ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseMetaDataChecker.java
mdayican: make sense. let's try.

mdayican:
> > > why do we need to clear metadata ? If you just release the lease after successful operation then next task would be to delete it anyways ?
> > 
> > 
> > deleting using the same logic. it checks metadata as well.
> 
> I mean what is the advantage of doing it this way ? Is the complexity reduced or we have less code ?

using same class for lease operation makes simpler . Later we can completely remove metadata stuff for all operations.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseAcquirer.java
```diff
@@ -98,4 +98,19 @@ private void release(BlobLeaseClient leaseClient, BlobClient blobClient) {
         }
     }
 
+    private void releaseAll(BlobLeaseClient leaseClient, BlobClient blobClient) {
```
nitinprabhu, line 101: method name doesn't look correct `clearMetadataAndReleaseLease` ?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/LeaseAcquirer.java
```diff
@@ -98,4 +98,19 @@ private void release(BlobLeaseClient leaseClient, BlobClient blobClient) {
         }
     }
 
+    private void clearMetadataAndReleaseLease(BlobLeaseClient leaseClient, BlobClient blobClient) {
+        try {
+            leaseMetaDataChecker.clearMetaData(blobClient, leaseClient.getLeaseId());
+            release(leaseClient, blobClient);
+        } catch (BlobStorageException exc) {
+            logger.warn(
+                "Could not clear metadata, lease with ID {}. Blob: {}, container: {}",
```
nitinprabhu, line 107: if release Lease fails this message won't be correct ?

PR #1475, opened 2020-08-18
Blobmanager - change copy blob

mdayican:

### Change description ###

Getting ready to use new version of blob lib in BlobProcessorTask.
New methods is not used  yet, next pr they will use them.
Method names will be fixed after everything upgraded to new version.  

```
[ ] Yes
[x] No
```


doncem (approved):
i reckon afters there will be some cleanup and those `new` prefixes can be removed

mdayican:
> i reckon afters there will be some cleanup and those `new` prefixes can be removed

sure. there are other names to fix as well. like starting with `get`  will be list.

doncem:
> like starting with get will be list

those are ok to me at least :)

aliveni (approved):

PR #1482, opened 2020-08-19
Upgrade blob lib part 8 use in blob processor

mdayican:

### Change description ###

Upgrade blob use in blob processor.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTask.java
```diff
@@ -50,132 +48,136 @@
...
-    ) throws StorageException, IOException {
-        Optional<String> leaseIdOption = blobManager.acquireLease(cloudBlockBlob, container.getName(), zipFilename);
+    ) throws BlobStorageException {
 
-        if (leaseIdOption.isPresent()) {
-            String leaseId = leaseIdOption.get();
+        leaseAcquirer.ifAcquiredOrElse(
+            cloudBlockBlob,
+            leaseId -> processZipFile(container, cloudBlockBlob, zipFilename, leaseId),
+            s -> {},
```
doncem, line 128: perhaps a comment here why nothing to do on failure?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
```diff
@@ -322,7 +328,7 @@ private void newMoveFileToRejectedContainer(
 
         try {
             inputBlob.deleteWithResponse(
-                DeleteSnapshotsOptionType.ONLY,
+                DeleteSnapshotsOptionType.INCLUDE,
```
mdayican, line 331: previous version we were using DeleteSnapshotsOption.NONE which `Specifies deleting the blob only. If the blob has snapshots, this option will result in an error from the service`
I think In new version it does not result in error if there is snapshots.  We can pass 'null'  it will delete just the blob. INCLUDE deletes all,  ONLY deletes only snapshots.

doncem (approved):

mdayican:
stuck in pipeline

PR #1485, opened 2020-08-19
Upgrade blob lib - use in blob processor

mdayican:

### Change description ###

previous #1482 one stuck.

Upgrade blob use in blob processor.
Because of integration test it can not make smaller than this.
Copying blob just with URL did not work in functional test so I add SAS token to URL.

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
```diff
@@ -315,14 +327,19 @@ private void newMoveFileToRejectedContainer(
             // next steps will overwrite the file, create a snapshot of current version
             rejectedBlob.createSnapshot();
         }
-        String copyId = rejectedBlob.copyFromUrl(inputBlob.getBlobUrl());
+
+        String copyId = rejectedBlob.copyFromUrl(
+            inputBlob.getBlobUrl() + "?" + inputBlob.generateSas(
+                new BlobServiceSasSignatureValues(OffsetDateTime.now().plusSeconds(50), SAS_READ_PERMISSION)
+            )
+        );
```
apereverzin, line 335: I suggest to use the same code for generating SAS token as in router (BlobMover class)

apereverzin (approved):

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/processor/BlobManager.java
mdayican: it is same. just this one created a static var for `BlobContainerSasPermission`

mdayican: the question is why do we need sas these are in the same account. I tried without sastoken and functional test are failed in Preview. it might be specific to `Osba`.

apereverzin: `sourceBlob
                .generateSas(
                    new BlobServiceSasSignatureValues(
                        OffsetDateTime.of(LocalDateTime.now().plus(5, ChronoUnit.MINUTES), ZoneOffset.UTC),
                        new BlobContainerSasPermission().setReadPermission(true)
                    )
                )`

it looks like this in router, the difference is minor but maybe it makes sense to eliminate even minor differences. router is definitely more advanced and I should follow its patterns rather than reinventing

mdayican: will add UTC.

doncem (approved):

PR #1486, opened 2020-08-20
ReceivedZipFileConverter

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1363 - reconciliation in processor

### Change description ###
ReceivedZipFileRepository fetches multiple entries for each envelope (containing cartesian product of all scannable item and payment dcns). This is difficult to compare with the content of the request from router.

ReceivedZipFileConverter converts the result from ReceivedZipFileRepository into the list of instances of ReceivedZipFileData which is much handier for further comparison.

Also moved ReceivedZipFileItem class to tests.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
@@ -1,0 +1,52 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.services.reports;
+
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.reports.ReceivedZipFile;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.models.ReceivedZipFileData;
+import uk.gov.hmcts.reform.bulkscanprocessor.services.reports.models.ZipFileIdentifier;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toList;
+
+public class ReceivedZipFileConverter {
+
+    public List<ReceivedZipFileData> convertReceivedZipFiles(List<ReceivedZipFile> receivedZipFiles) {
+        return receivedZipFiles.stream()
+            .collect(groupingBy(file ->
+                                    new ZipFileIdentifier(
+                                        file.getZipFileName(),
+                                        file.getContainer()
+                                    )
+            ))
+            .values().stream()
+            .map(this::getReceivedZipFileData)
+            .collect(toList());
+    }
+
+    private ReceivedZipFileData getReceivedZipFileData(List<ReceivedZipFile> receivedZipFilesList) {
+        Set<String> scannableItemDcns = new HashSet<>();
+        Set<String> paymentDcns = new HashSet<>();
+        receivedZipFilesList.forEach(
+            receivedZipFile -> {
+                if (receivedZipFile.getScannableItemDcn() != null) {
+                    scannableItemDcns.add(receivedZipFile.getScannableItemDcn());
+                }
+                if (receivedZipFile.getPaymentDcn() != null) {
+                    paymentDcns.add(receivedZipFile.getPaymentDcn());
+                }
+            }
+        );
+        return new ReceivedZipFileData(
+            new ZipFileIdentifier(
+                receivedZipFilesList.get(0).getZipFileName(),
+                receivedZipFilesList.get(0).getContainer()
+            ),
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
```diff
@@ -0,0 +1,52 @@
...
+public class ReceivedZipFileConverter {
+
+    public List<ReceivedZipFileData> convertReceivedZipFiles(List<ReceivedZipFile> receivedZipFiles) {
+        return receivedZipFiles.stream()
+            .collect(groupingBy(file ->
+                                    new ZipFileIdentifier(
+                                        file.getZipFileName(),
+                                        file.getContainer()
+                                    )
+            ))
```
doncem, line 24: so this is grouping by new object all the time? doesn't it make 1-to-1 map here? perhaps we are better of with `new tuple2` here. moreover - we have vavr library - it certainly supports such grouping collection

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
```diff
@@ -0,0 +1,52 @@
...
+            .values().stream()
```
doncem, line 25: both `stream()`s should be newlined

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
doncem: especially when it is dropped/disregarded next line - why even need extra model/class in our source code?

doncem (commented):
why `ZipFileIdentifier`? not only on design but in code we can flatten it to just being core elements of parent object: zip file + container

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReceivedZipFileConverter.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public List<ReceivedZipFileData> convertReceivedZipFiles(List<ReceivedZipFile> receivedZipFiles) {
+        return receivedZipFiles
+            .stream()
+            .collect(groupingBy(file ->
+                                    new Tuple2<>(
+                                        file.getZipFileName(),
+                                        file.getContainer()
+                                    )
+            ))
```
doncem, line 25: can we inline somehow? it's a simple operation and entire stack would be so much clearer:

```
stream
collect
values
stream
map
collect
happy days
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/models/ReceivedZipFileData.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.List;
+
+public class ReceivedZipFileData {
+    public final Tuple2 zipFileIdentifier;
+    public final List<String> scannableItemDcns;
+    public final List<String> paymentDcns;
+
+    public ReceivedZipFileData(
+        Tuple2 zipFileIdentifier,
```
doncem, line 13: perhaps we need flattened values here ;)

apereverzin: do you mean - single line of code? to be honest I like this format, if the line becomes too long it is difficult to read without scrolling

doncem: I mean flatten the `Tuple2` :wink:

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/models/ReceivedZipFileData.java
```diff
@@ -0,0 +1,22 @@
...
+    public final List<String> scannableItemDcns;
+    public final List<String> paymentDcns;
+
+    public ReceivedZipFileData(
+        String zipFileName,
+        String container,
+        List<String> scannableItemDcns,
+        List<String> paymentDcns
+    ) {
+        this.zipFileIdentifier = new Tuple2<>(zipFileName, container);
```
doncem, line 18: why this tuple is still here?

apereverzin: Oh, stupid me

doncem (approved):

PR #1491, opened 2020-08-24
Reconciliation service

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1363

### Change description ###
ReconciliationService along with unit tests.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationServiceTest.java
```diff
@@ -0,0 +1,420 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.services.reports;
```
rkondratowicz, line 1: please fix formatting in this file

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationServiceTest.java
```diff
@@ -0,0 +1,420 @@
...
+        ReconciliationStatement statement = new ReconciliationStatement(date, envelopes);
+
+        // when
+        List<Discrepancy> discrepancies = reconciliationService.getReconciliationReport(statement);
+
+        // then
+        assertThat(discrepancies)
+            .usingRecursiveFieldByFieldElementComparator()
+            .containsExactlyInAnyOrder(
+                new Discrepancy("file1", "c1", PAYMENT_DCNS_MISMATCH.text, "[pay-1, pay-2]","[]")
```
rkondratowicz, line 409: it would be good to have at least one test that covers a case when there are more than 1 discrepancy.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,114 @@
...
+    ) {
+        this.receivedZipFileRepository = receivedZipFileRepository;
+        this.receivedZipFileConverter = receivedZipFileConverter;
+    }
+
+    public List<Discrepancy> getReconciliationReport(ReconciliationStatement reconciliationStatement) {
+        List<ReceivedZipFile> receivedZipFiles =
+            receivedZipFileRepository.getReceivedZipFilesReportFor(reconciliationStatement.date);
+        List<ZipFileData> receivedZipFileDataList =
+            new ArrayList<>(receivedZipFileConverter.convertReceivedZipFiles(receivedZipFiles));
```
rkondratowicz, line 38: is this needed?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,114 @@
...
+
+        List<Discrepancy> discrepancies = new ArrayList<>();
+
+        Map<ZipFileData, ZipFileData> envelopesMap =
+            reconciliationStatement.envelopes.stream().collect(toMap(envelope -> envelope, envelope -> envelope));
```
rkondratowicz, line 43: what is this map for?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationServiceTest.java
apereverzin: `should_return_reported_but_not_received_and_received_but_not_reported_discrepancies` - this is hopefully what does this. I will add more tests

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: Yes. The repository returns cartesian product of payment DCNs and scannable document DCNs, and `receivedZipFileConverter` transforms it into 2 proper lists.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: This is to provide access to the reported envelopes. If we use list we can find (`contains(...)`) if the object is in the list (but inefficiently performance wise) but cannot get access to this object. Using map we can find if the object is in the map (`containsKey(...)`easily and easily get access to it. Keep in mind that `ZipFileData` class is essential here - it provides (thank to its `equals` and `hashCode` methods) comparison of reported and received file (`ReportedZipFileData` and `ReceivedZipFile`).

rkondratowicz: why not a set?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/models/ZipFileData.java
```diff
@@ -0,0 +1,41 @@
...
+        ZipFileData that = (ZipFileData) o;
+        return zipFileName.equals(that.zipFileName)
+            && container.equals(that.container);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(zipFileName, container);
+    }
+}
```
rkondratowicz, line 41: how about, instead of creating this base class and extending it in two places and then using it as a key in a map to find a matching file, we simply do something like this in the service code?

```java
Optional<ReportedZipFile> matchingFile = 
  reportedFiles
    .stream()
    .filter(reported -> reported.zipFileName.equals(actual.zipFileName)
    .filter(reported -> reported.container.equals(actual.container)
    .findFirst();
```

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: Because there is no `get` operation in `Set`. When `Set.contains(...)` is called in this case 2 different objects are compared - reported file and received file, and the comparison operation compares `zipFileName` and `container`. But if `contains` returns `true` (i.e. reported files have an entry with the same `zipFileName` and `container` as received files) we need to extract (i.e. `get`) this reported file to compare fields (`scannableDoumentDcns` and `paymentDcns`). `Map` has this `get` operation, `Set` does not.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/models/ZipFileData.java
apereverzin: Good point. But in this case we will have to iterate across the whole collection of reported files as many times as there are received files (i.e. 1000+ times with what Exela sends us these days)

rkondratowicz: how about just using a `Pair` (storing file name + container) as a key in that map then?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
rkondratowicz: can it be just
```
List<ReceivedZipFileData> receivedZipFileDataList =
 receivedZipFileConverter.convertReceivedZipFiles(receivedZipFiles));
```
? (no wrapping in `new ArrayList(...)`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,133 @@
...
+                .collect(
+                    toMap(
+                        receivedZipFile -> Pair.of(receivedZipFile.zipFileName, receivedZipFile.container), identity()
+                    )
+                );
+
+        List<Discrepancy> discrepancies = new ArrayList<>();
+
+        receivedZipFilesMap.keySet()
+            .forEach(receivedZipFileKey -> {
```
rkondratowicz, line 63: you can just iterate over `receivedZipFileDataList`

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: My bad

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,134 @@
...
+        List<ReceivedZipFileData> receivedZipFileDataList =
+            receivedZipFileConverter.convertReceivedZipFiles(receivedZipFiles);
+        Map<Pair<String, String>, ReceivedZipFileData> receivedZipFilesMap =
+            receivedZipFileDataList
+                .stream()
+                .collect(
+                    toMap(
+                        receivedZipFile -> Pair.of(receivedZipFile.zipFileName, receivedZipFile.container), identity()
+                    )
+                );
```
rkondratowicz, line 58: no need for this map, you can just iterate over `receivedZipFileDataList`

apereverzin: yes, indeed

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,134 @@
...
+                    toMap(
+                        envelope -> Pair.of(envelope.zipFileName, envelope.container), identity()
+                    )
+                );
+
+        List<ReceivedZipFile> receivedZipFiles =
+            receivedZipFileRepository.getReceivedZipFilesReportFor(reconciliationStatement.date);
+        List<ReceivedZipFileData> receivedZipFileDataList =
+            receivedZipFileConverter.convertReceivedZipFiles(receivedZipFiles);
+        Set<Pair<String, String>> receivedZipFilesSet =
```
rkondratowicz, line 54: this is not used until line 94

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,134 @@
...
+                            RECEIVED_BUT_NOT_REPORTED.text
+                        )
+                    );
+                }
+            });
+        reportedZipFilesMap.keySet()
+            .stream()
+            .filter(reportedZipFileKey -> !receivedZipFilesSet.contains(reportedZipFileKey))
+            .forEach(reportedZipFileKey -> {
+                final ReportedZipFile reportedZipFile = reportedZipFilesMap.get(reportedZipFileKey);
```
rkondratowicz, line 96: doesn't the key have both fields used below?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: agree. also extracted the method

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
apereverzin: agree

rkondratowicz (approved):

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private void findReportedButNotReceived(
+        Map<Pair<String, String>, ReportedZipFile> reportedZipFilesMap,
+        List<ReceivedZipFileData> receivedZipFileDataList,
+        List<Discrepancy> discrepancies
+    ) {
+        Set<Pair<String, String>> receivedZipFilesSet =
+            receivedZipFileDataList
+                .stream()
+                .map(receivedZipFile -> Pair.of(receivedZipFile.zipFileName, receivedZipFile.container))
```
doncem, line 101: extra stream is happening here with action already executed before. we could map over initial stream (list.foreach) and collect to set so this extra step can be avoided

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/ReconciliationService.java
```diff
@@ -0,0 +1,142 @@
...
+                .collect(toSet());
+        reportedZipFilesMap.keySet()
```
doncem, line 103: `ReportedZipFile` doesn't play anything here. perhaps we can pass key set straight away?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/reports/models/ReportedZipFile.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+
+public class ReportedZipFile {
+    public final String zipFileName;
+    public final String container;
+    public final String rescanFor;
+    public final List<String> scannableItemDcns;
+    public final List<String> paymentDcns;
+
+    @JsonCreator
```
doncem, line 15: is this needed?

doncem (approved):

PR #1493, opened 2020-08-24
upgrade blob lib - use in blob processor

mdayican:

### Change description ###

Use new version of blob lib blob processor. 

Changes were reverted with #1489  as copyfromurl was not working in bulk-scan staging and this PR again trying to merge 
same changes.
instead of copyfromurl, using copy blob which does download and upload . 

 
```
[ ] Yes
[x] No
```


doncem (approved):

apereverzin (approved):

mdayican:
stuck in pipeline

PR #1502, opened 2020-08-26
Use DiscrepancyType in Discrepancy class

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1363

### Change description ###
This is to make code cleaner

```
[ ] Yes
[x] No
```


doncem (commented):
and when serialised is it definitely using custom text and not name?

apereverzin:
If you mean DiscrepancyItem @doncem there is `ReconciationControllerTest#should_return_discrepancies_for_a_given_date`  to verify that the response body is correct. `Discrepancy` class is not supposed to be serialized

rkondratowicz (approved):

aliveni (approved):

PR #1512, opened 2020-09-07
Added retry count for OCR validation

apereverzin:

### Change description ###
This config parameter will be used when retrying to send OCR validation request in case of failure (5xx, etc)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -219,5 +219,4 @@ process-payments:
 
 use-wrapping-zip: true
 
-ocr-validation:
-  retry-count: 2
+ocr-validation-max-tries: 2
--- a/src/main/resources/application.yaml
+++ b/src/main/resources/application.yaml
@@ -219,4 +219,4 @@ process-payments:
 
 use-wrapping-zip: true
 
-ocr-validation-max-tries: 2
+ocr-validation-max-retries: 2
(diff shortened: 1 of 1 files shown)


src/main/resources/application.yaml
```diff
@@ -218,3 +218,7 @@ process-payments:
   enabled: ${PROCESS_PAYMENTS_ENABLED:true}
 
 use-wrapping-zip: true
+
+service-config:
```
aliveni, line 222: is it a service-specific config?

apereverzin: no, it is not. is there a better config for this situation?

rkondratowicz: you can make it a flat property.
maybe `ocr-valiration-max-tries`

src/main/resources/application.yaml
```diff
@@ -218,3 +218,6 @@ process-payments:
   enabled: ${PROCESS_PAYMENTS_ENABLED:true}
 
 use-wrapping-zip: true
+
+ocr-validation:
+  retry-count: 2
```
rkondratowicz, line 223: do you plan to add any other properties under `ocr-validation`?

rkondratowicz: how many times will it try to do the validation 2 or 3?

apereverzin: 2 total tries

apereverzin: and there will be no other props in ocr-validation - as of now, so yes, will make it flat

mdayican (approved):

doncem (approved):

PR #1514, opened 2020-09-09
Envelope ocr validator

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1386

### Change description ###
Because the new ocr validation scheduled job will be dealing with Envelope (a DB entity class) rather than InpuEnvelope (a memory class OcrValidator is deailing with atm) we need new validator classes. These new classes will be called by the new job.

The new classes are based on existing OcrValidator and OcrPresenceValidator classes.

```
[ ] Yes
[x] No
```


rkondratowicz (changes requested):
OCR presence validation cannot be moved - currently if this fails envelope is not created in DB

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/EnvelopeOcrPresenceValidator.java
```diff
@@ -0,0 +1,54 @@
...
+        if (docs.stream().filter(doc -> doc.getOcrData() != null).count() > 1) {
+            throw new OcrPresenceException(MULTIPLE_OCR_MSG);
+        }
+        if (docs.stream().anyMatch(doc -> !OCR_DOC_TYPES.contains(doc.getDocumentType()) && doc.getOcrData() != null)) {
+            throw new OcrPresenceException(MISPLACED_OCR_MSG);
+        }
+        if (docs.stream().anyMatch(doc -> OCR_DOC_TYPES.contains(doc.getDocumentType()) && doc.getOcrData() == null)) {
+            throw new OcrPresenceException(MISSING_OCR_MSG);
+        }
+        // TODO: Looks like this check is not necessary as documentSubtype is never null
```
doncem, line 41: if never null maybe empty string?

doncem (dismissed):

apereverzin:
Only OCR validation calls should be in the new scheduled job

PR #1516, opened 2020-09-09
EnvelopeOcrValidator

apereverzin:
https://tools.hmcts.net/jira/browse/BPS-1386

### Change description ###
`EnvelopeOcrValidator` will be executed in the new OcrValidationTask for the envelopes created created in the database by the main job. At the moment of creation of the envelope OcrPresenceValidator has already checked that the OCRs are properly set in the file that is why `EnvelopeOcrValidator::findDocWithOcr` method contains only short extract from `OcrPresenceValidator::assertHasProperlySetOcr` method.

`EnvelopeOcrValidatorTest` class contains all test methods from `OcrValidatorTest` class apart from `should_validate_the_presence_of_ocr_data` which checks exception thrown from `OcrPresenceValidator`.

Existing `OcrValidator` class will keep only the call to `OcrPresenceValidator::findDocWithOcr` method, everything else will go as soon as the new job becomes active.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/EnvelopeOcrValidator.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/EnvelopeOcrValidator.java
@@ -1,0 +1,179 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.validation;
+
+import io.vavr.control.Try;
+import joptsimple.internal.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.data.util.Optionals;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.HttpClientErrorException.NotFound;
+import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
+import uk.gov.hmcts.reform.bulkscanprocessor.config.ContainerMappings;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.Envelope;
+import uk.gov.hmcts.reform.bulkscanprocessor.entity.ScannableItem;
+import uk.gov.hmcts.reform.bulkscanprocessor.exceptions.OcrValidationException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/EnvelopeOcrValidator.java
```diff
@@ -0,0 +1,179 @@
...
+                .fields
+                .stream()
+                .map(it -> new OcrDataField(it.name.asText(), it.value.textValue()))
+                .collect(toList())
+        );
+    }
+
+    private String getFormType(ScannableItem item) {
+        return item.getDocumentSubtype();
+    }
```
rkondratowicz, line 178: do we need this method?

rkondratowicz: Also this is different from original
```
    private String getFormType(InputScannableItem item) {
        return item.documentType == InputDocumentType.SSCS1 ? "SSCS1" : item.documentSubtype;
    }
```
Can you explain?

Are there any other changes you've made?

apereverzin: Because there is no `DocumentType.SSCS1`. In `EnvelopeMapper` class `InputDocumentType.SSCS1` is converted into `DocumentType.FORM` and `DocumentSubtype.SSCS1`.

apereverzin: Looks like you are right because this is sent to the service. Will fix this.

apereverzin: There are no other changes in `EnvelopeOcrValidator` compared to `OcrValidator`. But later on it will be changed to not throwing `OcrValidationException` in case of OCR validation error response from the service.

apereverzin:
Will be retried using metadat

PR #1521, opened 2020-09-10
Metadata retry

apereverzin:

### Change description ###
`OcrValidationRetryManager` using blob metadata for storing retry related properties.

`ocr-validation-delay-retry-ms` property renamed to `ocr-validation-delay-retry-sec`

`OcrValidationRetryManager` will be autowired into `OcrValidator` in the next PR. An integration test using `azure-storage` docker image will also be introduced in this new PR.

```
[ ] Yes
[x] No
```


src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @BeforeEach
+    void setUp() {
+        ocrValidationRetryManager = new OcrValidationRetryManager(2, 120);
+
+        given(blobClient.getProperties()).willReturn(blobProperties);
+    }
+
+    @Test
+    void isReadyToRetry_should_return_true_if_delay_expiration_not_set() {
```
rkondratowicz, line 44: missing 'is'?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    void isReadyToRetry_should_return_false_if_delay_expiration_time_not_passed() {
+        // given
+        metadata = Map.of(
+            "ocrValidationRetryCount",
+            "1",
+            "ocrValidationRetryDelayExpirationTime",
+                now(EUROPE_LONDON_ZONE_ID).plusSeconds(10).toString()
+        );
```
rkondratowicz, line 62: formatting

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,138 @@
...
+        ocrValidationRetryManager = new OcrValidationRetryManager(2, 120);
+
+        given(blobClient.getProperties()).willReturn(blobProperties);
+    }
+
+    @Test
+    void isReadyToRetry_should_return_true_if_delay_expiration_not_set() {
+        // given
+        metadata = emptyMap();
+        given(blobProperties.getMetadata()).willReturn(metadata);
```
rkondratowicz, line 47: either put that in setUp or use local variable for `metadata`

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,138 @@
...
+    private BlobClient blobClient;
+
+    @Mock
+    private BlobProperties blobProperties;
+
+    private Map<String, String> metadata;
+
+    @BeforeEach
+    void setUp() {
+        ocrValidationRetryManager = new OcrValidationRetryManager(2, 120);
```
rkondratowicz, line 38: can 2 and 120 be constant that are then used in tests to get relavant times?

rkondratowicz (commented):
Can we have an integration test, please?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManager.java
```diff
@@ -0,0 +1,95 @@
...
+        Map<String, String> blobMetaData = blobClient.getProperties().getMetadata();
+        String retryDelayExpirationTime = blobMetaData.get(RETRY_DELAY_EXPIRATION_TIME_METADATA_PROPERTY);
+        var zipFilename = blobClient.getBlobName();
+        var containerName = blobClient.getContainerName();
+
+        logger.info(
+            "Checking if retry delay expired on file {} in container {}",
+            zipFilename,
+            containerName
+        );
```
doncem, line 47: dunno, i would only be interested if i'm not ready to retry. just a personal opinion here

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManager.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private int getRetryCount(Map<String, String> blobMetaData) {
+        String retryCountStr = blobMetaData.get(RETRY_COUNT_METADATA_PROPERTY);
+
+        if (StringUtils.isBlank(retryCountStr)) {
+            return 0;
+        } else {
+            return Integer.parseInt(retryCountStr);
+        }
```
doncem, line 93: looks like a good candidate for ternary return statement

src/main/resources/application.yaml
```diff
@@ -221,4 +221,4 @@ process-payments:
 use-wrapping-zip: true
 
 ocr-validation-max-retries: 2
-ocr-validation-delay-retry-ms: 120000
+ocr-validation-delay-retry-sec: 120
```
doncem, line 224: a bit small window. especially when only 2 retries i would expect bigger window and/or even apply some geometric progression to make window longer after each retry

doncem: p.s. can think about it and apply in another pr ;)

doncem (approved):

mdayican (commented):
Need to be Thread safe.
All blob operations should be synchronised, need to leased so other thread should not change the value.

rkondratowicz (commented):
Can you add integration tests? (with dockerized storage)

apereverzin:
`OcrValidationRetryManager` is not called from the task so it is not feasible at the moment to run an integration test involving processing real file uploaded to the docker image of the blob storage.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+        LeaseAcquirer leaseAcquirer,
+        @Value("${ocr-validation-max-retries}") int ocrValidationMaxRetries,
+        @Value("${ocr-validation-delay-retry-sec}") int ocrValidationRetryDelaySec
+    ) {
+        this.leaseAcquirer = leaseAcquirer;
+        this.ocrValidationMaxRetries = ocrValidationMaxRetries;
+        this.ocrValidationRetryDelaySec = ocrValidationRetryDelaySec;
+    }
+
+    public boolean isRetryDelayNotSetOrExpired(BlobClient blobClient) {
```
rkondratowicz, line 42: Quite a mouthful.
Does this basically mean the passed argument is ready for retry?

apereverzin: Renamed to `isReadyToRetry`

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,136 @@
...
+        dockerHost = dockerComposeContainer.getServiceHost("azure-storage", 10000);
+    }
+
+    @AfterAll
+    static void tearDownContainer() {
+        dockerComposeContainer.stop();
+    }
+
+    @Test
+    void should_read_blob_and_save_metadata_in_database_when_zip_contains_metadata_and_pdfs()
```
mdayican, line 75: I think this test has many tests in it. Better to divide and mock the time.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+        this.leaseAcquirer = leaseAcquirer;
+        this.ocrValidationMaxRetries = ocrValidationMaxRetries;
+        this.ocrValidationRetryDelaySec = ocrValidationRetryDelaySec;
+    }
+
+    public boolean isReadyToRetry(BlobClient blobClient) {
+        Map<String, String> blobMetaData = blobClient.getProperties().getMetadata();
+        String retryDelayExpirationTime = blobMetaData.get(RETRY_DELAY_EXPIRATION_TIME_METADATA_PROPERTY);
+
+        final boolean retryDelayExpired = isRetryDelayExpired(retryDelayExpirationTime);
```
mdayican, line 46: if there is no metadata but blob is locked should we return `true`?

mdayican: I think we need to lease first?

apereverzin: this method is renamed to `isReadyToRetry` (or maybe even `isReadyToBeProcessed` will be better) and its intent is to check if it is possible to retry, does not matter if the file is locked or not. Retry will be executed by the caller of this method. To check the property in metadata lease is not necessary.

mdayican: while this returning true meantime another thread can mark the metadata

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManager.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Check if number of retries is exceeding ocrValidationMaxRetries,
+     * if not set new retry delay expiration time, increment retryCount and return true,
+     * otherwise return false which means no more retries are possible.
+     *
+     * @return true if retry is possible and new retry delay expiration time has been set,
+     *         false if no more retries are possible
+     */
+    public boolean setRetryDelayIfPossible(BlobClient blobClient) {
+        Map<String, String> blobMetaData = blobClient.getProperties().getMetadata();
```
mdayican, line 68: this should be done after blob leased. 
Meantime another tread can increase the number but thread would continue with old value.

apereverzin: Agree with you here

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/services/storage/OcrValidationRetryManagerTest.java
```diff
@@ -0,0 +1,144 @@
...
+        throws Exception {
+        //Given
+        uploadToBlobStorage(SAMPLE_ZIP_FILE_NAME, zipDir("zipcontents/ok"));
+
+        BlobClient blobClient = testContainer.getBlobClient(SAMPLE_ZIP_FILE_NAME);
+
+        leaseAcquirer.ifAcquiredOrElse(
```
mdayican, line 85: this needs to be changed?
now you do not have anything with leaseAcquirer.
you can upload and get lease id. there should be sth `uploadToBlobStorageAndLease` method

mdayican (approved):

PR #1531, opened 2020-09-17
Add rescan_for property to the metafile schema

aliveni:
https://tools.hmcts.net/jira/browse/BPS-1395

### Change description ###
- Add `rescan_for` property to the metafile schema
- `rescan_for` is not a mandatory field. 
- When the value is present, the file name format should match the existing property `zip_file_name`.

```
[ ] Yes
[x] No
```


src/functionalTest/java/uk/gov/hmcts/reform/bulkscanprocessor/controllers/BlobProcessorTest.java
```diff
@@ -36,7 +36,7 @@ public void should_process_zipfile_after_upload_and_set_status() {
     }
 
     @Test
-    public void should_process_zipfile_with_supplementary_evidence_with_oce_classification() {
+    public void should_process_zipfile_with_supplementary_evidence_with_ocr_classification() {
```
aliveni, line 39: not related to the change, but noticed when adding the rescan_for property to the existing functional test data

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTestForInvalidFiles.java
```diff
@@ -258,6 +258,22 @@ public void should_not_parse_envelope_with_duplicate_payment_dcn_numbers() throw
             .hasMessageContaining("instance: {\"pointer\":\"/payments\"}");
     }
 
+    @Test
+    public void should_not_parse_envelope_with_invalid_zip_file_name_in_rescan_for_property() throws IOException {
```
rkondratowicz, line 262: will metadata with entry like `"rescan_for" : null` pass validation?

src/main/resources/metafile-schema.json
```diff
@@ -48,6 +48,12 @@
       "title": "The name of the zip file",
       "pattern": "^\\d+_([012][0-9]|30|31)-([0][0-9]|[1][012])-[2][0][0-9][0-9]-([01][0-9]|[2][0123])-[0-5][0-9]-[0-5][0-9]\\.(test\\.)?zip$"
     },
+    "rescan_for": {
+      "id": "#/properties/rescan_for",
+      "type": "string",
```
doncem, line 53: can it be null?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/MetafileJsonValidatorTestForInvalidFiles.java
rkondratowicz: let's add an extra test to verify it passes validation, when field is not included (similar to `should_parse_envelope_data_with_no_case_number`)

src/main/resources/metafile-schema.json
rkondratowicz, line 53: I think this should be `"type": ["string", "null"],`

rkondratowicz (changes requested):
let's add more tests

src/main/resources/metafile-schema.json
aliveni: no, I'll update the `type` values to support null

aliveni:
> let's add more tests

Added tests for `no rescan_for` scenario and `rescan_for` with null value

rkondratowicz (approved):

mdayican (approved):

PR #1551, opened 2020-09-24
Call ocr validation retry

apereverzin:

### Change description ###

```
[ ] Yes
[x] No
```


src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuiteForOcrValidation.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+@IntegrationTest
+@TestPropertySource(properties = {
+    "scheduling.task.scan.enabled=true"
+})
+abstract class ProcessorTestSuiteForOcrValidation {
```
akomisarek, line 44: Would it be possible to reuse existing ProcessorTestSuite? Now there is some duplication, and similar code to be maintained in two places. 

If not maybe similar methods should be extracted into the helpers, i.e.uploadToBlobStorage and similar.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForOcrValidationFailedOnServerSide.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.mockito.BDDMockito.given;
+import static uk.gov.hmcts.reform.bulkscanprocessor.entity.Status.CREATED;
+import static uk.gov.hmcts.reform.bulkscanprocessor.helper.DirectoryZipper.zipDir;
+import static uk.gov.hmcts.reform.bulkscanprocessor.model.common.Event.ZIPFILE_PROCESSING_STARTED;
+
+@IntegrationTest
+@TestPropertySource(properties = {
+    "scheduling.task.scan.enabled=true"
+})
+class BlobProcessorTaskTestForOcrValidationFailedOnServerSide extends ProcessorTestSuiteForOcrValidation {
```
akomisarek, line 29: I am not sure if this test is required here. I mean the integration was already proved above, and this repeats something, which I guess is already tested on the unit test level? 

But as a newcomer not sure how you approach those integration tests, so it's fine if you cover the functionality more thoroughly, but I think it's against the testing pyramid (a bit). 

My another point it's difficult to understand why this test returns warning and above one doesn't. Why retrying for the second time is special? I found it is driven by the config, but maybe it could be added as the comment

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/FileContentProcessor.java
```diff
@@ -96,6 +97,14 @@ public void processZipFileContent(
         } catch (PreviouslyFailedToUploadException ex) {
             log.warn("Rejected file {} from container {} - failed previously", zipFilename, containerName, ex);
             createEvent(DOC_UPLOAD_FAILURE, containerName, zipFilename, ex.getMessage());
+        } catch (OcrValidationServerSideException ex) {
+            log.error("Failed to process file {} from container {}, "
+                          + "message from OCR validation endpoint {} - will retry",
+                      zipFilename,
+                      containerName,
+                      ex.getMessage(),
```
akomisarek, line 105: I don't think logging ex.getMessage is required, it will be anyway logged as the whole ex is passed.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -83,14 +89,54 @@ public OcrValidator(
...
+                            + "Url: %s, DCN: %s, doc type: %s, doc subtype: %s, envelope: %s. "
+                            + "Error message: %s, status code: %s, status text: %s",
+                        validationUrl,
+                        docWithOcr.documentControlNumber,
+                        docWithOcr.documentType,
+                        docWithOcr.documentSubtype,
+                        zipFileName,
+                        ex.getMessage(),
+                        ((HttpServerErrorException) ex).getStatusCode().toString(),
+                        ((HttpServerErrorException) ex).getStatusText()
```
akomisarek, line 132: Instead of extracting data from ex, couldn't the ex be passed as param to the exception? Will be much easier to trace as the proper stacktraces will be preserved (and other information).

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidatorTest.java
```diff
@@ -79,8 +87,15 @@
...
-    public void setUp() throws Exception {
-        this.ocrValidator = new OcrValidator(client, presenceValidator, containerMappings, authTokenGenerator);
+    public void setUp() {
+        this.ocrValidator = new OcrValidator(
+            client,
+            presenceValidator,
+            containerMappings,
+            authTokenGenerator,
+            ocrValidationRetryManager
+
```
akomisarek, line 97: Super minor - not needed extra line.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidatorTest.java
```diff
@@ -456,6 +471,107 @@ public void should_log_info_when_ocr_is_present_but_theres_not_service_configure
...
+        );
+
+        // then
+        assertThat(exception.getMessage().matches(
+            "Error calling validation endpoint. Url: https://example.com/validate-ocr, "
+                + "DCN: (.*), doc type: Form, doc subtype: x, "
+                + "envelope: file.zip. Error message: 500 internal server error message, "
+                + "status code: 500 INTERNAL_SERVER_ERROR, status text: internal server error message")
+        )
+            .isTrue();
```
akomisarek, line 524:
```suggestion
        assertThat(exception.getMessage()).matches(
            "Error calling validation endpoint. Url: https://example.com/validate-ocr, "
                + "DCN: (.*), doc type: Form, doc subtype: x, "
                + "envelope: file.zip. Error message: 500 internal server error message, "
                + "status code: 500 INTERNAL_SERVER_ERROR, status text: internal server error message");
```

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeHandlerTest.java
```diff
@@ -144,7 +145,47 @@ void should_handle_and_save_envelope() {
...
+        verify(envelopeValidator).assertEnvelopeContainsOcrDataIfRequired(inputEnvelope);
+        verify(envelopeValidator).assertEnvelopeHasPdfs(inputEnvelope, pdfs);
+        verify(envelopeValidator).assertDocumentControlNumbersAreUnique(inputEnvelope);
+        verify(envelopeValidator).assertPaymentsEnabledForContainerIfPaymentsArePresent(
+            inputEnvelope, paymentsEnabled, containerMappings.getMappings()
+        );
+        verify(envelopeValidator).assertEnvelopeContainsDocsOfAllowedTypesOnly(inputEnvelope);
+        verify(envelopeProcessor).assertDidNotFailToUploadBefore(inputEnvelope.zipFileName, CONTAINER_NAME);
+
+        assertThat(exception.getMessage()).isEqualTo("msg");
```
akomisarek, line 188: Are those verification required? This test is checking if the EnvelopeHandler rethrows the exceptions, calling those other validator is side effect not related to the tests and checked already in the test above. I would probably skip those.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/BlobProcessorTaskTestForOcrValidationFailedOnServerSide.java
apereverzin: In the current production code in case of unavailability of the OCR validation endpoint on the service side the warning is raised straight away. 

In this PR (according the task) there are several attempts (the number of defined by the config) to hit this endpoint (delay between the attempts is also defined by the config). If the number of attempts is expired - the warning is raised and processing of the file goes on.

src/integrationTest/java/uk/gov/hmcts/reform/bulkscanprocessor/tasks/ProcessorTestSuiteForOcrValidation.java
apereverzin: Implemented `ProcessorTestSuiteForOcrValidation -> ProcessorTestSuite` inheritance. Getting rid fully of `ProcessorTestSuiteForOcrValidation` is too cumbersome for this task.

akomisarek (approved):
Code seems to do what's expected and contains test (as not happy path scenario, it makes sense to stop on integration tests). 

I left a couple of comments, which are not addressed but they are nothing major.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrValidationServerSideException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.hmcts.reform.bulkscanprocessor.exceptions;
+
+import org.springframework.web.client.HttpServerErrorException;
+
+public class OcrValidationServerSideException extends RuntimeException {
```
rkondratowicz, line 5: What do you mean by server side? can you add javadoc?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -83,14 +89,51 @@ public OcrValidator(
                         authTokenGenerator.generate()
                     ))
                     .onSuccess(res -> handleValidationResponse(res, envelope, docWithOcr))
-                    .onFailure(exc -> handleRestClientException(exc, validationUrl, envelope, docWithOcr))
+                    .onFailure(exc -> handleRestClientException(
+                        exc,
+                        validationUrl,
+                        envelope,
+                        docWithOcr
+                    ))
```
rkondratowicz, line 97: has something changed here?

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
```diff
@@ -83,14 +89,51 @@ public OcrValidator(
...
+    private OcrValidationWarnings handleOcrValidationError(
+        Throwable ex,
+        BlobClient blobClient,
+        String leaseId,
+        InputScannableItem docWithOcr,
+        String validationUrl,
+        String zipFileName
+    ) {
+        if (ex instanceof HttpServerErrorException) {
+            if (ocrValidationRetryManager.setRetryDelayIfPossible(blobClient, leaseId)) {
```
rkondratowicz, line 119: should both this scenario (maximum number of retries) and exceptions different than `HttpServerErrorException` be handled the same exact way?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidatorTest.java
```diff
@@ -456,6 +470,104 @@ public void should_log_info_when_ocr_is_present_but_theres_not_service_configure
...
+            .willReturn(false);
+
+        // when
+        Optional<OcrValidationWarnings> warnings = ocrValidator.assertOcrDataIsValid(envelope, blobClient, LEASE_ID);
+
+        // then
+        assertThat(warnings).isPresent();
+        assertThat(warnings.get().documentControlNumber).isEqualTo(scannableItemWithOcr.documentControlNumber);
+        assertThat(warnings.get().warnings).containsExactly("OCR validation was not performed due to errors");
+        verify(client).validate(anyString(), any(FormData.class), anyString(), anyString());
```
rkondratowicz, line 568: test says `should_throw_exception_if....`. Looks like it does not throw an exception.

rkondratowicz (changes requested):
Fix issues

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/exceptions/OcrValidationServerSideException.java
apereverzin: Added JavaDoc.

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
apereverzin: reverted

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidator.java
apereverzin: Yes. In the existing code any http status response code is handled by raising OCR validation warnings. After this change in case of 5xx status response code will cause several retries after which the workflow will fall back to raising OCR validation warnings.

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/validation/OcrValidatorTest.java
apereverzin: Renamed to `should_raise_ocr_validation_warnings_if_service_responded_with_500_and_retry_delay_is_not_possible`.

apereverzin:
Failing in CI

PR #1560, opened 2020-09-29
add rejected Discrepancy

mdayican:

https://tools.hmcts.net/jira/browse/BPS-1403

### Change description ###

add rejected Discrepancy,
if it envelope is not in envelope table assume it is rejected.  
new 'envelopeId' field added to`ReceivedZipFileItem`.
if it is rejected and reported, creates rejected discrepancy. if it is rejected but not reported creates received_but_not_reported discrepancy.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ReceivedZipFileRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ReceivedZipFileRepository.java
@@ -21,6 +21,7 @@ public interface ReceivedZipFileRepository extends JpaRepository<Envelope, UUID>
             + "  payments.documentControlNumber AS paymentDCN, "
             + "  envelopes.rescanFor, "
             + "  Cast(envelopes.id as varchar) as envelopeId " // null means no envelope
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ReceivedZipFileRepository.java
```diff
@@ -19,7 +19,8 @@
             + "  process_events.createdat AS processingStartedEventDate, "
             + "  scannable_items.documentControlNumber AS scannableItemDCN, "
             + "  payments.documentControlNumber AS paymentDCN, "
-            + "  envelopes.rescanFor "
+            + "  envelopes.rescanFor, "
+            + "  Cast(envelopes.id as varchar) as envelopeId " // null means no envelope
```
rkondratowicz, line 23: why casting to varchar here but then use UUID in java model?

mdayican: to map to UUID, it needs.

apereverzin (approved):

PR #1733, opened 2021-02-12
ZipFileFailureHandler

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-356

### Change description ###
ZipFileFailureHandler will be called from FileContentProcessor when catching generic exception in processZipFileContent method (in the next PR). 

Envelope record (with the newly introduced `ZIP_PROCESSING_FAILURE` status) is created in db in order to make the failure visible in `/stale-incomplete-blobs` endpoint, the zip file causing the problem will be moved to rejected container without sending error notification (also in the next PR).

After verifying the problem (based on the result of `/stale-incomplete-blobs` endpoint) and possible reporting it to the provider the envelope can be moved to COMPLETED status with help of devops.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/Status.java
@@ -10,6 +10,7 @@ public enum Status {
     CREATED,
     METADATA_FAILURE,   // when we are aware of envelope, but there are inconsistency among files and metadata info
     SIGNATURE_FAILURE,  // the zip archive failed signature verification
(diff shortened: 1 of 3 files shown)


mdayican (commented):
if we are going to save envelope why do not we do when first get the blob from storage.

mdayican (commented):
I thought we just need to change the exception type so it would understand it as unrecoverable and give up retrying and send notification. if we are going to save envelope I think we need to think deeply.

apereverzin:
There will be a task to create envelope record immediately finding a zip file.

PR #1752, opened 2021-02-26
Included scannable items into zip status

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-387

### Change description ###

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/zipfilestatus/ZipFileEnvelope.java
```diff
@@ -33,7 +47,11 @@ public ZipFileEnvelope(
         String ccdId,
         String envelopeCcdAction,
         boolean zipDeleted,
-        String rescanFor
+        String rescanFor,
+        List<ScannableItemResponse> scannableItems,
+        List<NonScannableItemResponse> nonScannableItems,
+        List<PaymentResponse> payments
```
mdayican, line 53: List<String> would be enough?

apereverzin: This is the same as in `EnvelopeResponse`, used in `EnvelopeController` for `/envelopes/{envelopeId}' endpoint.

mdayican (approved):

PR #1771, opened 2021-03-08
Dtsbps 393 change data type of created at in blob info

gauravchawla73463:

https://tools.hmcts.net/jira/browse/DTSBPS-393

### Change description ###

Datatype of variable CreatedAt to Instant in BlobInfo class

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/BlobInfo.java
```diff
@@ -11,12 +13,12 @@
     public final String fileName;
 
     @JsonProperty("created_at")
-    public final String createdAt;
+    public final Instant createdAt;
```
mdayican, line 16: this should be in new PR

mdayican: all changes in this class should be in new PR

gauravchawla73463:
deleted

PR #1774, opened 2021-03-09
Dts 393 new branch for blob info class

gauravchawla73463:
### Change description ###

Datatype change for variable "CreatedAt" in BlobInfo class

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/model/out/BlobInfo.java
```diff
@@ -11,12 +13,12 @@
     public final String fileName;
 
     @JsonProperty("created_at")
-    public final String createdAt;
+    public final Instant createdAt;
```
apereverzin, line 16: Should also have
`@JsonSerialize(using = InstantSerializer.class)`
annotation

gauravchawla73463: done

PR #1779, opened 2021-03-12
remove lease logic

mdayican:

https://tools.hmcts.net/jira/browse/DTSBPS-409

### Change description ###

we may need more than 1 minute to process big files so removing lease logic which has max 1 minute lock.
Using ETag for Optimistic Lock

```
[ ] Yes
[x] No
```


apereverzin:
The Jira link in the description looks confusing

mdayican:
> The Jira link in the description looks confusing

😃 kids books .

apereverzin:
We should also consider using db instead of ETag in the future, as far as I remember we discussed this option.

apereverzin (approved):

PR #1873, opened 2021-05-05
DTSBPS-483-repo

gauravchawla73463:

https://tools.hmcts.net/jira/browse/DTSBPS-483

### Change description ###
Include payments in zip-files-summary report - bulk-scan-processor

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
+++ b/src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
@@ -13,7 +13,7 @@ public interface ZipFilesSummaryRepository extends JpaRepository<Envelope, UUID>
     @Query(
         nativeQuery = true,
         value = "SELECT "
-            + "Cast(envelope.id as varchar) as envelopeId, "
+            + "payment.documentcontrolnumber, "
             + "createdEvent.container, "
             + "createdEvent.zipfilename, "
             + "createdEvent.createdDate, "
@@ -47,6 +47,8 @@ public interface ZipFilesSummaryRepository extends JpaRepository<Envelope, UUID>
             + "LEFT JOIN envelopes AS envelope "
             + "  ON createdEvent.zipfilename = envelope.zipfilename "
             + "  AND createdEvent.container = envelope.container "
+            + "LEFT JOIN payments AS payment "
+            + "  ON payment.envelope_id = envelope.id "
             + "ORDER BY createdEvent.createdDate ASC"
     )
     List<ZipFileSummary> getZipFileSummaryReportFor(@Param("date") LocalDate date);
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
```diff
@@ -13,7 +13,7 @@
     @Query(
         nativeQuery = true,
         value = "SELECT "
-            + "Cast(envelope.id as varchar) as envelopeId, "
+            + "payment.documentcontrolnumber, "
```
mdayican, line 16: why envelope.id removed ?

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/reports/zipfilesummary/ZipFileSummaryItem.java
```diff
@@ -15,7 +15,7 @@
     private final String classification;
     private final String ccdId;
     private final String ccdAction;
-    private final String envelopeId;
+    private final String documentControlNumber;
```
mdayican, line 18: there may be more than one payment record

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/entity/reports/ZipFilesSummaryRepository.java
gauravchawla73463: Its easy to get payment with document control number rather than envelope.id without using envelopeRepository

src/test/java/uk/gov/hmcts/reform/bulkscanprocessor/helper/reports/zipfilesummary/ZipFileSummaryItem.java
gauravchawla73463: Wouldn't there be another record for another payment record. We are mapping it with dcn. If we have more than one payment record, there will different records for them If that makes sense?

mdayican: in one envelope there might be more than 1 payment and you are search for envelope.

gauravchawla73463: Yes right but for second payment we will have second record with same envelopeID

mdayican: if it is the case it is completely wrong. You are counting the returned values and adding counts etc.

PR #1884, opened 2021-05-12
moveEnvelopeToCompleted

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-576

### Change description ###

```
[ ] Yes
[x] No
```


mdayican (commented):
can we first rename, refactor and then logic change

src/main/java/uk/gov/hmcts/reform/bulkscanprocessor/services/EnvelopeActionService.java
```diff
@@ -82,6 +103,25 @@ private void validateEnvelopeState(Envelope envelope) {
         }
     }
 
+    private void validateEnvelopeIsInInconsistentState(Envelope envelope) {
+        if (envelope.getStatus() != Status.NOTIFICATION_SENT) {
```
mdayican, line 107: do we care which state in envelope as long as it is completed?
we can check the event and if there is a completed event but the envelope has different status we can change the status?

mdayican: it would be more generic but this it is fine as well.

mdayican (approved):

PR #2005, opened 2021-07-13
Add fortify

apereverzin:
### Change description ###

```
[ ] Yes
[x] No
```


build.gradle
```diff
@@ -303,6 +303,7 @@ dependencies {
   }
   testImplementation group: 'org.apache.commons', name: 'commons-email', version: '1.5'
   testImplementation group: 'io.github.netmikey.logunit', name: 'logunit-core', version: '1.1.0'
+  testImplementation 'com.github.hmcts:fortify-client:1.0.1:all'
```
mdayican, line 306: format is wrong, it should be like one above

mdayican:
we do not need

PR #2093, opened 2021-09-24
ReceivedPaymentsRepo

apereverzin:
https://tools.hmcts.net/jira/browse/DTSBPS-753

### Change description ###

```
[ ] Yes
[x] No
```


mdayican (commented):
why do you need this endpoint?

apereverzin:
It is similar to scannable documents. Some people consider payments a type of scannable documents and need statistics for both scannable documents and payments

mdayican:
> It is similar to scannable documents. Some people consider payments a type of scannable documents and need statistics for both scannable documents and payments

I think we can not add all these reports as endpoint .
May be we need unprocessed payment details.

mdayican (approved):
unprocessed payment details  would be more useful.

PR #3325, opened 2024-08-19
FACT-1897 moved paymentsEnabled to values.yaml

justiceia:
https://tools.hmcts.net/jira/browse/FACT-1897

### Change description ###
moved hard coded values to values.yaml

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/resources/application-dev.yaml
+++ b/src/main/resources/application-dev.yaml
@@ -128,19 +128,19 @@ containers:
         - 12625
         - 12624
       ocrValidationUrl: ${OCR_VALIDATION_URL_PROBATE}
-      paymentsEnabled: {$PAYMENTS_ENABLED_PROBATE}
+      paymentsEnabled: {$PAYMENTS_ENABLED_PROBATE:true}
     - container: divorce
       jurisdiction: DIVORCE
       poBoxes:
         - 12706
       ocrValidationUrl: ${OCR_VALIDATION_URL_DIVORCE}
-      paymentsEnabled: {$PAYMENTS_ENABLED_DIVORCE}
+      paymentsEnabled: {$PAYMENTS_ENABLED_DIVORCE:true}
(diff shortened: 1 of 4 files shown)


src/main/resources/application-dev.yaml
```diff
@@ -128,19 +128,19 @@ containers:
         - 12625
         - 12624
       ocrValidationUrl: ${OCR_VALIDATION_URL_PROBATE}
-      paymentsEnabled: true
+      paymentsEnabled: {$PAYMENTS_ENABLED_PROBATE}
```
RuthKirby, line 131: I think we can be setting a default value of false with these new environment variables. So they are like the other binary ones in the application.yaml such as ${NOTIFICATIONS_TO_ORCHESTRATOR_TASK_ENABLED:false}, ${SCAN_ENABLED:false}, ${LAUNCH_DARKLY_OFFLINE_MODE:false} etc.

RuthKirby (changes requested):
Looking good! Just a comment about the default values on the variables. Also we may want to move PAYMENTS_ENABLED_PRIVATELAW: "false" to be next to these new ones.

src/main/resources/application-dev.yaml
RuthKirby: > added changes

It looks like some are default to true. All should default to false. Basically what we are saying by having all default as false is: if the payment enabled variable is missing/not set, then do not allow payments to be enabled.

RuthKirby (approved):

RuthKirby (changes requested):
Can you move the dollar signs outside of the curly braces so this:{$PAYMENTS_ENABLED_DIVORCE:false} becomes this ${PAYMENTS_ENABLED_DIVORCE:false} :)

RuthKirby (approved):

